PR #8, opened 2018-01-25
added keylisterner for DEL_ to delete all selected Presets during mul…

schlenkibus:
…tiple preset selection


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -462,6 +462,12 @@ public Control onKey(final KeyDownEvent event) {
 		} else if (keyCode == com.google.gwt.event.dom.client.KeyCodes.KEY_H) {
 			toggleHiddenBanks();
 			return this;
+		} else if(keyCode == com.google.gwt.event.dom.client.KeyCodes.KEY_DELETE) {
```
hhoegelo: maybe we should delete the one and only selected preset if there is no multi selection? Would be consistent, I think.

schlenkibus: Stephan approved that idea.

PR #11, opened 2018-01-30
added unresponding PleaseWaitLayout

schlenkibus:
I have never seen a loading time longer than a second so i don't see the need to be able to cancel the screen. Furthermore it would be hard to implement. due to the need to do the loading asynchronous .

### DIFF
--- a/playground/CMakeLists.txt
+++ b/playground/CMakeLists.txt
@@ -60,7 +60,7 @@ ADD_CUSTOM_TARGET(
 
 file(GLOB_RECURSE SOURCE_FILES src/*.cpp) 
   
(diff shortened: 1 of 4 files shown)


playground/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ ADD_CUSTOM_TARGET(
 
 file(GLOB_RECURSE SOURCE_FILES src/*.cpp) 
   
-add_executable(playground ${SOURCE_FILES})
+add_executable(playground ${SOURCE_FILES} src/proxies/hwui/panel-unit/boled/file/PleaseWaitLayout.cpp src/proxies/hwui/panel-unit/boled/file/PleaseWaitLayout.h)
```
hhoegelo: nö

PR #14, opened 2018-01-31
Feature bank position and Favicon

schlenkibus:


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/BankInfoDialog.java
```diff
@@ -256,7 +295,18 @@ private void updateInfo(Bank bank) {
 					name.setText(bankName);
 				}
 			}
+
```
hhoegelo: looks bit too complicated to me

playground/src/presets/BankActions.cpp
```diff
@@ -358,6 +359,27 @@ BankActions::BankActions(PresetManager &presetManager) :
...
+        int newPos = stoi(request->get("order-number"));
+        newPos = std::max(newPos, 0);
+        newPos = std::min(newPos, numBanks);
+
+        auto oldBankPos = m_presetManager.calcOrderNumber(bank.get());
+
+        auto scope = m_presetManager.getUndoScope().startTransaction("Changed Bank: %0 Order", bank->getName(true));
+        auto movedPos = newPos - oldBankPos;
+        if(movedPos != 0)
+          m_presetManager.undoableMoveBankBy(scope->getTransaction(), bank->getUuid(), movedPos);
```
hhoegelo: This method doesn't do what it looks like ....

hhoegelo: movedPos is meant to be a direction: >0 means move up, <0 means move down by 1

PR #28, opened 2018-02-16
Misc changes at the colour tagging feature 

schlenkibus:
moved colour line in info
changed css to be more centred horizontal
changed highlight
added automatic attribute setting for colour if not present (server proxy now has "setPresetAttributeHidden", that starts cuckoo transaction)

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/ServerProxy.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/ServerProxy.java
@@ -721,21 +721,8 @@ public class ServerProxy {
 		StaticURI uri = new StaticURI(path, new StaticURI.KeyValue("id", parameterID), new StaticURI.KeyValue("info", text));
 		queueJob(uri, false);
 	}
-
-	public void setPresetAttributeHidden(Preset thePreset, String key, String value) {
-		if(thePreset == null)
-			return;
-		
-		StaticURI.Path path = new StaticURI.Path("presets", "banks", "set-preset-attribute-hidden");
-		StaticURI uri = new StaticURI(path, new StaticURI.KeyValue("uuid", thePreset.getUUID()), new StaticURI.KeyValue("key", key),
-				new StaticURI.KeyValue("value", value));
-		queueJob(uri, false);
-	}
 	
 	public void setPresetAttribute(Preset thePreset, String key, String value) {
-		if(thePreset == null)
-			return;
-		
 		StaticURI.Path path = new StaticURI.Path("presets", "banks", "set-preset-attribute");
 		StaticURI uri = new StaticURI(path, new StaticURI.KeyValue("uuid", thePreset.getUUID()), new StaticURI.KeyValue("key", key),
 				new StaticURI.KeyValue("value", value));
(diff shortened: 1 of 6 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/ServerProxy.java
```diff
@@ -722,7 +722,20 @@ public void setMacroControlInfo(int parameterID, String text) {
...
+			return;
+		
+		StaticURI.Path path = new StaticURI.Path("presets", "banks", "set-preset-attribute-hidden");
+		StaticURI uri = new StaticURI(path, new StaticURI.KeyValue("uuid", thePreset.getUUID()), new StaticURI.KeyValue("key", key),
+				new StaticURI.KeyValue("value", value));
+		queueJob(uri, false);
+	}
+	
 	public void setPresetAttribute(Preset thePreset, String key, String value) {
+		if(thePreset == null)
```
hhoegelo: I think, if the method is called with thePreset == null, there is a bug on callers side which should be fixed.

NonMaps/src/com/nonlinearlabs/NonMaps/client/ServerProxy.java
```diff
@@ -722,7 +722,20 @@ public void setMacroControlInfo(int parameterID, String text) {
 		queueJob(uri, false);
 	}
 
+	public void setPresetAttributeHidden(Preset thePreset, String key, String value) {
```
hhoegelo: I don't believe that we need such an ugly function - as far as I see, all you want to achieve is defaulting empty color attribute to 'none' - this should better be done on playgrounds side - for example in Preset CTor.

NonMaps/src/com/nonlinearlabs/NonMaps/client/ServerProxy.java
schlenkibus: public void setPresetAttribute(Preset thePreset, String key, String value) {
		if(thePreset == null)
			return;

schlenkibus: Or do you mean that shouldn't be there?

hhoegelo: shouldn't

NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/PresetInfoDialog.java
```diff
@@ -360,9 +358,14 @@ private void updateInfo(Preset preset) {
 			Bank bank = preset.getParent();
 			bankName.setText(bank.getOrderNumber() + " - " + bank.getTitleName());
 			
-			ColorTag.Color c = ColorTag.Color.valueOf(preset.getAttribute("color"));
-			updateCurrentHighlight(colors[c.ordinal()]);
 			
+			String presetColorString = preset.getAttribute("color");
```
hhoegelo: hier wuerde ich die sonderbehandlung des Strings "" in ColorTag.Color.valueOf verschieben

playground/src/presets/BankActions.cpp
```diff
@@ -788,6 +788,20 @@ BankActions::BankActions(PresetManager &presetManager) :
     }
   });
 
+  addAction("set-preset-attribute-hidden", [&] (shared_ptr<NetworkRequest> request) mutable
```
hhoegelo: das kann doch nun weg, oder?

PR #37, opened 2018-02-21
Feature end multiple selection if all compare windows are closed

schlenkibus:
Implemented the correct behaviour:  closing multiple selection if compare is opened


hhoegelo:
not exactly what the branch name says, but should be ok as well

PR #41, opened 2018-02-21
is mustn't a word? Changed Click behaviour for presets in maps

schlenkibus:
moved selecting out of mouse down into click that a dragged preset mustn't be selected while being dragged, that helps with moving from one bank into a different bank in the belt


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/preset/Preset.java
```diff
@@ -212,11 +212,22 @@ public Control click(Position point) {
 			invalidate(INVALIDATION_FLAG_UI_CHANGED);
 			return this;
 		}
+
```
hhoegelo: this can't be correct - at least you added dead code, but probably it is also wrong and in any case misleading

PR #57, opened 2018-02-28
"Feature make belt great again" again

schlenkibus:
Various fixes and enhancements


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/presets/BeltPreset.java
```diff
@@ -37,8 +37,8 @@ protected BeltPreset(PresetList parent, Preset mapsPreset) {
 		color = addChild(new PresetColorTag(this));
 		number = addChild(new PresetNumber(this));
 		name = addChild(new PresetName(this));
-		((Label) name).setFontHeightInMM(4);
-		((Label) number).setFontHeightInMM(4);
+		((Label) name).setFontHeightInMM(4.5);
+		((Label) number).setFontHeightInMM(4.5);
 	}
```
hhoegelo: better store name and number as Label fields - so casts won't be necessary anymore.

PR #64, opened 2018-03-01
fix several issues with drag/drop multiple presets

hhoegelo:


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/preset/Preset.java
```diff
@@ -257,13 +258,18 @@ public Control startDragging(Position pos) {
...
 		PresetManager pm = world.getPresetManager();
 		MultiplePresetSelection selection = pm.getMultiSelection();
+		selection.add(this);
 
 		Control ret = null;
-		double yMargin = getPixRect().getTop() - pos.getY();
+		double yMargin = getPixRect().getTop() - mousePos.getY();
+		double xMargin = getPixRect().getLeft() - mousePos.getX();
+
+		Tracer.log("yMargin: " + yMargin);
```
schlenkibus: we shouldn't leave Traces in the Code after we are done Debugging.
It is just a unusable flood of information.

PR #66, opened 2018-03-01
fix performance degradation by DragProxy-traces; allow preset drop on…

hhoegelo:
… own banks header


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/DragProxy.java
```diff
@@ -11,7 +11,6 @@
 import com.nonlinearlabs.NonMaps.client.Animator.DoubleClientData.Client;
 import com.nonlinearlabs.NonMaps.client.Millimeter;
 import com.nonlinearlabs.NonMaps.client.NonMaps;
-import com.nonlinearlabs.NonMaps.client.Tracer;
```
schlenkibus: :top:

PR #69, opened 2018-03-01
Fix no select or load on context menu

hhoegelo:


schlenkibus:
This branch includes changes made by another fix-branch

PR #71, opened 2018-03-01
added dirty hack to meet specifications

schlenkibus:


playground/src/presets/BankActions.cpp
```diff
@@ -1111,12 +1111,15 @@ PresetManager::tBankPtr BankActions::importBank(InStream& stream, const Glib::us
 
   newBank->undoableEnsurePresetSelection(transaction);
   newBank->undoableSetAttribute(transaction, "Name of Import File", fileName);
-  auto lastModifiedSeconds = stoull(lastModified) / 1000;
```
hhoegelo: if lastModified is not used anymore, please remove it from the whole callstack (including webUI)

playground/src/presets/BankActions.cpp
```diff
@@ -1111,12 +1111,15 @@ PresetManager::tBankPtr BankActions::importBank(InStream& stream, const Glib::us
...
   newBank->undoableSetAttribute(transaction, "Date of Import File", TimeTools::getAdjustedIso());
   newBank->undoableSetAttribute(transaction, "Name of Export File", "");
   newBank->undoableSetAttribute(transaction, "Date of Export File", "");
   newBank->undoableSelect(transaction);
+
   m_presetManager.undoableSelectBank(transaction, newBank->getUuid());
+
+  newBank->hackLastChangeTimeForImportOnly();
```
hhoegelo: What do you want to achieve here? Why messing up the lastModified timestamp?

playground/src/presets/PresetBank.cpp
```diff
@@ -772,24 +772,25 @@ const Glib::ustring PresetBank::calcStateString() const
   std::string lastModTimeIso = TimeTools::getIsoTime(m_lastChangedTimestamp);
   std::string lastImportTime = getAttribute("Date of Import File", "-1");
 
-  if(lastExportTimeIso.compare("-1") == 0)
```
hhoegelo: How about removing the hackLastChangeTimeForImportOnly() and instead calculating a time span between last modification time and import time. Then you can add some hacky tolerance in comparison without messing up the timestamp wich is used by other parts of the class, too?

playground/src/presets/BankActions.cpp
schlenkibus: done

PR #72, opened 2018-03-01
added cancel callback to renamepresetlayout to enable additional stuff

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/AppendOverwriteInsertButtonMenu.cpp
```diff
@@ -146,6 +146,8 @@ void AppendOverwriteInsertButtonMenu::pushRenameScreen()
       }
     }
     animate();
+  }, [=](){
+    animate();
```
hhoegelo: really animate even though the action was cancelled? I guess the Schulz-Zug should only run if there is an action to visualise. Probably you just want to do: Application::get().getHWUI()->undoableSetFocusAndMode(UIMode::Select); ?

schlenkibus: The Preset gets saved anyways. I would keep the Schulz-Zug

PR #83, opened 2018-03-07
Fix - Preset Drag on Itself also remove Insert Lines in Banks for unnecessary options

schlenkibus:
added Control setting that allows for dragging on itself or not. Overwritten by BeltPreset to fix a bug

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/DragProxy.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/DragProxy.java
@@ -138,7 +138,7 @@ public class DragProxy extends OverlayControl {
 							foundControls.add(new FoundControl(p, ctrl, ranking));
 						}
 					} else if(ctrl == getOrigin()) {
-						if(ctrl.canBeDraggedOnSelf()) {
+						if(ctrl.stopDragCompletelyIfDraggedOn()) {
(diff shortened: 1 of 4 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/DragProxy.java
```diff
@@ -137,8 +137,12 @@ public boolean onWayUpFound(Control ctrl) {
 						if (ranking >= 0) {
 							foundControls.add(new FoundControl(p, ctrl, ranking));
 						}
+					} else if(ctrl == getOrigin()) {
+						if(ctrl.canBeDraggedOnSelf()) {
+							foundControls.clear();
```
hhoegelo: canBeDraggedOnSelf now triggers a 'stop handle drag if ctrl is under mouse or drag proxy' - which is hard to understand and even more hard to name. I suggested another behaviour:
- if ctrl can not be dragged on self: do not add it to foundControls
- if it can, add it

in according controls drag method, check again if 'origin == this', and just do nothing.

As it is implemented now, the methods name is misleading. So either find a better name for canBeDraggedOnSelf describing the behaviour explained above, or implement a behaviour matching the given function name.

schlenkibus:
i still dont think that the name fits 100% but it is better

PR #85, opened 2018-03-07
Fix panning should do no harm

schlenkibus:


hhoegelo:
Please see my PR (branch feature-space-down-makes-world-pan). Let's discuss which of the two approaches we should merge. Your approach contains more code changes but probably has less risk of breaking things, while mine is probably more elegant but more risky as it globally changes mouse handling.

PR #87, opened 2018-03-08
if shift is held down while selecting preset, start multiple selection

hhoegelo:


schlenkibus:
postClick breaks all other ways of normaly starting and navigating in multiple Preset selection

PR #123, opened 2018-03-23
Fix performance search dialog

schlenkibus:


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -153,14 +153,15 @@ public void update(Node presetManagerNode) {
...
-
+			
 			Preset newPresetSelection = getSelectedPreset();
 
 			if (oldPresetSelection != newPresetSelection) {
 				onPresetSelectionChanged(newPresetSelection);
 			}
+			
+			if(NonMaps.get().getNonLinearWorld().getParameterEditor().isModified() == false)
+				refreshFilter(false);
```
hhoegelo: Wo do not refresh the filter if the edit buffer is dirty? This seems not correct to me.

schlenkibus: The Editbuffer has nothing to do with the Preset-Search.
What changes can be made to the editbuffer that affect the search?

NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -151,16 +153,17 @@ public void update(Node presetManagerNode) {
 				Node child = children.item(i);
 
 				if (child.getNodeName().equals("banks"))
-					updateBanks(child);
+					shouldUpdateFilter = updateBanks(child) || shouldUpdateFilter;
```
hhoegelo: usually one writes shouldUpdateFilter |= updateBanks(child); in situations like these, but however - it makes no difference

PR #125, opened 2018-03-26
Fix undo compare

schlenkibus:


playground/src/presets/PresetManagerActions.cpp
```diff
@@ -140,6 +141,9 @@ PresetManagerActions::PresetManagerActions(PresetManager &presetManager) :
       auto transaction = scope->getTransaction();
       auto xml = http->get("xml", "");
 
+      auto lpc = Application::get().getLPCProxy();
```
hhoegelo: I am a bit afraid merging this as this code would leave the program in a incorrect, not self-healing state if anywhere between lines 147 and 164 an exception is thrown.
I would feel happier if you could invent an object that calls toggleSupres... in Ctor and in Dtor. This would render this code exception safe and could be used in other places, too. I think ther ust be code in Preset.cpp suffering from the same problem.

playground/src/presets/PresetManagerActions.cpp
```diff
@@ -134,12 +135,28 @@ PresetManagerActions::PresetManagerActions(PresetManager &presetManager) :
 
   addAction("load-editbuffer", [&] (shared_ptr<NetworkRequest> request) mutable
   {
+    class LPCParameterChangeSupressor {
```
hhoegelo: could you please move this object into a separate file and use the new class where ever toggleSuppress... is currently used?

PR #132, opened 2018-04-18
fixed overwriting preset that a colortag will stay if eb has no color…

schlenkibus:
… tag


playground/src/presets/PresetBank.cpp
```diff
@@ -327,13 +327,20 @@ void PresetBank::undoableOverwritePreset(UNDO::Scope::tTransactionPtr transactio
...
 }
 
+void cleanupPresetColorTagForOverwriteSave(Preset *presetToOverwrite, Preset *src) {
+  if(src->getAttribute("color", "none") == "none")
+    src->setAttribute("color", presetToOverwrite->getAttribute("color", "none"));
+}
+
 void PresetBank::undoableOverwritePreset(UNDO::Scope::tTransactionPtr transaction, Preset *presetToOverwrite, Preset *src)
 {
+  cleanupPresetColorTagForOverwriteSave(presetToOverwrite, src);
```
hhoegelo: I don't get it: You are setting the src color attribute to the one in the target?

hhoegelo: You don't use the undo system, shouldn't you?

PR #136, opened 2018-04-19
swapped the indices for displaying the setting for glitch surpression

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/setup/PresetGlitchSuppressionEditor.cpp
```diff
@@ -25,6 +25,6 @@ const vector<ustring> &PresetGlitchSuppressionEditor::getDisplayStrings () const
 
 int PresetGlitchSuppressionEditor::getSelectedIndex () const
 {
-  return (int) Application::get ().getSettings ()->getSetting<PresetGlitchSuppression> ()->get ();
+  return (int) !Application::get ().getSettings ()->getSetting<PresetGlitchSuppression> ()->get ();
```
hhoegelo: don't get it - why?

PR #171, opened 2018-07-04
changed a few things regarding export and import -> readable name and…

schlenkibus:
… bigger file list

### DIFF
--- a/playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
+++ b/playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
@@ -12,6 +12,7 @@
 #include <tools/SpawnCommandLine.h>
 #include <tools/TimeTools.h>
 #include <algorithm>
+#include <tools/StringTools.h>
 #include "USBStickAvailableView.h"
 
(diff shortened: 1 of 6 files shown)


playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
```diff
@@ -69,7 +71,11 @@ void ExportBackupEditor::exportBanks ()
 
   auto timeStamp = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count();
 
-  auto targetNameWithStamp(std::string("/mnt/usb-stick/") + std::to_string(timeStamp) + c_backupTargetFile);
+  auto humanDate = std::to_string(TimeTools::getDisplayStringFromStamp(timeStamp));
+
+  humanDate.replace(humanDate.find(" "), 1, "-");
```
hhoegelo: Will replace only the first occurrence. Will also crash if find returns string::npos. Consider using
boost::replace_all_copy(s, " ", "-") or - even better - add a replaceAll(in, pattern, replacement) function in our StringTools - maybe use boost::replace_all_copy from within there.

PR #172, opened 2018-07-04
hot fix lock menu changing entry count creates invalid index 

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/ParameterEditButtonMenu.cpp
```diff
@@ -43,6 +43,9 @@ void ParameterEditButtonMenu::setup()
     addButton("Unlock all", bind(&ParameterEditButtonMenu::unlockAll, this));
 
   eb->onSelectionChanged(sigc::mem_fun(this, &ParameterEditButtonMenu::onParameterSelectionChanged));
+
+  sanitizeIndex();
+
   selectButton(s_lastAction);
```
hhoegelo: Shouldn't s_lastAction be sanitised as well? I suggest sanitizeIndex() to call "int sanitizeIndex(int foo)" - and use this new method to samitize s_lastAction.

PR #173, opened 2018-07-11
Feature copy multiple preset selection

schlenkibus:


playground/src/clipboard/Clipboard.cpp
```diff
@@ -122,6 +129,26 @@ void Clipboard::copyBank (const Glib::ustring &bankUuid)
...
+  auto scope = getUndoScope ().startTrashTransaction ();
+  m_currentContentWasCut = false;
+  m_content.reset (new PresetBank (pm.get ()));
+
+  auto uuids = StringTools::splitStringOnAnyDelimiter(csv, ',');
+
+  auto bank = static_pointer_cast<PresetBank>(m_content);
+
+  for(auto uuid: uuids) {
+    if(auto preset = pm->findPreset(uuid)) {
```
hhoegelo: here, you can pass preset into 'undoableCopyAnd....'

playground/src/clipboard/Clipboard.cpp
```diff
@@ -122,6 +129,26 @@ void Clipboard::copyBank (const Glib::ustring &bankUuid)
...
+      bank->undoableCopyAndAppendPreset(scope->getTransaction(), uuid);
+    }
+  }
+  onChange (UpdateDocumentContributor::ChangeFlags::Generic);
```
hhoegelo: onChange call shouldn't be necessary - it is called from within the undo command anyway

playground/src/clipboard/Clipboard.cpp
```diff
@@ -122,6 +129,26 @@ void Clipboard::copyBank (const Glib::ustring &bankUuid)
...
+  return true;
```
hhoegelo: what is the result for?

playground/src/clipboard/Clipboard.cpp
```diff
@@ -122,6 +129,26 @@ void Clipboard::copyBank (const Glib::ustring &bankUuid)
   }
 }
 
+bool Clipboard::copyPresets(const Glib::ustring &csv) {
```
hhoegelo: can be void - I think

playground/src/presets/PresetBank.cpp
```diff
@@ -357,6 +357,26 @@ void PresetBank::undoableInsertPreset(UNDO::Scope::tTransactionPtr transaction,
   });
 }
 
+void PresetBank::undoableCopyAndAppendPreset(UNDO::Scope::tTransactionPtr transaction, const Uuid &uuid)
```
hhoegelo: actually it is a prepend, as far as I see
So please either adapt the algorithm to the name or vice versa.

playground/src/presets/PresetBank.cpp
```diff
@@ -357,6 +357,25 @@ void PresetBank::undoableInsertPreset(UNDO::Scope::tTransactionPtr transaction,
   });
 }
 
+void PresetBank::undoableCopyAndPrependPreset(UNDO::Scope::tTransactionPtr transaction, tPresetPtr preset)
+{
+  if(preset) {
+    tPresetPtr newPreset(preset);
```
hhoegelo: don't need to do this copy of a shared_ptr here, it is copied anyway in createSwapData

playground/src/clipboard/MultiplePresetSelection.cpp
```diff
@@ -0,0 +1,35 @@
...
+  if(presetToCopy) {
+    PresetBank::tPresetPtr newPreset(presetToCopy);
+    auto swapData = UNDO::createSwapData(newPreset);
+    transaction->addSimpleCommand([ = ] (UNDO::Command::State) mutable
+    {
+      m_presets.push_back (swapData->get<0>());
+      onChange();
+    }, [ = ] (UNDO::Command::State) mutable
+    {
+      m_presets.erase (m_presets.end());
```
hhoegelo: will crash

PR #187, opened 2018-07-27
Feature change file selection layout

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/file/FileDialogLayout.cpp
```diff
@@ -72,6 +76,9 @@ bool FileDialogLayout::onButton(int i, bool down, ButtonModifiers modifiers)
         if(fileCount > 0)
           overlayInfo();
         return true;
+      case BUTTON_A:
+        hwui->undoableSetFocusAndMode({UIFocus::Setup, UIMode::Select});
```
hhoegelo: better use the current HWUI-Focus instead of hard coded 'Setup'

PR #192, opened 2018-07-27
moved mc carousel 1 pixel down

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ModulateableParameterLayouts.cpp
```diff
@@ -272,7 +272,7 @@ void ModulateableParameterSelectLayout2::setMode (Mode desiredMode)
       m_modeOverlay->addControl(new MCPositionLabel(Rect(0, BUTTON_VALUE_Y_POSITION, 64, 12)));
       m_modeOverlay->addControl(new MCAmountLabel(Rect(131, BUTTON_VALUE_Y_POSITION, 58, 12)));
 
-      setCarousel(new ModulationCarousel(ModulationCarousel::Mode::None, Rect(195, 0, 58, 64)));
```
hhoegelo: nitpicking, but now bottom is at 65 which is out of the display. In fact it is no problem, as the nested controls are only 62 in height, but I think it would be better to have the height of 62 here as well.

PR #195, opened 2018-07-27
added functionality to insert bank in clusters

schlenkibus:
We have a problem making this feature look good. Because we still snap to the (way too fine) grid, banks are in weird positions if clustered. That would call for a more sophisticated algorithm to determine the width and height of "InsertTapes".


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/DragProxy.java
```diff
@@ -21,7 +21,13 @@
 
 public class DragProxy extends OverlayControl {
 
-	class FoundControl {
+	private Position lastMousePos;
```
hhoegelo: better use PointerState.get().getPosition()

PR #198, opened 2018-08-01
Feature locksymbol changes

schlenkibus:


playground/src/proxies/hwui/panel-unit/boled/undo/UndoIndicator.cpp
```diff
@@ -0,0 +1,22 @@
+#include <Application.h>
+#include "UndoIndicator.h"
+#include <http/UndoScope.h>
+
+UndoIndicator::UndoIndicator(const Rect &pos) : Label("\ue20b",
+                                                      [=]() {
```
hhoegelo: consider Rect(pos.getLeft(), pos.getTop(), 7, 8)

PR #201, opened 2018-08-02
added parameter recall to loaded value

schlenkibus:


playground/src/presets/PresetManager.cpp
```diff
@@ -589,6 +589,17 @@ bool PresetManager::isLoading() const
   return m_loading.isLocked();
 }
 
+PresetManager::tPresetPtr PresetManager::getLoadedPreset() {
```
hhoegelo: consider: findPreset(eb->getUUID....);

playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
```diff
@@ -73,6 +96,20 @@ bool ParameterLayout2::onRotary (int inc, ButtonModifiers modifiers)
...
 }
 
+void ParameterLayout2::handlePresetValueRecall() {
+  if(isCurrentParamDiffFromLoaded()) {
+    getOLEDProxy().setOverlay(new ParameterRecallLayout2());
+  }
+}
+
+bool ParameterLayout2::isCurrentParamDiffFromLoaded() const {
+  if(auto currentParam = getCurrentEditParameter()) {
```
hhoegelo: maybe use the getOriginalParameter function instead?

PR #239, opened 2018-08-16
Feature cluster sort

schlenkibus:
#220


playground/src/presets/PresetManager.h
```diff
@@ -52,6 +52,7 @@ class PresetManager : public ContentSection
     tBankPtr getBank (int pos);
     const tBankPtr getBank (int pos) const;
     const vector<tBankPtr> &getBanks () const;
+    const vector<tBankPtr> getBanksIf (std::function<bool(const tBankPtr&)> predicate) const;
```
hhoegelo: do we still need this method?

playground/src/presets/PresetBank.cpp
```diff
@@ -900,10 +893,13 @@ const Glib::ustring PresetBank::directionEnumToString(AttachmentDirection direct
   static_assert(AttachmentDirection::none == 0, "Nicht den Enum ändern ohne diese Funktion und Java Seite zu ändern!");
   static_assert(AttachmentDirection::top == 1, "Nicht den Enum ändern ohne diese Funktion und Java Seite zu ändern!");
   static_assert(AttachmentDirection::left == 2, "Nicht den Enum ändern ohne diese Funktion und Java Seite zu ändern!");
+  static_assert(AttachmentDirection::count == 3, "Nicht den Enum ändern ohne diese Funktion und Java Seite zu ändern!");
 
   switch(direction)
   {
-
+    case count:
+      assert(false);
```
hhoegelo: consider g_warn_if_reached()

playground/src/presets/BankSorter.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "PresetManager.h"
+
+class PresetBank;
+
+class BankSorter {
```
hhoegelo: Do we need the BankSorter? In fact, it could all be a static method of ClusterEnforcement, couldn't it?

playground/src/presets/BankActions.cpp
```diff
@@ -978,6 +978,10 @@ BankActions::BankActions(PresetManager &presetManager) :
       bank->undoableMovePosition(transaction, x, y);
     }
   });
+
+  addAction("sort-bank-numbers", [&](auto request) mutable {
+    BankSorter::sort();
```
hhoegelo: Huh?

PR #241, opened 2018-08-16
Add systemd services as part of the project

pascalhuerst:
Systemd services belong to the project and not to the build environment.

This adds all necessary changes in order to have *.service files installed
with the project. Also installation target nonlinear/playground is part
of the project, rather than set by a command line argument.

Removed some unnecessary whitespaces.

### DIFF
--- a/bbbb/CMakeLists.txt
+++ b/bbbb/CMakeLists.txt
@@ -37,9 +37,13 @@ IF(DEV_PC)
   set(LIBS ${LIBS} gtkmm-3.0 gdkmm-3.0)
 ENDIF(DEV_PC)
 
+IF(NOT DEFINED SYSTEMD_CONFIGURATION_FILES_DIR)
+  set(SYSTEMD_CONFIGURATION_FILES_DIR "${CMAKE_INSTALL_PREFIX}/etc/systemd/system")
+ENDIF()
+
 function(addLib name)
   pkg_check_modules(${name} REQUIRED ${name})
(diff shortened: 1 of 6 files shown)


bbbb/CMakeLists.txt
```diff
@@ -74,8 +78,15 @@ foreach(lib ${LIBS})
   linkLib(${lib})
 endforeach(lib)
 
+configure_file(${PROJECT_SOURCE_DIR}/systemd/bbbb.service.in
+	${PROJECT_BINARY_DIR}/systemd/bbbb.service @ONLY)
+
+install(FILES ${PROJECT_BINARY_DIR}/systemd/bbbb.service
```
hhoegelo: don't we need sym links to the .service files in "multi-user..." sub directory?

pascalhuerst: Yes we do. But I think it's not the job of the installation to decide, whether we enable the service or not. Instead, I want to use the first-boot feature from systemd to setup stuff like that in the future:
https://www.freedesktop.org/software/systemd/man/systemd-firstboot.html

hhoegelo: OK, I don't understand. Just wanted to be sure, you thought about it.

pascalhuerst: I thought about it. But for the images we build using buildroot, the symlinks are still in place in fs-overlay. This will change, once we use yocto to build our bbb images.

PR #243, opened 2018-08-16
Fix export

schlenkibus:
Changed resulting filename because std::fs::copy has a problem with any ':' in filename. 
removed Homegrown solution for moving file and replaced by std::filesystem
it copies and removes original file because it has to do work across different file systems.  

FIXES: #237

### DIFF
--- a/playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
+++ b/playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
@@ -72,13 +72,13 @@ void ExportBackupEditor::exportBanks ()
 {
   Application::get ().stopWatchDog ();
   writeBackupFileXML ();
-  const auto humanDate = StringTools::replaceAll(
+  const auto humanDate = StringTools::replaceAll(StringTools::replaceAll(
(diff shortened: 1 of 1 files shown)


playground/src/proxies/hwui/panel-unit/boled/setup/ExportBackupEditor.cpp
```diff
@@ -69,17 +71,25 @@ void ExportBackupEditor::exportBanks ()
 {
   Application::get ().stopWatchDog ();
   writeBackupFileXML ();
-
-  auto timeStamp = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count();
-
-  auto humanDate = std::to_string(TimeTools::getDisplayStringFromStamp(timeStamp));
-
-  humanDate = StringTools::replaceAll(humanDate, " ", "-");
+  const auto humanDate = StringTools::replaceAll(StringTools::replaceAll(
```
hhoegelo: really ugly

schlenkibus: disagree

PR #257, opened 2018-08-17
For the upcomming yocto builds, we need to change the filename

pascalhuerst:
For yocto builds:

hostapd.conf -> hostapt.nonlinear.conf


pascalhuerst:
This does not need to be merged right now. I just start collecting a series for the upcomming switch to yocto

pascalhuerst:
Actually I don't like this solution. Try to find another one....

PR #275, opened 2018-08-22
added simple algorithm for calculating bank position

schlenkibus:
#253


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/presets/BankContextMenu.java
```diff
@@ -157,6 +159,24 @@ public Control click(Position eventPoint) {
...
+	private void createNewBank() {
+		List<Bank> banks = NonMaps.get().getNonLinearWorld().getPresetManager().getBanks();
+
+		NonPosition np = new NonPosition();
+		if(!banks.isEmpty()) {
+			Bank rightMost = banks.get(0);
+			for(Bank b : banks)
+			{
+				double currentX = b.getNonPosition().getLeft();
+				double rightMostX = rightMost.getNonPosition().getTop();
```
hhoegelo: getLeft() instead of getTop(), no?

PR #343, opened 2018-08-30
added bank changed connection + refactor

schlenkibus:
#340

### DIFF
--- a/playground/src/proxies/hwui/base-unit/soled/BanksLayout.cpp
+++ b/playground/src/proxies/hwui/base-unit/soled/BanksLayout.cpp
@@ -38,7 +38,8 @@ void BanksLayout::onBankSelected(shared_ptr<PresetBank> bank)
 {
   m_bank = bank;
   m_bankconnection.disconnect();
-  m_bankconnection = bank->onBankChanged(sigc::mem_fun(this, &BanksLayout::onBankChanged));
+  if(bank)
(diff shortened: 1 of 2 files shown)


playground/src/proxies/hwui/base-unit/soled/BanksLayout.cpp
```diff
@@ -36,11 +36,19 @@ BanksLayout::~BanksLayout ()
 
 void BanksLayout::onBankSelected(shared_ptr<PresetBank> bank)
 {
+  m_bank = bank;
+  m_bankconnection.disconnect();
+  m_bankconnection = bank->onBankChanged(sigc::mem_fun(this, &BanksLayout::onBankChanged));
```
hhoegelo: I think this will crash when the last bank is deleted.

playground/src/proxies/hwui/base-unit/soled/BanksLayout.h
```diff
@@ -19,9 +19,14 @@ class BanksLayout : public DFBLayout
...
     void onAutoLoadSettingChanged(const Setting *s);
+    void onBankChanged();
 
     Label *m_number;
     Label *m_name;
+    sigc::connection m_bankconnection;
     DirectLoadIndicator *m_directLoad;
+    shared_ptr<PresetBank> m_bank;
+
+  void updateFromBank(const shared_ptr<PresetBank> &bank) const;
```
hhoegelo: I would really like to have member functions and fields sorted in blocks.

PR #344, opened 2018-08-30
added TextCropper that takes control and text and shortens if neccess…

schlenkibus:
…ary, -> extracted. also uses cropper in bank and preset soled layout

#339

### DIFF
--- a/playground/src/proxies/hwui/TextCropper.cpp
+++ b/playground/src/proxies/hwui/TextCropper.cpp
@@ -1,18 +1,31 @@
 #include <proxies/hwui/controls/Control.h>
 #include <utility>
+#include <testing/TestDriver.h>
+#include <proxies/hwui/controls/Label.h>
 #include "TextCropper.h"
+#include "Oleds.h"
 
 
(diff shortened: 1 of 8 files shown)


playground/src/proxies/hwui/TextCropper.cpp
```diff
@@ -0,0 +1,76 @@
+#include <proxies/hwui/controls/Control.h>
+#include <utility>
```
hhoegelo: Here, I think it is time for ........... TESTING, woohoo!

I would really like to see some test cases here, as the code is not too easy to read, it cares for edge cases and is a candidate for frequent change requests etc. Please have a look at Dimension.cpp for how to add a TestDriver. 

In short:
Place a "static TestDriver<TextCropper> tests;" in your cpp file, add a static TextCropper::registerTests() function like this:

void TextCropper::registerTests() 
{
  g_test_add_func ("/TextCropper/name-of-the-test", []()
  {
    g_assert(TextCropper::crop("Foo") == "Fo");
    g_assert(TextCropper::crop("FooBar") == "FoBa");
  });
};

PR #362, opened 2018-09-12
changed on button and on rotary behaviour for recall layout. so that …

schlenkibus:
…inc to max and dec to min works with shift modifier

#361

### DIFF
--- a/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
+++ b/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
@@ -262,6 +262,11 @@ bool ParameterRecallLayout2::onButton(int i, bool down, ButtonModifiers modifier
   if(i == BUTTON_C || i == BUTTON_B)
     return true;
 
+  if(i == BUTTON_INC || i == BUTTON_DEC) {
+    getOLEDProxy().resetOverlay();
+    return super::onButton(i, down, modifiers);
+  }
+
   getOLEDProxy().resetOverlay();
   return false;
 }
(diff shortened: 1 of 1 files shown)


playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
```diff
@@ -262,14 +262,19 @@ bool ParameterRecallLayout2::onButton(int i, bool down, ButtonModifiers modifier
...
+
   getOLEDProxy().resetOverlay();
   return false;
 }
 
 bool ParameterRecallLayout2::onRotary(int inc, ButtonModifiers modifiers)
 {
   getOLEDProxy().resetOverlay();
-  return false;
+  return super::onRotary(inc, modifiers);
```
hhoegelo: 'This' is the layout you reset (destruct) just a line above - so super::onRotary will be called on a destructed object, no?

playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
```diff
@@ -262,14 +262,19 @@ bool ParameterRecallLayout2::onButton(int i, bool down, ButtonModifiers modifier
   if(i == BUTTON_C || i == BUTTON_B)
     return true;
 
+  if(i == BUTTON_INC || i == BUTTON_DEC) {
+    getOLEDProxy().resetOverlay();
+    return super::onButton(i, down, modifiers);
```
hhoegelo: 'This' is the layout you reset (destruct) just a line above - so super::onRotary will be called on a destructed object, no?

PR #378, opened 2018-09-12
added check if current bank is in cluster and enabling/disabling inse…

schlenkibus:
…rt Tape accordingly
#371

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
@@ -39,7 +39,21 @@ public class Tape extends MapsControl {
 
 	@Override
 	public boolean isVisible() {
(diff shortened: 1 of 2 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
```diff
@@ -39,7 +39,21 @@ public Bank getParent() {
...
+	}
+
+	private boolean currentDraggedBankNotInCluster() {
+		Overlay o = getNonMaps().getNonLinearWorld().getViewport().getOverlay();
+
+		if(o.isCurrentlyDraggingATypeOf(Bank.class.getName())) {
+			for (DragProxy d : o.getDragProxies()) {
+				if(d.getOrigin() instanceof Bank) {
+					Bank b = (Bank) d.getOrigin();
+					return !b.hasSlaves() && b.getMaster() != null;
```
hhoegelo: why not use isInCluster() ?

PR #398, opened 2018-09-19
added signal for onRestore -> triggers PresetList setup

schlenkibus:
#389


hhoegelo:
we found and fixed a bug in Throttler preventing the display from updating reliably. Most probably this bug was causing this issue, too. This means, we do not need this workaround anymore.

PR #406, opened 2018-09-21
added POD StoreModeData that tracks the current selected Preset and B…

schlenkibus:
#402

### DIFF
--- a/playground/src/presets/PresetManager.cpp
+++ b/playground/src/presets/PresetManager.cpp
@@ -49,7 +49,8 @@ PresetManager::PresetManager(UpdateDocumentContributor *parent) :
     m_editBuffer(EditBuffer::createEditBuffer(this)),
     m_initSound(EditBuffer::createEditBuffer(this)),
     m_saveJob(bind(mem_fun(this, &PresetManager::doSaveTask))),
-    m_autoLoadThrottler(std::chrono::milliseconds(500))
+    m_autoLoadThrottler(std::chrono::milliseconds(500)),
+    m_storeModeData(nullptr)
 {
   m_actionManagers.emplace_back(new PresetManagerActions(*this));
   m_actionManagers.emplace_back(new BankActions(*this));
--- a/playground/src/presets/StoreModeData.cpp
+++ b/playground/src/presets/StoreModeData.cpp
@@ -1,0 +1,14 @@
+#include <Application.h>
+#include "StoreModeData.h"
+#include "PresetManager.h"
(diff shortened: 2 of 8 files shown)


playground/src/presets/PresetManager.cpp
```diff
@@ -49,7 +49,8 @@ PresetManager::PresetManager(UpdateDocumentContributor *parent) :
     m_editBuffer(EditBuffer::createEditBuffer(this)),
     m_initSound(EditBuffer::createEditBuffer(this)),
     m_saveJob(bind(mem_fun(this, &PresetManager::doSaveTask))),
-    m_autoLoadThrottler(std::chrono::milliseconds(500))
+    m_autoLoadThrottler(std::chrono::milliseconds(500)),
+    m_storeModeData(nullptr)
```
hhoegelo: The PresetManager doesn't use its member at all.
Public members for classes with invariants are not acceptable anyway.
The m_storeModeData is kind of static, globally shared - so why don't express this explicitly by adding static std::unique_ptr<...> s_storeModeData in PresetManagerLayout.cpp? The PresetManagerLayout is the only context that uses the POD, it could pass a pointer to the 'PresetListSelectStorePosition'-children on creation.

playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/PresetListSelectStorePosition.h
```diff
@@ -31,8 +30,8 @@ class PresetListSelectStorePosition : public PresetListBase
 
     static constexpr int invalidIndex = -1;
 
-    int m_presetPosition = invalidIndex;
-    int m_bankPosition = invalidIndex;
+    int* m_presetPosition;
```
hhoegelo: better store a pointer to StoreModeData instead of ptr to members of store mode data

playground/src/presets/StoreModeData.cpp
```diff
@@ -0,0 +1,15 @@
+#include <Application.h>
+#include "StoreModeData.h"
+#include "PresetManager.h"
+#include "PresetBank.h"
+
+StoreModeData::StoreModeData() {
+  if(auto pm = Application::get().getPresetManager()) {
+    if(auto banks = pm->getBanks(); !banks.empty())
+      bankPos = static_cast<int>(std::find(banks.begin(), banks.end(), pm->getSelectedBank()) - banks.begin());
```
hhoegelo: consider using std::distance

PR #426, opened 2018-09-27
changed mc ammounts in diff to -100..100 and added %

schlenkibus:
#388

### DIFF
--- a/playground/src/parameters/ModulateableParameter.cpp
+++ b/playground/src/parameters/ModulateableParameter.cpp
@@ -225,7 +225,8 @@ void ModulateableParameter::writeDifferences(Writer& writer, Parameter* other) c
 
   if(getModulationAmount() != pOther->getModulationAmount())
   {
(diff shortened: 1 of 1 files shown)


playground/src/parameters/ModulateableParameter.cpp
```diff
@@ -225,7 +225,8 @@ void ModulateableParameter::writeDifferences(Writer& writer, Parameter* other) c
 
   if(getModulationAmount() != pOther->getModulationAmount())
   {
-    writer.writeTextElement("mc-amount", "", Attribute("a", getModulationAmount()), Attribute("b", pOther->getModulationAmount()));
```
hhoegelo: Please use the according ScaleConverter to translate from control position to display string:
auto c = ScaleConverter::get<LinearBipolar100PercentScaleConverter>(); c.getDimension().stringize(c.controlPositionToDisplay(foo));

PR #445, opened 2018-10-10
412

schlenkibus:
#412

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/ParameterClippingLabel.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/ParameterClippingLabel.java
@@ -1,0 +1,28 @@
+package com.nonlinearlabs.NonMaps.client.world.overlay.belt.parameters;
+
+import com.google.gwt.canvas.dom.client.Context2d;
+import com.nonlinearlabs.NonMaps.client.Tracer;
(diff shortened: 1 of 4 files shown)


playground/src/parameters/ModulateableParameter.cpp
```diff
@@ -214,29 +218,29 @@ void ModulateableParameter::undoableIncrementMCAmount(UNDO::Scope::tTransactionP
 void ModulateableParameter::writeDocProperties(Writer &writer, tUpdateID knownRevision) const
 {
   Parameter::writeDocProperties(writer, knownRevision);
+  auto modRange = getModulationRange(true);
```
hhoegelo: modRange doesn't seem to be used

NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/ParameterClippingLabel.java
```diff
@@ -0,0 +1,28 @@
...
+	String m_text = "";
+	
+	ParameterClippingLabel(OverlayLayout parent, BeltParameterLayout.Mode mode) {
+		super(parent);
+		setFontHeightInMM(9);
+		setClipping(false);
+	}
+
+	public void setClipping(boolean b) {
+		Tracer.log("setting clip to " + b);
```
hhoegelo: shouldn't setClipping call invalidate if the value changes?
Also, I would prefer storing a boolean member and have getDrawText() something like 'return clipping ? "!" : ""';.

PR #508, opened 2018-10-30
504

schlenkibus:
removing spaces from display strings in mod Carousel, now !100.0%F fits!

#504


playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/CurrentModulatedValueLabel.cpp
```diff
@@ -10,29 +10,33 @@
...
 {
 }
 
-void CurrentModulatedValueLabel::updateText (MacroControlParameter *mcParam, ModulateableParameter *modulatedParam)
+void CurrentModulatedValueLabel::updateText(MacroControlParameter *mcParam, ModulateableParameter *modulatedParam)
 {
-  if (isHighlight () && Application::get ().getHWUI ()->isModifierSet(ButtonModifier::FINE))
-    setText (modulatedParam->getDisplayString () + " F", 2);
+  auto displayString = modulatedParam->getDisplayString();
+  std::string displayStringWithoutSpaces = displayString;
```
hhoegelo: Consider introducing a service method in namespace StringTools (src/tools/StringTools.h) "removeSpaces()" and use this instead of duplicating the code three times.

PR #543, opened 2018-11-12
clipping indicator is set to false every time the belt is redrawn to …

schlenkibus:
…prevent sticky and buggy indicator

#542


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/BeltParameterLayout.java
```diff
@@ -449,8 +449,12 @@ public String getCurrentValuesName(double width) {
 	}
 
 	public String[] getDecoratedValue(boolean b) {
+		
+		mcUpperClip.setClipping(false);
```
hhoegelo: let's discuss this soon - I don't think a method 'getDecoratedValue()'  should change the state of something. setClipping might (and probably will) invalidate the UI, causing a redraw. Redrawing will most likely call getDecoratedValue() - an infinite loop is born. I would prefer to have no setClipping() method at all, but calculate the clipping state on each redraw.

PR #549, opened 2018-11-13
changed quantization of Upper/Lower Bound to 1000 to allow for fine

schlenkibus:
#547


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/BeltParameterLayout.java
```diff
@@ -317,6 +319,9 @@ public void setValueChanger(IncrementalChanger changer) {
 
 	@Override
 	public Control mouseDrag(Position oldPoint, Position newPoint, boolean fine) {
+		
+		fine = NonMaps.get().getNonLinearWorld().isShiftDown();
```
hhoegelo: whats wrong with the fine argument? If Shift is supposed to be the 'fine' modifier, shouldn't we respect it at the origin of the fine arg?

schlenkibus: True, i removed it

PR #550, opened 2018-11-13
changed limit controls to use real denominator instead of some weird …

schlenkibus:
…smaller than 8000 calculation stuff 

#546


playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/LowerModulationBoundControl.cpp
```diff
@@ -94,8 +94,9 @@ double LowerModulationBoundControl::calcDominator(const ButtonModifiers &modifie
                                                   const ModulateableParameter *modulatedParam) const
 {
   auto fine = modifiers[FINE];
-  auto fineDenominator = modulatedParam->getModulationAmountFineDenominator();
+  auto fineDenominator = modulatedParam->getValue().getFineDenominator();
   auto coarseDenominator = modulatedParam->getModulationAmountCoarseDenominator();
+  DebugLevel::warning("Coarse: ", coarseDenominator);
```
hhoegelo: pls remove the warning if not needed anymore

PR #560, opened 2018-11-20
done the default value setting for upper/lower limits in webui and boled

schlenkibus:
#555


playground/src/functional/LambdaFactory.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
+#include <Application.h>
+#include <parameters/ModulateableParameter.h>
+#include <presets/PresetManager.h>
+#include <presets/EditBuffer.h>
+
+class LambdaFactory
+{
+ public:
+  static auto setCurrentModulatedParamAmountToZero()
```
hhoegelo: I don't get why you have a factory creating lambdas which you immediately invoke afterwards. What is the benefit compared to just having functions?  
I would prefer ModulateableParameter::resetModAmount() and ModulateableParameter::setDefault()

PR #581, opened 2018-12-06
Minimize / Full Size All

schlenkibus:
added minimize all and full size all (banks) in bank context menu, default action is minimize all, if any bank is maximized, matter to change?
#579


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -97,6 +97,20 @@ public void endStoreSelectMode() {
...
 		}
 	}
+	
+	public void setAllBanksMinimizeState(boolean min) {
+		for(Bank b: getBanks()) {
+			b.setMinimized(min);
+		}
+	}
+	
+	public boolean isAnyBankMinimized() {
```
hhoegelo: the implemented algorithm does not match to its name - it actually implements: areAllBanksMinimized().

PR #611, opened 2018-12-18
added floating window close button color on mousedown, setup and undo…

schlenkibus:
… now same color

#476


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/FloatingWindowHeader.java
```diff
@@ -51,19 +57,37 @@ private void drawX(Context2d ctx) {
...
 				ctx.setLineCap(LineCap.SQUARE);
 				ctx.setLineJoin(LineJoin.MITER);
 
 				ctx.stroke();
 			}
-
+			
 			@Override
 			public Control mouseDown(Position eventPoint) {
+				mouseDown = true;
```
hhoegelo: you can get rid of the additional member and just invalidate() in all of the mouse handlers. Then, in drawX, use "isCaptureControl()" instead of 'mousedown'.

PR #614, opened 2018-12-18
Feature wi fi setting

schlenkibus:
#573

### DIFF
--- a/playground/src/device-settings/WifiSetting.cpp
+++ b/playground/src/device-settings/WifiSetting.cpp
@@ -1,0 +1,31 @@
+#include <tools/SpawnCommandLine.h>
+#include "WifiSetting.h"
+
+WifiSetting::WifiSetting(Settings &settings)
+    : BooleanSetting(settings, WifiSetting::pollAccessPointRunning())
+{
+}
+
+WifiSetting::~WifiSetting() = default;
+
+bool WifiSetting::set(Glib::ustring value)
+{
+  auto val = value.find("on") != Glib::ustring::npos;
+  
+  if(val)
+    SpawnCommandLine cmd("systemctl enable accesspoint && systemctl start accesspoint");
+  else
+    SpawnCommandLine cmd("systemctl disable accesspoint && systemctl stop accesspoint");
+
+  return BooleanSetting::set(val ? BooleanSetting::tEnum::BOOLEAN_SETTING_TRUE
+                                 : BooleanSetting::tEnum::BOOLEAN_SETTING_FALSE);
+}
+
+bool WifiSetting::pollAccessPointRunning()
+{
+#ifdef DEV_PC
+  return true;
+#endif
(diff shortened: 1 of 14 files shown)


playground/src/device-settings/WifiSetting.cpp
```diff
@@ -0,0 +1,41 @@
...
+                                : BooleanSetting::tEnum::BOOLEAN_SETTING_FALSE);
+}
+
+bool WifiSetting::pollAccessPointRunning()
+{
+#ifdef DEV_PC
+  return true;
+#endif
+  SpawnCommandLine cmd("systemctl is-active accesspoint");
+  return cmd.getStdOutput().find("active") != Glib::ustring::npos;
```
hhoegelo: inactive services do return "inactive" where the substring "active" will be found

hhoegelo: better use the return value (0 means service is active)

playground/src/proxies/hwui/panel-unit/boled/setup/WiFiSettingEditor.cpp
```diff
@@ -0,0 +1,24 @@
...
+WiFiSettingEditor::WiFiSettingEditor() {
+    Application::get().getSettings()->getSetting<WifiSetting>()->onChange(
+            mem_fun(this, &WiFiSettingEditor::onSettingChanged));
+}
+
+WiFiSettingEditor::~WiFiSettingEditor() = default;
+
+void WiFiSettingEditor::incSetting(int inc) {
+    auto setting = Application::get().getSettings()->getSetting<WifiSetting>();
+    setting->set(!setting->get() ? "on" : "off");
```
hhoegelo: better use setting->inc(1, true)

playground/src/device-settings/WifiSetting.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+
+auto printRet = [](SpawnCommandLine & cmd)
+{
+  DebugLevel::warning("COut:", cmd.getStdOutput(), "CError:", cmd.getStdError(), "Exit Status:", cmd.getExitStatus());
+};
+
+WifiSetting::~WifiSetting() = default;
+
+bool WifiSetting::set(Glib::ustring value)
```
hhoegelo: better override EnumSetting::set(tEnum m) like this:

WifiSetting::set(tEnum m)
{
 base_type::set(m);

if(get())
doThis();
else
doThat();
}

playground/src/proxies/hwui/TextEditUsageMode.cpp
```diff
@@ -410,3 +410,8 @@ void TextEditUsageMode::registerTests()
     g_assert_true(mode.m_layout == Layout::Shift);
   });
 }
+
+void TextEditUsageMode::setText(ustring newText) {
```
hhoegelo: please remove unused artefact

PR #652, opened 2019-01-03
640

schlenkibus:
#640


playground/src/parameters/names/ParameterDB.cpp
```diff
@@ -66,7 +66,10 @@ tControlPositionValue ParameterDB::parseSignalPathIndication(const std::string &
 {
   try
   {
-    return stod(c);
```
hhoegelo: seems wrong to me: before, a non-existent signal path indication value was indicated by std::numeric_limits<tControlPositionValue>::max(). Now, the function will return '0' instead.

PR #682, opened 2019-01-16
make use of signal flow indication instead of comparing against defau…

hhoegelo:
…lt value


hhoegelo:
Has to be reviewed and tested carefully, as I am not that familiar with the signal flow indication feature. Still, I think I found some bugs (mostly comparing values against their default instead of the 'Inactive CP').

PR #684, opened 2019-01-16
activating/deactivating AP takes quite some time, UI is stuck during …

hhoegelo:
…operation, watch dog may be angry. So better do blocking wifi configuration in background


hhoegelo:
fixes #573

PR #685, opened 2019-01-16
calling the animation callback from dtor - which will switch the HWUI…

hhoegelo:
…::focusAndMode is just wrong


hhoegelo:
fixes #568

PR #691, opened 2019-01-22
245

schlenkibus:
#245

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
@@ -1,6 +1,5 @@
 package com.nonlinearlabs.NonMaps.client.world.maps.presets.bank;
 
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/bank/Tape.java
```diff
@@ -41,25 +43,79 @@ public Bank getParent() {
...
 		}
 
 		return false;
 	}
+	
+	static private NonPosition getCompareablePosition(MapsControl c) {
+		return c.getNonPositionRelativeToParent((MapsLayout)NonMaps.get().getNonLinearWorld().getPresetManager()).getCenterPoint();
+	}
+
+	private Tape getTopMostIntersectingTape(Tape other) {
```
hhoegelo: Consider using getPixRect as comparable rects. It would make code easier and probably improve performance as the rect already exists as member of each control.

schlenkibus:
I refactored the whole Foo around getTopMostTape() function and was able to remove most of the confusing / misleading accompanying functions. 
Still I have to do some hacking around how to get the Pix-Rect of a dragged Control.

PR #735, opened 2019-02-18
Repaired Recall and Compare with Init-Sound

schlenkibus:
cleanup code paths in compare / recall for the case that no original preset exists -> init sound, now if no original preset exists the value with which current value is being compared is the default value. if no previous preset exists and editbuffer parameter is different than default -> current parameter changed

### DIFF
--- a/playground/src/parameters/Parameter.cpp
+++ b/playground/src/parameters/Parameter.cpp
@@ -235,13 +235,11 @@ bool Parameter::isChangedFromLoaded() const
 
   if(auto originalParameter = getOriginalParameter())
   {
-    const auto rawOld = originalParameter->getValue();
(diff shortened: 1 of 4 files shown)


playground/src/parameters/Parameter.cpp
```diff
@@ -229,15 +229,18 @@ PresetParameter *Parameter::getOriginalParameter() const
 
 bool Parameter::isChangedFromLoaded() const
 {
+  const auto rawNow = getControlPositionValue();
+  const auto epsilon = 0.5 / getValue().getFineDenominator();
+  DebugLevel::warning("Using", epsilon, "as epsilon for Parameter::isChangedFromLoaded!", getLongName());
```
hhoegelo: I think the level "gassy" is enough for this message?

PR #742, opened 2019-02-20
Sorry for Huge Commit: Removed Compare Button from WebUI Belt and add…

schlenkibus:
…ed 2 Settings that indicate whether or not the changed parameters should be highlighted in WebUI. Setup has the option to toggle global persistent setting and a force setting that is enabled if shift is pressed in HWUI PresetMode or RecallLayout is open on changed Parameter, also removed changed border for HW-Sources

Fixes:
#740
#738 
#739


NonMaps/src/com/nonlinearlabs/NonMaps/client/presenters/LocalSettingsProvider.java
```diff
@@ -66,6 +68,13 @@ private LocalSettingsProvider() {
 			notifyClients();
 			return true;
 		});
+		
+		s.forceHighlightChanged.onChange(t -> {
+			NonMaps.get().getNonLinearWorld().getParameterEditor().getChildren().forEach(control -> {
+				control.invalidate(Control.INVALIDATION_FLAG_UI_CHANGED);
```
hhoegelo: you don't need to call invalidate on each child. Invalidate anyway bubbles up to the world and invalidates this globally. So NonMaps.get().getNonLinearWorld().invalidate() would have the same effect with less overhead.

NonMaps/src/com/nonlinearlabs/NonMaps/client/useCases/SystemSettings.java
```diff
@@ -96,4 +96,10 @@ public void setPresetGlitchSurpression(BooleanValues c) {
 		Setup.get().systemSettings.presetGlitchSuppression.setValue(c);
 		NonMaps.theMaps.getServerProxy().setSetting("PresetGlitchSuppression", netify(c.name()));
 	}
+
+	public Object setHighlightChangedParameters(BooleanValues c) {
```
hhoegelo: Whats the return value for?

PR #760, opened 2019-02-22
758

schlenkibus:
#758

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
@@ -14,6 +14,7 @@ import com.nonlinearlabs.NonMaps.client.NonMaps;
 import com.nonlinearlabs.NonMaps.client.Renameable;
 import com.nonlinearlabs.NonMaps.client.ServerProxy;
 import com.nonlinearlabs.NonMaps.client.StoreSelectMode;
+import com.nonlinearlabs.NonMaps.client.dataModel.editBuffer.EditBufferModel;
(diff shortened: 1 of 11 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -821,28 +822,31 @@ public boolean isEmpty() {
...
-
-			if (b != null) {
-				NumberFormat f = NumberFormat.getFormat("000");
-				String ret = b.getOrderNumber() + "-" + f.format(p.getNumber());
-
-				if (NonMaps.theMaps.getNonLinearWorld().getParameterEditor().isModified())
-					ret += " *";
+	private String getBaseLoadedPresetNumberString() {		
+		EditBufferModel eb = EditBufferModel.get();
+		String ret = "";
```
hhoegelo: In my opinion all this logic could be moved into the presenter. The UI element should get rid of all logic, if possible.

NonMaps/src/com/nonlinearlabs/NonMaps/client/world/maps/presets/PresetManager.java
```diff
@@ -821,28 +822,31 @@ public boolean isEmpty() {
...
-			}
-		} else {
-			String ret = "Init";
-			if (NonMaps.theMaps.getNonLinearWorld().getParameterEditor().isModified())
-				ret += " *";
+		return ret;
+	}
+	
+	public String getLoadedPresetNumberString() {
+		boolean mod = NonMaps.theMaps.getNonLinearWorld().getParameterEditor().isModified();
```
hhoegelo: same here, pls move logic in presenter

PR #762, opened 2019-02-22
757

schlenkibus:
#757

### DIFF
--- a/playground/src/presets/UndoableVector.h
+++ b/playground/src/presets/UndoableVector.h
@@ -57,7 +57,11 @@ template <typename Element> class UndoableVector
 
   Element *at(size_t idx) const
   {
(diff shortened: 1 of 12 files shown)


playground/src/presets/UndoableVector.h
```diff
@@ -57,7 +57,11 @@ template <typename Element> class UndoableVector
 
   Element *at(size_t idx) const
   {
-    return getElements().at(idx).get();
```
hhoegelo: do we need this? Or does the other change fix the problem already? If so, I would like to keep the exception, so maybe we find some more errors like the origin.

schlenkibus: I would like to discus this with you next week in Person! I will leave it as is and merge a test branch for Stephan.

PR #768, opened 2019-02-22
761

schlenkibus:
#761 
#769 
#770 
#771 
#772 
#773


playground/src/parameters/MacroControlParameter.cpp
```diff
@@ -125,7 +125,7 @@ void MacroControlParameter::propagateMCChangeToMCViews(const Initiator &initiati
 
   if(valueD != lastBroadcastedControlPosition)
   {
-    const auto str = "MCVIEW&ID="s.append(idString).append("&VAL=").append(value).append("&UUID=").append(uuid);
+    const auto str = "MCVIEW&ID="s.append(idString).append("&VAL=").append(value).append("&UUID=").append(uuid).append("&NAME=").append(getGivenName());
```
hhoegelo: Consider UNDO::buildString("MCVIEW&ID=", idString, "&VAL=", value ...). It could be moved into StringTools, though.
Or, we could have a class URIBuilder that uses a stringstream internally and has a public API like this: 

class URIBuilder
{
 public:
  URIBuilder(const std::string &baseURL);
  URIBuilder &operator << (const std::pair<std::string, std::string> &keyAndValue);
  std::string finalize();
};

playground/src/http/MCViewContentManager.cpp
```diff
@@ -10,6 +11,16 @@ void MCViewContentManager::connectWebSocket(SoupWebsocketConnection *connection)
 {
   g_signal_connect(connection, "message", G_CALLBACK(&MCViewContentManager::onWebSocketMessage), this);
   m_webSockets.emplace_back(std::make_shared<WebsocketConnection>(connection));
+  for(auto& param: Application::get().getPresetManager()->getEditBuffer()->getParameterGroupByID("MCs")->getParameters()) {
+      if(auto mc = dynamic_cast<MacroControlParameter*>(param)) {
+          const auto idString = to_string(mc->getID());
+          const auto valueD = mc->getValue().getClippedValue();
+          const auto value = to_string(valueD);
+          const auto uuid = "FORCE"s;
+          const auto str = "MCVIEW&ID="s.append(idString).append("&VAL=").append(value).append("&UUID=").append(uuid).append("&NAME=").append(mc->getGivenName());
```
hhoegelo: consider use of StringTools::...

PR #777, opened 2019-02-25
added 'generic' modal GWT dialog to still show the warning to user wh…

schlenkibus:
…ilst going around a chrome bug (72)

#775

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/tools/CallableVoid.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/tools/CallableVoid.java
@@ -1,0 +1,5 @@
+package com.nonlinearlabs.NonMaps.client.tools;
+
+public interface CallableVoid {
+	public void action();
+}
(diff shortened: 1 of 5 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/tools/CallableVoid.java
```diff
@@ -0,0 +1,5 @@
+package com.nonlinearlabs.NonMaps.client.tools;
+
```
hhoegelo: Consider interface Runnable

NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/presets/PresetManagerContextMenu.java
```diff
@@ -1,5 +1,7 @@
 package com.nonlinearlabs.NonMaps.client.world.overlay.belt.presets;
 
+import java.util.concurrent.Callable;
```
hhoegelo: does this even compile with GWT?

PR #804, opened 2019-03-07
Audio engine

hhoegelo:
This moves the audio engine from nlaudio to C15 project


hhoegelo:
I've fixed the midi issues.

MatthiasSeeber:
- MIDI messages (Note Off, Note On, Ctrl Change) seem valid now (Poly AT will be testable when a TCD Sender (LPC) is connected)
- the performance seems to have suffered: audio is glitchy, latency is noticeable, cpu load on mobile Bill: ~85%, (smaller latencies cause more glitches or x-run errors, glitches also occur at large latencies)
- test MIDI sequence now only manages to produce 7 out of 16 notes? (the remaining ones seem to be fired but make no sound)
- the following message is shown after starting the engine: "Could not set thread priority - consider 'sudo setcap 'cap_sys_nice=eip' <application>'"

MatthiasSeeber:
- after starting, an ALSA error is shown: "Error:  Alsa Error: Invalid argument" / is this a problem?
- Key Assignments are working
- Latency seems okay (I currently use -L 10 -P 20 in Release and there are no glitches or xrun errors)
- MIDI Input is now glitch-free as well
- performance on mobile bill (i3, 2.4 GHz - Release, -P 20 -L 10): ~ 60...70%
- readability of messages printed to terminal is not optimal (due to performance and info strings and many line breaks - e.g. when printing the TCD Log), but not much of an issue either (maybe, we can optimize this in the future?)
- so, from my perspective, this state is good to merge.. (I'm just not sure about the ALSA error)

hhoegelo:
The Alsa error is not a problem: I try to open the audio hardware in float format first, if this fails (tyhe rror is thrown), I try it with signed int 32, if this fails, I use signed int 16. As long as one of the 3 options works, we have no problem.

PR #808, opened 2019-03-15
add option -e for measuring the audio engine performance without in a

hhoegelo:
reproducible way


MatthiasSeeber:
I've tried several command line variations (with "-e", "-e 0", -"e 1" or without) and everytime I get this message and the program terminates:
Error:  Alsa Error: Invalid argument 
Info:  Alsa Frames per Period: 480 
Info:  Alsa Periods: 2 
Info:  Midi2Audio latency is: 10 ms. 
Press <RETURN> to close this window...
Am I doing something wrong?

audio-engine/CMakeLists.txt
```diff
@@ -24,7 +51,7 @@ endforeach(lib)
 
 include_directories(src)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -Wdouble-promotion")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -Wdouble-promotion -march=broadwell -Ofast")
```
AntonSchmied: if i leave out the flags after the word "promotion", only then i can measure the performance ... 
Is it possible that it is a bit to specific for a general test?

AntonSchmied:
`set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -Wdouble-promotion -march=broadwell -Ofast")`

If I leave out the the flags "-march=broadwell" and "-Ofast", only then I can test the performance.
Is this maybe too specific and not suitable for general tests?

MatthiasSeeber:
...what Anton is saying, now it's working.

PR #809, opened 2019-03-18
798

schlenkibus:
#798 
#810


playground/src/presets/EditBuffer.cpp
```diff
@@ -312,23 +317,41 @@ void EditBuffer::writeDocument(Writer &writer, tUpdateID knownRevision) const
...
+      { Attribute("selected-parameter", m_selectedParameter ? m_selectedParameter->getID() : 0),
+        Attribute("loaded-preset", getUUIDOfLastLoadedPreset().raw()), Attribute("loaded-presets-name", getName()),
+        Attribute("loaded-presets-bank-name", bankName), Attribute("preset-is-zombie", zombie),
+        Attribute("is-modified", m_isModified), Attribute("hash", getHash()), Attribute("changed", changed) },
+      [&]() {
+        if(changed)
+        {
+          super::writeDocument(writer, knownRevision);
+        }
+        if(auto originPreset = getOrigin())
```
hhoegelo: With this change you require the implementation here in the else branch to be in sync with the implementation in Preset::writeDetailDocument. 
How about this: if there is no origin preset, create a new (empty) preset on stack and write this? Would this work?

playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/AppendOverwriteInsertButtonMenu.cpp
```diff
@@ -198,6 +198,7 @@ Preset* AppendOverwriteInsertButtonMenu::overwritePreset(UNDO::Transaction* tran
     pm->selectBank(transaction, bank->getUuid());
   }
 
+  preset->onChange();
```
hhoegelo: this is most probably wrong: if onChange has to be called here, it will have to be called on undo/redo, too. So, one of the actions before has to call onChange from within an action.

playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/AppendOverwriteInsertButtonMenu.cpp
```diff
@@ -213,6 +214,7 @@ void AppendOverwriteInsertButtonMenu::insertPreset(Bank* bank, size_t pos, bool
 
   bank->selectPreset(scope->getTransaction(), preset->getUuid());
   pm->selectBank(transaction, bank->getUuid());
+  preset->onChange();
```
hhoegelo: same here, see comment above

playground/src/presets/EditBuffer.cpp
schlenkibus: This part is also hard for me to understand, but I would leave the logic as is, because filling the Preset with the defaults is not wanted in general, if the logic, so what values we are defaulting to, is approved and fixed, we can revisit this. 
But for now it would make the Preset more confusing and lead to bugs by initializing the Parameter Groups with unexpected Values.

PR #820, opened 2019-03-21
fixes 819

hhoegelo:


MatthiasSeeber:
Synth.cpp wants to include io/audio/AlsaAudioOutput.h, which is not present - so i cannot build.
In the given local directory, only AudioOutput.h, AudioOutputMock.h and AudioWriter.h are present. In the online repo (https://github.com/nonlinear-labs-dev/C15/tree/819/audio-engine/src/io/audio), it's the same..?
The commit features AudioOutput.h, but no files containing "Alsa" in their names.
I'm not sure if my checkout went wrong, or if files were renamed and not accomodated in the code, or if new files have been forgotten in the commit?

hhoegelo:
Sorry, added missing files

PR #824, opened 2019-03-21
The AlsaMidiInput and the AlsaAudioOutput classes had both code for p…

hhoegelo:
…rioritizing threads.

I refactored it to share code in class HighPriorityTask.
Also, I fixed setting high priorities not only to '50', but to the max available value.
Then, I found that conversion of float to 24 bit samples were missing the upper most bit and fixed it.

### DIFF
--- a/audio-engine/CMakeLists.txt
+++ b/audio-engine/CMakeLists.txt
@@ -49,7 +49,15 @@ endforeach(lib)
 
 include_directories(src)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -Wdouble-promotion")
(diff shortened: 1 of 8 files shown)


MatthiasSeeber:
- scheduling priority: 99
- x-runs still occur, but performance (~60%) seems fairly stable on mobile bill
- I didn't experiment much with the alsa args yet, though
- tested again with steinberg (32-bit) and ni traktor (24-bit) interfaces (traktor seems to perform not as good as steinberg and delivers less gain - certainly nothing of our responsibility?)
- so, a merge seems appropriate to me, shall I?

MatthiasSeeber:
Tested [823] against [master] by measuring output signal gain
- for 16 and 32 bit interfaces, no differenct is detected between both implementations
- for 24 bit, a signal gain of ~+6dB was measured, so the [823] implementation certainly is an improvement
- still, the redundancy of subtracting one for 16 bit levels seems odd to me

Issue #822, opened 2019-03-21
Bug Collection 1


schlenkibus:
- [x] Valuelabel in Parameter Belt not centered
- [x] Mod Src OG Label too narrow
- [x] Store Editbuffer with modulated Parameters does not remove changed indication in WebUI

PR #825, opened 2019-03-21
822

schlenkibus:
#822


NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
```diff
@@ -27,7 +27,11 @@ public BasicParameterModel getValue() {
 	}
 
 	public boolean isChanged() {
-		return value.value.getValue() != originalValue.getValue();
+		double roundedVal = (int) (value.value.getValue() * 1000);
+		roundedVal /= 1000.0;
```
hhoegelo: I don't think we need to divide by 1000 here. As far as I can see the result should be the same if you compare this ints.

PR #835, opened 2019-03-25
806

schlenkibus:
PhysicalControlParameter always says its not changedFromLoaded from preset now
#806

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/recall/ParameterRecallArea.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/recall/ParameterRecallArea.java
@@ -2,6 +2,7 @@ package com.nonlinearlabs.NonMaps.client.world.overlay.belt.parameters.recall;
 
 import com.nonlinearlabs.NonMaps.client.dataModel.editBuffer.BasicParameterModel;
 import com.nonlinearlabs.NonMaps.client.dataModel.editBuffer.EditBufferModel;
+import com.nonlinearlabs.NonMaps.client.dataModel.editBuffer.PhysicalControlParameter;
 import com.nonlinearlabs.NonMaps.client.world.overlay.belt.parameters.BeltParameterLayout;
 import com.nonlinearlabs.NonMaps.client.world.overlay.belt.parameters.BeltParameterLayout.Mode;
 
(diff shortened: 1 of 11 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/world/overlay/belt/parameters/recall/ParameterRecallArea.java
```diff
@@ -17,9 +18,12 @@ public ParameterRecallArea(BeltParameterLayout parent) {
 	public boolean isChanged() {
 		EditBufferModel eb = EditBufferModel.get();
 		BasicParameterModel bpm = eb.getSelectedParameter();
-		if(bpm != null)
-			return bpm.value.value.getValue() != bpm.originalValue.getValue();
-		return false;
+		if(bpm == null)
+			return false;
+		if(bpm instanceof PhysicalControlParameter)
```
hhoegelo: I still don't get why we can have a virtual function in in BasicParameterModel, overridden ind PhysicalControlParameter. All it needs to get this clean, is a handy name, no?

playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
```diff
@@ -114,10 +115,16 @@ bool ParameterLayout2::onRotary(int inc, ButtonModifiers modifiers)
 
 void ParameterLayout2::handlePresetValueRecall()
 {
-  if(getCurrentEditParameter()->Parameter::isChangedFromLoaded())
-  {
+  auto param = getCurrentEditParameter();
+  auto recall{false};
+
+  if(auto modP = dynamic_cast<ModulateableParameter*>(param))
```
hhoegelo: better have a method: isValueChangedFromLoaded that checks the value only. Would improve readability and produce less awkward code as you can get rid of the cast and the if etc.

playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/ParameterNameLabel.cpp
```diff
@@ -75,7 +75,7 @@ const Glib::ustring ParameterNameLabel::truncateMCName(const bool changed, const
 
 void ParameterNameLabel::handleParameterName(const Parameter *pParameter)
 {
-  const auto changed = pParameter->Parameter::isChangedFromLoaded();
```
hhoegelo: same here: isValueChangedFromLoaded would help

playground/src/parameters/Parameter.cpp
```diff
@@ -229,20 +229,25 @@ PresetParameter *Parameter::getOriginalParameter() const
...
-    return std::fabs(originalParameter->getValue() - rawNow) > epsilon;
-  }
-  else
-  {
-    return std::fabs(getDefaultValue() - rawNow) > epsilon;
-  }
+bool Parameter::isValueChangedFromLoaded() const {
+    const auto rawNow = getControlPositionValue();
+    const auto epsilon = 0.5 / getValue().getFineDenominator();
+    DebugLevel::gassy("Using", epsilon, "as epsilon for Parameter::isChangedFromLoaded!", getLongName());
```
hhoegelo: do we still need the gassy output here?

PR #839, opened 2019-03-27
print min, average and max performance, change algorithm to calc perf…

hhoegelo:
…ormance

### DIFF
--- a/audio-engine/src/io/Log.h
+++ b/audio-engine/src/io/Log.h
@@ -55,7 +55,8 @@ class Log
     std::cout << std::flush;
   }
 
-  template <LogMode mode, typename... Args> static void output(const Args&... args)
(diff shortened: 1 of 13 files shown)


hhoegelo:
The variance of the time needed for process the buffer is huge.I got outputs like these:

[14.5% ... 44.1% ... 117.0%]

I double checked it is actually the measured performance. I doubt values like this will be very meaningful.

MatthiasSeeber:
I have similar outputs like [34.8% ... 55.7% ... 96.7%].
When using Unison, the peak can even reach > 130%. I even got a 171.8% peak after loading the initial preset and just waiting.
These values may be meaningful when testing engine behavior with renderers that always perform (as opposed to the current 'on demand' mechanism). But I'm not sure about that.

PR #842, opened 2019-03-28
cmake creates a copy of the built executable with name

hhoegelo:
"audio-engine-branchname" now

### DIFF
--- a/audio-engine/CMakeLists.txt
+++ b/audio-engine/CMakeLists.txt
@@ -9,20 +9,17 @@ set(LIBS
   alsa
   )
 
-IF(CLANG)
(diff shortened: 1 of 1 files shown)


AntonSchmied:
Great, it works!! However the last built binary will remain in the Debug/Release Folder. 

So, if I am working on **branch_1** and build then we will have _audio-engine-branch_1_. Then when I switch to **branch_2** and build, it will respectively create an _audio-engine-branch_2_. _audio-engine-branch_1_ will still be in the same folder.

Potential binary grave? will this bother anyone? @MatthiasSeeber, what do you say?

PR #847, opened 2019-04-01
845

schlenkibus:
round all values in isChanged in boled and webui
#845

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
@@ -27,7 +27,7 @@ public class BasicParameterModel extends Notifier<BasicParameterModel> {
 	}
 
 	public boolean isValueChanged() {
-		int denominator = Math.max(value.metaData.fineDenominator.getValue(), 1000);
+		int denominator = value.metaData.fineDenominator.getValue();
 		
 		int roundedVal = (int) (value.value.getValue() * denominator);
 		int roundedOgVal = (int) (originalValue.getValue() * denominator);
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
@@ -27,8 +27,10 @@ public class BasicParameterModel extends Notifier<BasicParameterModel> {
 	}
 
 	public boolean isValueChanged() {
-		double roundedVal = (int) (value.value.getValue() * 1000);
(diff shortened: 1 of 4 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
```diff
@@ -27,8 +27,10 @@ public BasicParameterModel getValue() {
 	}
 
 	public boolean isValueChanged() {
-		double roundedVal = (int) (value.value.getValue() * 1000);
-		double roundedOgVal = (int) (originalValue.getValue() * 1000);
+		int denominator = Math.max(value.metaData.fineDenominator.getValue(), 1000);
```
hhoegelo: We have the max() here? Normally, it should be unnecessary. In case of "Release Time", "Pitch", "Pitch Key Track" it will be even wrong, won't it?

playground/src/parameters/ModulateableParameter.cpp
```diff
@@ -462,9 +462,12 @@ bool ModulateableParameter::isModAmountChanged() const
...
 
-  if(auto original = getOriginalParameter())
-  {
-    return original->getModulationAmount() != getModulationAmount();
+  if(auto original = getOriginalParameter()) {
+      const int denominator = static_cast<const int>(getValue().getFineDenominator());
+      const int roundedNow = static_cast<const int>(getModulationAmount() * denominator);
+      const int roundedOG = static_cast<const int>(original->getModulationAmount() * denominator);
+      const auto epsilon = 0.5 / denominator;
+      return std::fabs(roundedOG - roundedNow) > epsilon;
```
hhoegelo: This is somewhat weird mix of integer, float and double logic. I think that comparing roundedNow and roundedOG should be enough - just like you do in WebUI.

Proof:
roundedNow and roundedOg are in the range of [-1...1] multiplied by 'den', so [-den ... den].
(roundedOG - roundedNow) is therefore an int in the range [-2 * den ... 2 * den].
epsilon is tiny -> it is meant to be used in the control position range [-1....1]
Comparing a value in the range  [-2 * den ... 2 * den] with an epsilon for the range [-1....1] is meaningless.
It will return true for all (roundedOG - roundedNow) != 0;

playground/src/parameters/Parameter.cpp
```diff
@@ -233,16 +233,19 @@ bool Parameter::isChangedFromLoaded() const
 }
 
 bool Parameter::isValueChangedFromLoaded() const {
-    const auto rawNow = getControlPositionValue();
-    const auto epsilon = 0.5 / getValue().getFineDenominator();
+    const int denominator = static_cast<const int>(getValue().getFineDenominator());
```
hhoegelo: Here, we have still the over-complicated way of doing a check for two ints being equal.

PR #849, opened 2019-04-01
#844

schlenkibus:
requesting input from @hhoegelo, bug seems clear for me, but not sure how to advance from that
#844 
modulate able parameters who are uni polar, and not of type ModulateableParameterWithUnusualModUnit
have this bug, if mod-amount is negative the parameter value scale converter will not be able to stringize the mod-amount
Is the correct solution to instantiate all offending Parameters with their fitting ModAmountConverter?

### DIFF
--- a/playground/src/parameters/ModulateableParameter.cpp
+++ b/playground/src/parameters/ModulateableParameter.cpp
@@ -306,14 +306,8 @@ Glib::ustring ModulateableParameter::stringizeModulationAmount() const
 
 Glib::ustring ModulateableParameter::stringizeModulationAmount(tControlPositionValue amount) const
 {
-  auto converter = getValue().getScaleConverter();
-  if(!converter->isBiPolar())
-  {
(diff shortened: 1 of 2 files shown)


hhoegelo:
Please have a look at master. There, it is implemented even simpler:

Glib::ustring ModulateableParameter::stringizeModulationAmount() const
{
  auto amount = getModulationAmount();
  LinearBipolar100PercentScaleConverter converter;
  return converter.getDimension().stringize(converter.controlPositionToDisplay(amount));
}

I don't know, if someone fixed the issue for master or introduced it for this branch, though.
Please investigate and - if it was introduced here - try to understand why the change has been made and what will be the implication if we revert it.

schlenkibus:
I introduced this bug, I don't recall if it was specified verbaly or displaying the mod amounts in their unit was introduced by me when I worked on #706. The changes from stringize never made it to writeDocument so the bug / unspecified Feature Neve made it to the WebUI.
@nonlinear-labs Feature or Bug? Should Modulation Amounts be displayed in their parameters unit?

PR #871, opened 2019-04-11
Playground: Fix performance issues

schlenkibus:


playground/src/proxies/lpc/LPCProxy.h
```diff
@@ -103,6 +104,7 @@ class LPCProxy
 
   int m_lpcSoftwareVersion = 0;
   Signal<void, int> m_signalLPCSoftwareVersionChanged;
+  std::shared_ptr<DebugLevel> m_debugSetting;
```
hhoegelo: I don't think this is useful. In fact, I would like to make the settings uniquely owned by the Settings-object, so spreading its ownership all over the app is the wrong direction.

playground/src/serialization/ParameterGroupsSerializer.cpp
```diff
@@ -29,6 +30,11 @@ void ParameterGroupsSerializer::readTagContent(Reader &reader) const
 {
   reader.onTag(ParameterGroupSerializer::getTagName(), [&](const Attributes &attr) mutable {
     auto group = m_paramGroups->getParameterGroupByID(attr.get("id"));
-    return new ParameterGroupSerializer(group);
+    if(group) {
+        return new ParameterGroupSerializer(group);
+    } else {
+        DebugLevel::error("could not parse parameter group");
```
hhoegelo: Did this happen to you? If so, we should probably know the reason?

playground/src/presets/EditBuffer.h
```diff
@@ -105,7 +105,9 @@ class EditBuffer : public ParameterGroupSet
 
   Uuid m_lastLoadedPreset;
   tUpdateID m_updateIdWhenLastLoadedPresetChanged = 0;
-  Glib::ustring m_name;
+
+private:
```
hhoegelo: the private tag in the private section is redundant

playground/src/parameters/Parameter.h
```diff
@@ -139,6 +139,7 @@ class Parameter : public UpdateDocumentContributor,
   Signal<void, const Parameter *> m_signalParamChanged;
 
   uint16_t m_id;
+  mutable PresetParameter* m_cachedOGParam;
```
hhoegelo: I am not happy with the m_cachedOGParam. Can we be sure, that the object lives as long as the edit buffer - or is the pointer maintained by the parameter pointed to?

playground/src/serialization/ParameterGroupsSerializer.cpp
schlenkibus: It was premature, I still had my Layer-Foo editbuffer in my preset-manager. I will remove that.

playground/src/parameters/Parameter.h
schlenkibus: I will review that, also Stephan found a bug which I can pinpoint to not being cautious enough with this pointer. I will see what implementation will be less error prone and easier to maintain

PR #873, opened 2019-04-11
Defines to enums

hhoegelo:
I turned a lot of defines into enums, so that in later refactorings we will get better support by the compiler, as we won't be able to mismatch signals, parameters and the various param types.


hhoegelo:
As far as I know, 'int' is the type that should be used for array index. But there is a huge discussion about this in internet - I don't think it makes any difference.
Having enums with multiple entries having the same number is no problem and often used.

PR #880, opened 2019-04-17
Code quality only. This patch should not change programs behavior in any way.

hhoegelo:
- Make parameter info type safe
- Rename PARAM_XYZ_TYPES::XYZ_FOO into Xyz::Foo to have less clutter and improve readability


hhoegelo:
The first part of the patch (introduction of types in parameter info list) is important to me.
The second part - renaming of PARAM_XYZ_TYPES - is not that important. If you insist in having the long, capital letter names, I can easily revert it.

MatthiasSeeber:
made some corrections - rest seems good to me

PR #898, opened 2019-04-24
891

schlenkibus:
#891

### DIFF
--- a/playground/src/presets/Preset.cpp
+++ b/playground/src/presets/Preset.cpp
@@ -277,12 +277,10 @@ void Preset::writeDiff(Writer &writer, const Preset *other) const
 
   auto enabled = [&](const Preset *p) {
     auto eb = pm->getEditBuffer();
-    auto ebUUID = eb->getUUIDOfLastLoadedPreset();
-    auto isLoaded = p->getUuid() == ebUUID;
-    auto isEditBuffer = posString(p) == "Edit Buffer";
-    auto loadEnabledEditbuffer = isLoaded && eb->anyParameterChanged();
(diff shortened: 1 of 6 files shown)


playground/src/presets/PresetParameter.cpp
```diff
@@ -35,6 +35,8 @@ tControlPositionValue PresetParameter::getValue() const
 
 void PresetParameter::setValue(UNDO::Transaction *transaction, tControlPositionValue v)
 {
+  v = std::min<tControlPositionValue>(1, std::max<tControlPositionValue>(v, -1));
```
hhoegelo: consider ScaleConverter::getControlPositionRangeBipolar().clip(v)

playground/src/presets/Preset.cpp
```diff
@@ -253,30 +258,56 @@ void Preset::writeDocument(Writer &writer, UpdateDocumentContributor::tUpdateID
...
 
-    for(auto &group : m_parameterGroups)
-      group.second->writeDiff(writer, group.first, other->findParameterGroup(group.first));
+    writeGroups(writer, other);
   });
 }
+
+void Preset::writeGroups(Writer &writer, const Preset *other) const
+{
+  for(auto id : { "Env A", "Env B", "Env C",    "Osc A", "Sh A",   "Osc B",  "Sh B",   "FB",  "Comb", "SVF",  "Mixer",
```
hhoegelo: As the order of groups is important for displaying in the web ui compare dialog only, I would prefer to reorder the items there. Because, other developers would probably look for the order definition there. Also, writing the doc happens all the time, decreasing performance, while displaying happens only once in a while....

PR #900, opened 2019-04-24
replace occurences of ParameterEditor.isChanged to use EditBufferMode…

schlenkibus:
…l, fixes load button that was not updated after store

#883

### DIFF
--- a/NonMaps/src/com/nonlinearlabs/NonMaps/client/presenters/PresetManagerPresenter.java
+++ b/NonMaps/src/com/nonlinearlabs/NonMaps/client/presenters/PresetManagerPresenter.java
@@ -50,7 +50,7 @@ public class PresetManagerPresenter {
 	}
 	
 	public String getLoadedPresetNumberString() {
-		boolean mod = NonMaps.theMaps.getNonLinearWorld().getParameterEditor().isModified();
+		boolean mod = EditBufferModel.get().isAnyParamChanged();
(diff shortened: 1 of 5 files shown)


NonMaps/src/com/nonlinearlabs/NonMaps/client/presenters/PresetManagerPresenter.java
```diff
@@ -50,7 +50,7 @@ public String getBaseLoadedPresetNumberString() {
 	}
 	
 	public String getLoadedPresetNumberString() {
-		boolean mod = NonMaps.theMaps.getNonLinearWorld().getParameterEditor().isModified();
+		boolean mod = EditBufferModel.get().isAnyParamChanged();
```
hhoegelo: Does isAnyParamChanged() respect color & info attributes, modulation sources, pedal & ribbon properties etc? Did getParameterEditor().isModified() do it? If respected scopes of the two methods differ, which one do we want to have?

schlenkibus: The EditBufferModel isAnyParamChanged does less than getParameterEditor.isModified() did. But I believe that the old way, would not have been correct as it takes the attributes of EB into account for  m_isModified.

modulation sources (and ammounts) do count into change
pedalmodes and ribbon properties are of no interest for recall -> unlike modulation

PR #901, opened 2019-04-24
add missing onRecallValue signal in EditBuffer that is used by Parame…

schlenkibus:
…terNameLabel and MiniParameter to update onStore/undo

#899

### DIFF
--- a/playground/src/presets/EditBuffer.cpp
+++ b/playground/src/presets/EditBuffer.cpp
@@ -47,10 +47,6 @@ void EditBuffer::initRecallValues(UNDO::Transaction *transaction, const Preset *
     m_recallSet.copyParamSet(transaction, p);
   else
     m_recallSet.onPresetDeleted(transaction);
-
-  transaction->addSimpleCommand([this](auto) {
-      m_signalRecallValues.send();
-  });
 }
 
 const Glib::ustring &EditBuffer::getRecallOrigin() const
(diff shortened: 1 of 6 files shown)


playground/src/presets/EditBuffer.cpp
```diff
@@ -47,6 +47,10 @@ void EditBuffer::initRecallValues(UNDO::Transaction *transaction, const Preset *
     m_recallSet.copyParamSet(transaction, p);
   else
     m_recallSet.onPresetDeleted(transaction);
+
+  transaction->addSimpleCommand([this](auto) {
```
hhoegelo: on undo, this command will be executed first - receivers that react synchronously will see the wrong values. I think, the edit buffer should bot send signals on behalf of its children. Better implement the signal in RecallSet and connect clients to it.

playground/src/presets/EditBuffer.h
```diff
@@ -102,6 +103,7 @@ class EditBuffer : public ParameterGroupSet
   Signal<void> m_signalChange;
   Signal<void> m_signalPresetLoaded;
   Signal<void> m_signalLocksChanged;
+  Signal<void> m_signalRecallValues;
```
hhoegelo: should be moved into RecallSet

playground/src/presets/EditBuffer.cpp
```diff
@@ -130,6 +134,10 @@ connection EditBuffer::onLocksChanged(slot<void> s)
   return m_signalLocksChanged.connectAndInit(s);
 }
 
+connection EditBuffer::onRecallValuesChanged(slot<void> s) {
+    return m_signalRecallValues.connect(s);
```
hhoegelo: either move or at least redirect to a Signal living in the RecallSet

PR #907, opened 2019-04-25
890

schlenkibus:
#890


NonMaps/src/com/nonlinearlabs/NonMaps/client/dataModel/editBuffer/BasicParameterModel.java
```diff
@@ -27,10 +27,9 @@ public BasicParameterModel getValue() {
 	}
 
 	public boolean isValueChanged() {
-		int denominator = value.metaData.fineDenominator.getValue();
-		
-		int roundedVal = (int) (value.value.getValue() * denominator);
-		int roundedOgVal = (int) (originalValue.getValue() * denominator);
+		int compareDenominator = value.metaData.fineDenominator.getValue() * 2;
```
hhoegelo: I do not understand why this is necessary.

playground/src/groups/EnvelopeAGroup.cpp
```diff
@@ -44,7 +44,7 @@ void EnvelopeAGroup::init()
 
   appendParameter(
       new ModulateableParameterWithUnusualModUnit(this, 12, ScaleConverter::get<LinearBipolar24DbScaleConverter>(),
-                                                  ScaleConverter::get<LinearBipolar48DbScaleConverter>(), 0, 48, 480));
+                                                  ScaleConverter::get<LinearBipolar48DbScaleConverter>(), 0, 48, 240));
```
hhoegelo: fine step size documented in ParameterList 1.57 is 480. So, i doubt this change is correct.

PR #910, opened 2019-04-29
change parameter copyTo to presetParam to take quantized clipped valu…

schlenkibus:
…e instead of raw value to prevent saving invalid values into recall.

#909

### DIFF
--- a/playground/src/parameters/Parameter.cpp
+++ b/playground/src/parameters/Parameter.cpp
@@ -181,7 +181,7 @@ void Parameter::copyFrom(UNDO::Transaction *transaction, const PresetParameter *
 
 void Parameter::copyTo(UNDO::Transaction *transaction, PresetParameter *other) const
 {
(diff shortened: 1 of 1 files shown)


playground/src/parameters/Parameter.cpp
```diff
@@ -181,7 +181,7 @@ void Parameter::copyFrom(UNDO::Transaction *transaction, const PresetParameter *
 
 void Parameter::copyTo(UNDO::Transaction *transaction, PresetParameter *other) const
 {
-  other->setValue(transaction, getValue().getRawValue());
```
hhoegelo: I just changed it the other way around in 

https://github.com/nonlinear-labs-dev/C15/commit/b86813bdecf566f08d8a09aab126fafc7d289906

because the "original" value and the current value differed in case the quantised value is stored - leading to huge amount of yellow rects in the webUI. So, this change will re-introduce issues with incorrectly shown "changed values".

PR #929, opened 2019-05-07
change on bank delete -> cluster rearrangement

schlenkibus:
#925


playground/src/presets/PresetManager.cpp
```diff
@@ -450,10 +450,61 @@ void PresetManager::moveBank(UNDO::Transaction *transaction, const Bank *bankToM
...
+    if(bottom)
+    {
+      bottom->setAttachedToBank(transaction, master->getUuid());
+      bottom->setAttachedDirection(transaction, std::to_string(static_cast<int>(dir)));
+
+      if(right)
+      {
+        right->setAttachedToBank(transaction, bottom->getUuid());
+        right->setAttachedDirection(transaction, std::to_string(static_cast<int>(Bank::AttachmentDirection::left)));
+        return true;
```
schlenkibus: move return to upper scope

PR #931, opened 2019-05-08
move call to EditBufferSnapshotMaker::addSnapshotIfRequired to only EditBuffer::copyFrom

schlenkibus:
#924

### DIFF
--- a/playground/src/http/UndoScope.cpp
+++ b/playground/src/http/UndoScope.cpp
@@ -1,11 +1,6 @@
 #include "UndoScope.h"
 #include "xml/Writer.h"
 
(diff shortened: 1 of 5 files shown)


playground/src/http/UndoScope.cpp
```diff
@@ -1,10 +1,8 @@
 #include "UndoScope.h"
 #include "xml/Writer.h"
-#include <presets/EditBufferSnapshotMaker.h>
 
 void UndoScope::onAddTransaction(UNDO::Transaction* transaction)
```
hhoegelo: the overridden method can be removed now

PR #953, opened 2019-06-07
[dsp_optimization] fixed glitchy pitch behavior for key down and note shift

MatthiasSeeber:
- glitchy pitch issues seem fixed now
- simpler event logic for key events , avoided (most) redundant calculations (less cpu spike intensity is expected)
- fix is implemented via #define and can be disabled
- a similar fix will follow soon for the master branch (but because of large structural differences between master and dsp_optimization, cherry picking will not be helpful - so the fix will be made from scratch)

### DIFF
--- a/audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
+++ b/audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
@@ -823,6 +823,7 @@ void dsp_host::keyApply(uint32_t _voiceId)
 
 void dsp_host::keyUp156(const float _velocity)
 {
+#if test_preload_update != 2
   uint32_t voiceId = m_decoder.m_voiceFrom;
   const uint32_t unisonVoices = 1 + static_cast<uint32_t>(m_params.getParameterValue(Parameters::P_UN_V));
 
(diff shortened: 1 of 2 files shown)


audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
```diff
@@ -833,10 +834,44 @@ void dsp_host::keyUp156(const float _velocity)
...
+    m_params.getBody(index).m_preload = 0;
+    m_params.m_preload = 0;
+    m_decoder.m_listId = 0;
+    // unison loop
+    for(uint32_t v = 0; v < unisonVoices; v++)
+    {
+        //m_params.m_event.m_poly[voiceId].m_velocity = velocity;
+        //m_params.m_event.m_poly[voiceId].m_type = 0;
+        float notePitch = m_params.getBody(index).m_value
+            + (m_params.getParameterValue(Parameters::P_UN_DET) * m_params.m_unison_detune[uVoice][v])
```
hhoegelo: consider calling m_params.getParameterValue(Parameters::P_UN_DET) and m_params.getParameterValue(Parameters::P_MA_T) once before entering the loop.

audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
```diff
@@ -833,10 +834,44 @@ void dsp_host::keyUp156(const float _velocity)
...
+            + m_params.getParameterValue(Parameters::P_MA_T) + m_params.m_note_shift[voiceId];
+        m_params.newEnvUpdateStop(voiceId, notePitch, velocity);
+        // increment voice id
+        voiceId++;
+        index++;
```
hhoegelo: Consider const 'voiceId' and const 'index', and add 'v' at the according places instead of modifying the variables. This might unveil optimization potential to the compiler.

audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
```diff
@@ -833,10 +834,44 @@ void dsp_host::keyUp156(const float _velocity)
   }
 
   preloadUpdate(2, 0);
+#else
+    // preparation
+    uint32_t voiceId = m_decoder.m_voiceFrom;
+    const uint32_t uVoice = static_cast<uint32_t>(m_params.getParameterValue(Parameters::P_UN_V));
+    const uint32_t unisonVoices = 1 + uVoice;
+    uint32_t index = m_params.getHead(Parameters::P_KEY_BP).m_index + voiceId;
+    const float velocity = m_params.scale(
```
hhoegelo: The loop below seems to expect, that 

`(voiceId + unisonVoices) < dsp_number_of_voices`

Please consider to make this expectation clear to the reader by adding an assertion or at least a comment.

audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
```diff
@@ -855,6 +890,83 @@ void dsp_host::keyDown156(const float _velocity)
   }
 
   preloadUpdate(2, 0);
+#else
```
hhoegelo: See comments for keyUp156() regarding 
* less state maintaining (voiceId++, index++) 
* less redundant method calls (getParameterValue) 
* making important relationships between state variables clear to the reader

audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
MatthiasSeeber: - 'index' is needed once within the loops for keyUp and keyDown
- 'voiceId' however is needed twice for keyUp and six times for keyDown, so is it really better practice to perform the same addition six times? or shall I provide a variable within the loop that does this once?

hhoegelo: I am not 100% sure, but adding two integers is done in Nullkommanix, while incrementing, storing, loading needs Nullkommanix + 1. On the other hand, the compiler will probably detect this optimisation potential anyway choose the best. So, if you consider my thoughts it but decide against - no problem.

audio-engine/src/synth/c15-audio-engine/dsp_host.cpp
```diff
@@ -855,6 +895,89 @@ void dsp_host::keyDown156(const float _velocity)
...
+    // unison loop
+    /** !! !! !!
+     * NOTE that at Milestone 1.56, the LPC still performs the Voice Allocation,
+     * and Unison Clusters are GUARANTEED to always be within the allowed number of Voices,
+     * never rising above or wrapping around. Example: Unison Voices = 3 --> possible Unison Clusters:
+     * [0,1,2], [3,4,5], [6,7,8], [9,10,11], [12,13,14], [15,16,17] < 20
+     * !! !! !! **/
+    for(uint32_t v = 0; v < unisonVoices; v++)
+    {
+        m_params.m_note_shift[voiceId + v] = noteShift;
```
hhoegelo: now, that I see the code I have to admit that 

`
const auto paramIdx = voiceId + v;
m_params.m_note_shift[paramIdx] = noteShift;
...` 

looks better, is obviously less code, less error-prone and less redundant.

MatthiasSeeber: agreed. btw, in [master], there are ten additions, so two more reasons..

PR #969, opened 2019-07-03
Ae fixes

MatthiasSeeber:
- provided disable option for level kt clipping (as there is none in R5 - envelopes and fb mixer are affected)
- fixed Oscillator tuning for negative pitches
- provided new State Variable Filter (which should be similar to the R5 version)


audio-engine/src/synth/c15-audio-engine/dsp_defines_signallabels.h
```diff
@@ -95,112 +95,114 @@ enum class Signals
   = 53,  // SVF Upper Filter FM Amount (Cutoff, Cutoff KT, Cutoff Env C Amount, Env C, Spread, FM, Master Tune, Reference, Key Pitch)
   SVF_F2_FM
   = 54,  // SVF Lower Filter FM Amount (Cutoff, Cutoff KT, Cutoff Env C Amount, Env C, Spread, FM, Master Tune, Reference, Key Pitch)
-  SVF_RES = 55,    // SVF Resonance (Resonance, Res KT, Res Env C Amount, Env C, Master Tune, Key Pitch)
```
hhoegelo: What is the reason for having explicit numbers in this enum? Are the signals referred by others using their number? If not, we should just get rid of the assignments and let the compiler assign the numbers automatically.

audio-engine/src/synth/c15-audio-engine/ae_svfilter_ni.h
```diff
@@ -0,0 +1,37 @@
...
+    void init(float _samplerate);
+    void apply(const FloatVector &_sampleA, const FloatVector &_sampleB,
+               const FloatVector &_sampleComb, SignalStorage &signals);
+    void resetDSP();
+
+    //**************************** State Variables ****************************//
+    FloatVector m_first_int1_stateVar, m_first_int2_stateVar;
+    FloatVector m_second_int1_stateVar, m_second_int2_stateVar;
+
+    FloatVector m_first_sv_sample; // seems obsolete?
```
AntonSchmied: this does not seem to be needed anymore! can be kicked out from .h and .cpp

PR #990, opened 2019-07-21
Dynamic Layouts

schlenkibus:
Dynamic-Layout Library provides following features:

- Primitives with Properties and Styling capabilities
- Property value providing Events
- JSON description of layouts, controls, events
- instantiating, styling, grouping of Primitives into Control-Classes
- instantiating, positioning and mapping events in layouts
- button mapping to event sinks
- layout/ui flow described using UIState and layout-selectors
- documentation: full control-library reference, tutorials, [see wiki](https://github.com/nonlinear-labs-dev/C15/wiki)

Playground extensions:
- Layout Mode: _Old_, _Mixed_, _New_ (Mixed =  favors new, falls back to old on exception, New = developer option -> shows debug and errors)
- Force old Setup to be shown on 3 consecutive tries inside _New_ Layout mode (setup -> error-layout -> enter -> setup) repeat 3 times and the old setup layout will be shown, this is necessary, if no setup layout is defined in the new mode, you cant enter the setup to change the layout mode, could be removed on completion)
- new (Dynamic) Parameter Layouts as reference


schlenkibus:
TODO: Update deploy to take nltools into account, I propose keeping the nltools lib inside the playground directory and starting the playground / bbbb with:
LD_LIBRARY_PATH=/nonlinear/playground ./playground --bbbb-host=localhost --layouts=/nonlinear/playground/resources/Templates

this way we do not have to care for older version of the lib, as they are not being found

PR #1104, opened 2019-11-18
Include Heartbeat MIDI-HDW per Default

AntonSchmied:
Is this also needed in the Master Branch?


hhoegelo:
> Is this also needed in the Master Branch?

I don't think so - there will be a more direct communication via nltools-messages.

PR #1141, opened 2019-11-25
avoid z-delay for fx signals by #define

MatthiasSeeber:
In pe_definines_config.h, a new flag determines the z-delay behavior of the FX signals for the Feedback Mixer:
- disabled (0): LinuxEngine behavior up to now - one-sample z-delay on FX dry/wet
- enabled (1): R5 Prototype behavior and new default - no z-delay on FX dry/wet

### DIFF
--- a/audio-engine/src/synth/c15-audio-engine/pe_defines_config.h
+++ b/audio-engine/src/synth/c15-audio-engine/pe_defines_config.h
@@ -12,6 +12,9 @@
 
 #include <stdint.h>
 
+// switch signal processing modes: 0 - FX have z-delay (Linux Engine up to now), 1 - FX have no z-delay (like R5 Prototype)
+#define REWORK_POLY_DSP 1
+
 /* Param Interface Testing */
 #define PARAM_ITERATOR 1  // 0: render param bodies by clockIds; 1: render param bodies by direct iteration (default)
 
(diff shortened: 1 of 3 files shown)


audio-engine/src/synth/c15-audio-engine/pe_defines_config.h
```diff
@@ -12,6 +12,9 @@
 
 #include <stdint.h>
 
+// switch signal processing modes: 0 - FX have z-delay (Linux Engine up to now), 1 - FX have no z-delay (like R5 Prototype)
+#define REWORK_POLY_DSP 1
```
hhoegelo: I'd recommend using 

inline constexpr bool REWORK_POLY_DSP = true;

instead of 

#define REWORK_POLY_DSP 1

It has the same cost of evaluation at runtime (=> zero), but it forces the compiler to look at the 'inactive' lines of code, too. This will protect you from breaking the inactive code by changing / moving / renaming members, stack variables, function signatures etc.

MatthiasSeeber: As it is expected that there won't be many changes/improvements of the 1.5 Audio Engine, this strategy is rather planned to be used in further development of the 1.7 milestone. See [https://github.com/nonlinear-labs-dev/C15/issues/1170](url)

PR #1168, opened 2019-12-09
fixed EditSmoothingTime getDisplayString() to truncate trailing zeros

### LINKED ISSUE #1096: Edit Smoothing Time
If fine is enabled (see #1057) the setting can be changed in increments of 0.1 ms. 

- [ ] Edit Smoothing Time unnecessary precision in the low end of the range (0.100 ms is unnecessary, could be 0.1 ms)

- [ ] not enough precision in the upper end of the range if fine is enabled 200 ms will be shown for up to 10 presses on dec

schlenkibus:
fixes #1096


nltools/include/nltools/StringTools.h
```diff
@@ -14,6 +14,15 @@ namespace nltools
...
     }
   }
+
+  namespace Truncate
+  {
+    std::string trimEnd(const std::string &in);
+    std::string numberWithoutZeros(const std::string &in);
+  }
+
+  std::vector<std::string> splitStringOnAnyDelimiter(const std::string &s, char delimiter);
```
hhoegelo: Seems to be copied from the according playground/tools/StringTools.h/cpp. If so, please remove the original implementation.

nltools/src/StringTools.cpp
```diff
@@ -17,4 +17,50 @@ namespace nltools
   {
     return g_str_has_prefix(string.c_str(), test.c_str());
   }
+
+  std::vector<std::string> splitStringOnAnyDelimiter(const std::string &s, char delimiter)
```
hhoegelo: I suggest renaming. 

std::vector<std::string> split(const std::string &s, char delimiter) would be enough for me.

nltools/src/StringTools.cpp
```diff
@@ -17,4 +17,50 @@ namespace nltools
...
+    while(std::getline(tokenStream, token, delimiter))
+    {
+      strings.emplace_back(token);
+    }
+    return strings;
+  }
+
+  namespace Truncate
+  {
+    std::string trimEnd(const std::string &in)
```
hhoegelo: Can't understand this one: As I read the algorithm, it should produce "DCBA" from "0A0BC00D0" - which is not what the name promises. Also, trimming - as fas as I understand - is related to white spaces, not to the 0-character.

nltools/src/StringTools.cpp
```diff
@@ -17,4 +17,50 @@ namespace nltools
...
+    std::string token;
+    std::istringstream tokenStream(s);
+    while(std::getline(tokenStream, token, delimiter))
+    {
+      strings.emplace_back(token);
+    }
+    return strings;
+  }
+
+  namespace Truncate
```
hhoegelo: I don't think we need extraneous namespaces for file-local, static helper functions.

schlenkibus:
This PR can be closed and #1096 revisited until the desired behavior for #1057 is defined.

PR #1176, opened 2019-12-10
Fix Heatbeat implementation 

schlenkibus:
Heartbeat was successfully tested in with LPC software version 103


playground/src/proxies/lpc/MessageParser.cpp
```diff
@@ -102,14 +102,14 @@ size_t MessageParser::getNumInitialBytesNeeded()
 
 const MessageParser::NLMessage &MessageParser::getMessage() const
```
hhoegelo: it should be 

MessageParser::NLMessage MessageParser::getMessage() const

schlenkibus: fixing that

PR #1183, opened 2019-12-11
update playground deploy and service

schlenkibus:
updated service to include audio-engine host and CMakeLists to install libnltools next to playground exe


schlenkibus:
@hhoegelo I agree that the current solution is suboptimal and not going to get better over time. If you ping me when you are ready we could discuss and test deploy your approach.

PR #1185, opened 2019-12-11
Move functionality from LPCProxy to AudioEngineProxy

### LINKED ISSUE #1171: 1.7 Send Settings on Audio-Engine wake up
Send all settings required for the audio engine when audio engine starts.

schlenkibus:
fixes #1171 
and fixes hwsources from lpc

remove unused functionality from lpcproxy (could be even more? is sendParameter still needed?)


playground/src/presets/EditBufferActions.cpp
```diff
@@ -19,8 +19,6 @@ EditBufferActions::EditBufferActions(EditBuffer* editBuffer)
     : super("/presets/param-editor/")
 {
   addAction("sync-lpc", [=](std::shared_ptr<NetworkRequest> request) mutable {
```
hhoegelo: Please rename to "sync-audioengine" and do according changes in WebUI, too.

playground/src/proxies/lpc/LPCProxy.cpp
```diff
@@ -138,15 +167,10 @@ void LPCProxy::onParamMessageReceived(const MessageParser::NLMessage &msg)
   gint16 value = separateSignedBitToComplementary(msg.params[1]);
   auto vg = Application::get().getHWUI()->getCurrentVoiceGroup();
 
-  auto param = Application::get().getPresetManager()->getEditBuffer()->findParameterByID({ id, vg });
-
-  if(!param)
-    param = Application::get().getPresetManager()->getEditBuffer()->findParameterByID({ id, VoiceGroup::Global });
-
-  if(auto p = dynamic_cast<PhysicalControlParameter *>(param))
+  if(auto *param = dynamic_cast<PhysicalControlParameter *>(findPhysicalControlParameterFromID(id)))
```
hhoegelo: is id really 0...7 nowadays? As far as I remember we used to use the parameter id?

playground/src/proxies/lpc/ParameterMessageComposer.cpp
```diff
@@ -3,6 +3,8 @@
 #include "MessageParser.h"
 #include "device-settings/DebugLevel.h"
 
+#warning "kann das weg?"
```
hhoegelo: Yes, I think so.

playground/src/proxies/lpc/LPCProxy.cpp
schlenkibus: It's only 0..7 now no other parameters from LPC now, renamed the functions accordingly

PR #1202, opened 2019-12-13
1198

### LINKED ISSUE #1198: ParameterChangedMsg issues
- [x] Split Point: should be modulateable parameter and send modulateable parameter message (currently, it is unmodulateable and sends unmodulateable parameter message)
- [x] Mono Glide: same thing
- [x] Voice Group: Volume, Tune and To_FX should be modulateable parameters and should send modulateable parameter messages (To_FX should be implemented - possibly besides Output Mixer Level)
- [x] Feedback Mixer: new Parameters (see ParameterList_1.7B.ods) Osc, Osc_Source, Comb_Source, SFV_Source, FX_Source should also be implemented (as modulateable parameters)

- [x] Remove Locks from Parameter / Preset Messages

schlenkibus:
fixes #1198


nltools/include/nltools/messaging/Message.h
```diff
@@ -284,8 +275,8 @@ namespace nltools
       std::array<ParameterGroups::MacroParameter, 6> macros;
       std::array<ParameterGroups::UnmodulateableParameter, 6> macrotimes;
 
-      std::array<ParameterGroups::ModulateableParameter, 100> modulateables;
-      std::array<ParameterGroups::UnmodulateableParameter, 97> unmodulateables;
+      std::array<ParameterGroups::ModulateableParameter, 101> modulateables;
```
hhoegelo: don't we need all the new parameters (env a, b, c...) to be added here?

schlenkibus: no because the added parameters were mocked before and already included in the arrays

playground/src/parameters/mono-mode-parameters/MonoParameter.h
```diff
@@ -1,11 +1,44 @@
 #pragma once
 #include <parameters/Parameter.h>
+#include <parameters/ModulateableParameter.h>
+#include <proxies/hwui/panel-unit/boled/parameter-screens/ModulateableParameterLayouts.h>
+#include <proxies/hwui/panel-unit/boled/parameter-screens/MonoModeParameterLayout.h>
+#include <proxies/hwui/panel-unit/boled/parameter-screens/ParameterInfoLayout.h>
 
-class MonoParameter : public Parameter
+template <typename tBaseParameter> class MonoParameter : public tBaseParameter
```
hhoegelo: Please re-consider the introduction of this template intermediate class. I fail to see its purpose. The only functionality it offers (createLayout) is dependent of the base class - so it seems to me that two separate non-template base class (ModulateableMonoParam, UnmodulateableMonoParam) are more appropriate - but maybe I just don't get your point.

PR #1204, opened 2019-12-13
1203

### LINKED ISSUE #1203: Import Old Presets into 1.7

schlenkibus:
fixes #1203


playground/src/presets/Preset.cpp
```diff
@@ -366,3 +366,17 @@ void Preset::writeGroups(Writer &writer, const Preset *other) const
...
+
+PresetParameterGroup *Preset::createParameterGroup(GroupId id)
+{
+  try
+  {
+    auto &vgMap = m_parameterGroups[static_cast<size_t>(id.getVoiceGroup())];
+    auto ret = vgMap.insert(std::pair{ id.getName(), std::make_unique<PresetParameterGroup>() });
+    return ret.first->second.get();
+  }
+  catch(...)
```
hhoegelo: which exceptions do you expect here and how could "find" handle an exception thrown by "create". The logic looks inverted and odd to me.

PR #1215, opened 2019-12-17
HWUI misc fixes

### LINKED ISSUE #1187: Mono Glide should be modulateable-Parameter

### LINKED ISSUE #1188: Unison + Layer
- [ ] Unison Voices should appear "Global" in Layer mode
- [ ] Other Unison Parameters should be per voicegroup

### LINKED ISSUE #1189: Mono + Layer
Mono Parameters will be global (from vg I) and mono glide will be per voice group

### LINKED ISSUE #1198: ParameterChangedMsg issues

### LINKED ISSUE #1205: Active Overview broken
Way too many parameter LED are lit up, compare to previous version, before moving to LibparameterDB

### LINKED ISSUE #1213: Sound-Screen: "Convert To.."-Menü durch Druck auf "Sound"-Button wieder schliessen (und im Screen bleiben)
Bisher lässt sich das Menü nicht wieder schliessen. Mit dem Sound-Button wäre es möglich, aber man müsste dabei im Screen bleiben. Wenn das Menü nicht offen ist, verlässt man den Screen mit dem Sound-Button.

(Alternative: Menü mit 3 States: "Split", "Layer" und Menü=Aus)

schlenkibus:
fixes #1188 
fixes #1187 
fixes #1205 
fixes #1189 
fixes #1198 
fixes #1213


playground/src/groups/UnisonGroup.cpp
```diff
@@ -61,6 +62,26 @@ namespace Detail
         : tParameter(args...){};
 
     DFBLayout *createLayout(FocusAndMode focusAndMode) const override
+    {
```
hhoegelo: I don't think its worth another template class here: There is actually nothing shared between both users - only the dispatching code that tries to separate both classes again.

PR #1258, opened 2019-12-23
fixed splitpoint display value

schlenkibus:


playground/src/parameters/SplitPointParameter.cpp
```diff
@@ -32,12 +33,10 @@ DFBLayout *SplitPointParameter::createLayout(FocusAndMode focusAndMode) const
 
 std::string SplitPointParameter::getDisplayValue(VoiceGroup vg) const
 {
-  auto converter = ScaleConverter::get<KeyWithOctaveScaleConverter>();
```
hhoegelo: I would prefer the old version without 'magic' value. Is there any chance to get the old approach working?

PR #1284, opened 2019-12-30
Engine additions and fixes

MatthiasSeeber:
- enabled rigorous midi logging
- fixed cabinet ctrl signals


MatthiasSeeber:
The merge conflict seems to originate from "Merge pull request #1283", in which the ParameterList was changed: The Volume and Tune Labels from the Master Group were changed to be "Master Volume", "Master Tune" (instead of just "Volume" and "Tune").
I fail to see the reason behind this, as the labels just are "Volume" and "Tune" on C15 devices with Milestone 1.5 on int..?

PR #1297, opened 2020-01-02
added type labels for webui

### LINKED ISSUE #1288: WebUI: Die Presets sollten auch die Split/Layer-Symbole bekommen (wie im BOLED)

schlenkibus:
fixes #1288

### DIFF
--- a/NonMaps/src/main/java/com/nonlinearlabs/client/world/maps/presets/bank/preset/Name.java
+++ b/NonMaps/src/main/java/com/nonlinearlabs/client/world/maps/presets/bank/preset/Name.java
@@ -11,11 +11,6 @@ class Name extends Label {
 	}
 
 	@Override
-	public boolean setText(String text) {
-		return super.setText(text);
-	}
-
-	@Override
 	protected TextAlign getTextAlignment() {
 		return TextAlign.LEFT;
 	}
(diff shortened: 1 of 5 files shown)


NonMaps/src/main/java/com/nonlinearlabs/client/world/maps/presets/bank/preset/Name.java
```diff
@@ -10,6 +10,11 @@
 		super(parent, text);
 	}
 
+	@Override
```
hhoegelo: Please remove useless override

NonMaps/src/main/java/com/nonlinearlabs/client/world/maps/presets/bank/preset/TypeLabel.java
```diff
@@ -0,0 +1,44 @@
...
+class TypeLabel extends Label {
+
+	public TypeLabel(MapsLayout parent, String text) {
+		super(parent, text);
+	}
+
+	@Override
+	public void doFirstLayoutPass(double levelOfDetail) {
+		super.doFirstLayoutPass(levelOfDetail);
+		setNonSize(15 + getWidthMargin(), getNonPosition().getHeight());
```
hhoegelo: what is the 15? Please consider overriding getLeftMargin or getRightMargin or at least introduce a describing variable. (Of course, most of the code in NonMaps does it the same dirty way as it is done here, but if possible, we should try to get rid of those magic numbers)

PR #1300, opened 2020-01-02
add loading animation to load to part load

schlenkibus:
finishes #1232

### DIFF
--- a/playground/src/proxies/hwui/panel-unit/boled/preset-screens/SelectVoiceGroupLayout.cpp
+++ b/playground/src/proxies/hwui/panel-unit/boled/preset-screens/SelectVoiceGroupLayout.cpp
@@ -7,8 +7,9 @@
 #include <presets/PresetManager.h>
 #include <presets/EditBuffer.h>
 
-SelectVoiceGroupLayout::SelectVoiceGroupLayout()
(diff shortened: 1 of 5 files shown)


playground/src/proxies/hwui/panel-unit/boled/preset-screens/SelectVoiceGroupLayout.cpp
```diff
@@ -26,12 +27,18 @@ bool SelectVoiceGroupLayout::onButton(Buttons i, bool down, ::ButtonModifiers mo
     switch(i)
     {
       case Buttons::BUTTON_B:
-        eb->undoableLoadSelectedPresetPartIntoPart(VoiceGroup::I, currentVG);
-        hwui->setFocusAndMode({ UIFocus::Presets, UIMode::Select, UIDetail::Init });
+        m_parent->animateSelectedPreset([=]() {
```
hhoegelo: As far as I understand, the ...resetOverlay-calls in line 34 and 41 will delete 'this', while the callbacks handed over in 'animateSelectedPreset' are still referencing the implicitly captured 'this'. Please consider capturing 'eb' and 'hwui' explicitly or maybe even capture nothing and grab both objects from within static lambdas.

PR #1313, opened 2020-01-03
fixes #1292

### LINKED ISSUE #1292: Erweiterte Bedingungen für Activity LEDs durch neue Parameter
Envelope A/B - Sustain: (Sustain > 0 %) OR (Split_Levels != 0 %)
Flanger Mix: (Mix != 0 %) OR (Tremolo > 0 %)
Echo Mix: (Mix > 0 %) AND (Send > 0%)
Reverb Mix: (Mix > 0 %) AND (Send > 0%)

schlenkibus:


playground/src/proxies/hwui/panel-unit/PanelUnitPresetMode.cpp
```diff
@@ -122,6 +130,49 @@ void PanelUnitPresetMode::setStateForButton(Buttons buttonId, const std::list<in
   }
 }
 
+bool PanelUnitPresetMode::specialCaseParameterMatches(Parameter* selParam)
```
hhoegelo: the switch block here and the list in 'isSpecialCaseParameterForActivityOverview' is kind of redundant. Maybe you can manage to come up with something like 

if(auto specialCaseResult = trySpecialCaseParameter(parameter)) { states[buttonId] = (specialCaseResult == XYZ ? foo : bar; }
else if(auto mc = dynamic_cast<MacroControlParameter*>(parameter)) {...}
else if(signalFlowIndicator != invalidSignalFlowIndicator) {...}

where trySpecialCaseParameter would return something evaluating to 'false' in the default branch?

Just a thought.

schlenkibus: replaced the two functions with one which returns if the parameter, is indeed an special parameter, and a flag if the button should be turned on, by using the structured binding I find that very readable now

Issue #1295, opened 2020-01-02
use mono enable parameter from message


schlenkibus:
https://github.com/nonlinear-labs-dev/C15/pull/1291

PR #1320, opened 2020-01-05
Audio Engine NAMM Updates

MatthiasSeeber:
Dieser PR beinhaltet die in #1314 erwähnten Punkte:
- Integration der Gruppen Unison, Mono (für ParameterChangedMsg, PresetMsg) (4 + 4 Parameter) (#1295)
- 14 bit Velocity Range (@nonlinear-labs der LPC sollte hier nachziehen, falls noch nicht passiert) (#1285)
- Logging vorbereitet für verbleibende MIDI-Issues (#1279 - HW Ranges, Vel Ranges)
- Feedback Mixer Drive hat 70 dB Range (in der Audio Engine, #1319)
- Scale Parameter werden in Echtzeit angewendet
- Integration der Mono-Modes, allerdings noch buggy (gesamter Mono-Betrieb kann im Zweifelsfall abgeschaltet werden, ist momentan aber aktiviert)
- ParameterList beinhaltet alle relevanten Änderungen (#1304 + #1318 + #1319 + Audio-Engine-spezifische Mono, Unison und Scale updates)


audio-engine/src/synth/c15-audio-engine/mappable_list.h
```diff
@@ -12,6 +12,144 @@
 
 #include <stdint.h>
 
+// singular sorted List
+template <uint32_t Length> class SortedList
```
hhoegelo: It's hard for me to understand the purpose of this class, especially as it has a 'appendElement' method which I would't expect in a 'Sorted'Something. Also the number of elements stored is 2 x Length + 2, which seems wasteful to me. 

Since I do not fully understand the class, the following proposal might be wrong, but would something like this serve the purpose as well?

`template <uint32_t Length> class SortedList
{
  std::vector<uint32_t> elements;

 public:
  SortedList()
  {
    elements.reserve(Length);
  }

  inline void reset()
  {
    elements.clear();
  }

  inline void appendElement(const uint32_t _element)
  {
    // really?? This would break the sorting and thus break insert and find as well
    elements.push_back(_element);
  }

  inline void insertElement(const uint32_t _element)
  {
    elements.insert(std::upper_bound(elements.begin(), elements.end(), _element), _element);
  }

  inline void removeElement(const uint32_t _element)
  {
    elements.erase(std::find(elements.begin(), elements.end(), _element));
  }

  inline const bool isFirstElement(const uint32_t _element)
  {
    return elements.front() == _element;
  }

  inline const int32_t getFirstElement()
  {
    return elements.front();
  }

  inline const bool isLastElement(const uint32_t _element)
  {
    return elements.back() == _element;
  }

  inline const int32_t getLastElement()
  {
    return elements.back();
  }
};`

audio-engine/src/synth/c15-audio-engine/voice_allocation.h
```diff
@@ -53,14 +64,226 @@ struct VoiceAssignment
...
+        else
+        {
+          // if key is not lowest
+          m_retrigger_env = false;    // envelope does not restart
+          m_retrigger_glide = false;  // glide does not start
+          m_retrigger_phase = false;  // osc phases do not reset
+        }
+        break;
+    }
+    //return m_valid;
```
hhoegelo: bool function should return something

audio-engine/src/synth/c15-audio-engine/voice_allocation.h
```diff
@@ -53,14 +64,226 @@ struct VoiceAssignment
...
+        }
+        else
+        {
+          // if key is not highest:
+          m_key_position = m_highest.getLastElement();  // hold highest pos
+          m_retrigger_env = false;                      // envelopes keep running
+          m_retrigger_glide = false;                    // glide does not start
+        }
+        break;
+    }
```
hhoegelo: bool function should return something

audio-engine/src/synth/c15-audio-engine/mappable_list.h
MatthiasSeeber: This class serves two purposes at once:
- maintain the "latest" order of incoming (mono) key positions (hence the append method)
- maintain the "highest" order of incoming (mono) key positions (by value, hence the insert method)
- (technically, the remove method applies to both lists and is always by value, hence sorted)
The internal elements serve as pointer-like structures facilitating the methods and produce the mentioned length of the whole class.
Since similar structures are present in the engine for quite some time and produce expected results, I would suggest to postpone possible optimizations for after NAMM.

audio-engine/src/synth/c15-audio-engine/voice_allocation.h
```diff
@@ -53,14 +64,226 @@ struct VoiceAssignment
...
+  }
+  inline void reset()
+  {
+    m_latest.reset();
+    m_highest.reset();
+  }
+
+ private:
+  SortedList<Keys> m_latest, m_highest;
```
hhoegelo: From what I understand, m_highest should be a SortedList to have fast access to the highest held key, whereas m_latest should be unsorted (so no instance of SortedList), but a LRU (Least Recently Used) vector. 

LRU could be implemented like this:

template <uint32_t Length> class LRUList
{
  std::vector<uint32_t> elements;

 public:
  LRUList()
  {
    elements.reserve(Length);
  }

  inline void reset()
  {
    elements.clear();
  }

  inline void add(const uint32_t _element)
  {
    remove(_element);
    elements.push_back(_element);
  }

  inline void remove(const uint32_t _element)
  {
    elements.erase(std::find(elements.begin(), elements.end(), _element));
  }

  inline const bool isLast(const uint32_t _element)
  {
    return elements.back() == _element;
  }

  inline const int32_t getLast()
  {
    return elements.back();
  }
};

audio-engine/src/synth/c15-audio-engine/mappable_list.h
hhoegelo: ok

PR #1334, opened 2020-01-06
fixup multiple split point issues

schlenkibus:
#1372, switch voicegroup button for split point, split point short label in parameter db


playground/src/proxies/hwui/controls/SwitchVoiceGroupButton.cpp
```diff
@@ -53,24 +53,33 @@ void SwitchVoiceGroupButton::onVoiceGroupChanged(VoiceGroup newVoiceGroup)
   rebuild();
 }
 
-std::unique_ptr<UNDO::TransactionCreationScope>
-    SwitchVoiceGroupButton::createToggleVoiceGroupWithParameterHighlightScope()
+bool SwitchVoiceGroupButton::toggleVoiceGroup()
 {
   auto pm = Application::get().getPresetManager();
```
hhoegelo: the control flow is quite hard to parse here -> as far as I can see, allowToggling is false for the SplitPointParameter - only because of this the toggling is not done twice. I would prefer to sort out the split parameter first (and do an early return), only then do the if(allowToggling)... stuff. Other approaches to emphasize the exclusiveness of the two if-branches are welcome, too.

playground/src/proxies/hwui/controls/SwitchVoiceGroupButton.cpp
```diff
@@ -53,24 +53,33 @@ void SwitchVoiceGroupButton::onVoiceGroupChanged(VoiceGroup newVoiceGroup)
...
   auto eb = pm->getEditBuffer();
   auto selected = eb->getSelected();
 
-  if(allowToggling(selected, eb))
+  auto didToggleAction = false;
```
hhoegelo: this one is unused now, no?

parameter-db/generated/parameter_list.h
```diff
@@ -1901,7 +1901,7 @@ namespace C15
       356,
       0.4f,
       { Smoothers::Poly_Sync::Split_Split_Point, None, Properties::SmootherScale::Linear, 60.0f, 0.0f, false },
-      { "", "Split", "Split", "Split Point", "Split P.", "" } },
+      { "", "Split", "Split", "Split Point", "Split Point", "" } },
```
schlenkibus: @schlenkibus Matthias will take care of the parameter table -> remove changes in param_list and ods

PR #1337, opened 2020-01-07
sound edit rework

### LINKED ISSUE #1332: BOLED: Single Sound Screen: Softbutton 1: "Init" statt "Convert to.."; "Convert to Split" und "Convert to Layer" in Edit-Liste
In der Edit-Liste sollten die ersten beiden Zeilen sein:
"Convert to Split" (Selektieren + Enter)
"Convert to Layer" (Selektieren + Enter)

Nach dem Ausführen der Konvertierung die Edit-Liste schliessen.

Der "Init"-Softbutton sollte so funktionieren wie früher. (Highlight + Enter)

schlenkibus:
fixes #1332


playground/src/proxies/hwui/descriptive-layouts/concrete/sound/menus/SoundEditMenu.h
```diff
@@ -8,7 +8,38 @@ class SoundEditMenu : public ScrollMenu
...
+  void removeBlockOnTypeChanged();
+
  private:
   void onTypeChanged();
   sigc::connection m_connection;
-};
+
+  bool m_supressOnTypeChanged = false;
+
+  struct OnTypeChangedSupressor
```
hhoegelo: This type should
- be implemented in cpp file
- either be non-copyable or maintain not a bool but a counter (to avoid 'flickering' of the flag when copies are destroyed)

hhoegelo: Anyway a counter might be the better choice, because no one saves you from creating several OnTypeChangedSupressor at the same time - resulting in most probably unwanted behavior.

PR #1382, opened 2020-01-10
multiple fixes 

### LINKED ISSUE #1335: Unison Voices Scale Converter Issues
- [ ] converting a sound from single to dual leads to changed voices in part even if voices <= 12. Voices should be taken as is an only be truncated if necessary

### LINKED ISSUE #1371: "Load to Part": Globale Parameter (MCs, Master Vol/Tune etc) sollten unverändert bleiben.
Im Moment werden fälschlicherweise die globalen Parameter durch die des in den Part geladenen Presets überschrieben.

### LINKED ISSUE #1374: Mono II / Unison II Module Header
Single Sound displays Mono II / Unison II in header

### LINKED ISSUE #1376: Unison - Voices geht auch bei Single-Sounds nur bis 12, wirkt aber stärker (2*Wert - 1)
Eingestellt -> Ergebnis im Audio:
1 Voice -> 1 Voice
2 Voices -> 3 Voices
3 Voices -> 5 Voices
4 Voices -> 7 Voices
usw.

### LINKED ISSUE #1377: "Load to I" geht, "Load to II" geht nicht

schlenkibus:
fixes #1377 
fixes #1374 
fixes #1371 
fixes #1376 
fixes #1335


playground/src/groups/ParameterGroup.cpp
```diff
@@ -88,7 +88,7 @@ void ParameterGroup::copyFrom(UNDO::Transaction *transaction, const PresetParame
 {
   for(auto &myParameter : getParameters())
   {
-    if(auto otherParameter = other->findParameterByID(myParameter->getID()))
+    if(auto otherParameter = other->findParameterByID({ myParameter->getID().getNumber(), other->getVoiceGroup() }))
```
schlenkibus: Do you think findParameterByNumber makes sense here? I would prefer that over my hack-y way of obtaining the other voice group

playground/src/parameters/UnisonVoicesParameter.cpp
```diff
@@ -22,22 +22,40 @@ bool UnisonVoicesParameter::shouldWriteDocProperties(UpdateDocumentContributor::
 
 void UnisonVoicesParameter::updateScaling(SoundType type)
 {
-  auto value = getValue().getRawValue();
+  auto scope = UNDO::Scope::startTrashTransaction();
```
schlenkibus: @hhoegelo could you please point me into the correct direction how to do this properly? This is how I got the test to pass, but I am unhappy with the trashtransaction obviously

playground/src/presets/EditBuffer.cpp
```diff
@@ -724,79 +735,20 @@ void EditBuffer::undoableLoadPresetIntoDualSound(UNDO::Transaction *transaction,
 
 const SplitPointParameter *EditBuffer::getSplitPoint() const
 {
-  if(getType() == SoundType::Split)
-    return dynamic_cast<const SplitPointParameter *>(findParameterByID({ 356, VoiceGroup::Global }));
-  return nullptr;
```
schlenkibus: returning nullptr if the soundtype is not Split was a bad idea in the first place

playground/src/testing/unit-tests/dual-mode-acceptance/LoadDualPartIntoDualPart.cpp
```diff
@@ -10,90 +10,60 @@
...
-}
-
 template <SoundType tType, VoiceGroup tLoadFromPart, VoiceGroup tLoadToVoiceGroup>
 void LoadDualPresetPartWithValueIntoInitDualSoundPart(Preset *preset, tControlPositionValue paramValue)
 {
-  WHEN("converted to " + toString(tType) + " load into " + toString(tLoadToVoiceGroup) + " from "
-       + toString(tLoadFromPart) + "of type " + toString(preset->getType()))
-  {
-    static_assert(tType == SoundType::Split || tType == SoundType::Layer, "");
+  //WHEN("converted to " + toString(tType) + " load into " + toString(tLoadToVoiceGroup) + " from "
```
schlenkibus: Adding sections indefinitely does lead to degraded test-performance very quickly. But not adding Sections leads to missing information if a test fails -> no way of knowing which parameter comparison failed for example

playground/src/groups/ParameterGroup.cpp
hhoegelo: findParameterByNumber is n * O(n) while findParameterById id n * O(log(n)) - so I would prefer it the way it is written here.

playground/src/parameters/UnisonVoicesParameter.cpp
hhoegelo: I'd say, updateScaling should be undoable: UnisonVoicesParameter::updateScaling(UNDO::Transaction* transaction, SoundType type)
The only user, as far as I see, is EditBuffer::initUnisonVoices(), which is called from undoableSetType, where we have a transaction already available (outside of the undo command, of course). The connection established in EditBuffer-Ctor ( onSoundTypeChanged(sigc::mem_fun(this, &EditBuffer::initUnisonVoices));) is not useful, as undoableSetType() can do the work already.

schlenkibus: fixed

playground/src/groups/ParameterGroup.cpp
hhoegelo: If you don't like the 'hacky' way, you can maybe hide it in a method "ParameterId ParameterId::toVoiceGroup(...) const"

playground/src/presets/PresetParameterGroup.cpp
```diff
@@ -33,6 +35,14 @@ PresetParameter *PresetParameterGroup::findParameterByID(ParameterId id) const
   return nullptr;
 }
 
+VoiceGroup PresetParameterGroup::getVoiceGroup() const
```
hhoegelo: I would have expected 'm_voiceGroup' to be returned here. Is it really possible that m_voiceGroup and the voicegroup of the first parameter in the group are different?

playground/src/presets/PresetParameterGroup.cpp
```diff
@@ -15,6 +15,8 @@ PresetParameterGroup::PresetParameterGroup(const ::ParameterGroup &other)
 {
   for(auto &g : other.getParameters())
     m_parameters[g->getID()] = std::make_unique<PresetParameter>(*g);
+
+  m_voiceGroup = other.getVoiceGroup();
 }
 
 PresetParameterGroup::PresetParameterGroup(const PresetParameterGroup &other)
```
hhoegelo: This ctor misses setting m_voiceGroup.

playground/src/presets/PresetParameterGroup.h
```diff
@@ -37,11 +37,14 @@ class PresetParameterGroup
...
   }
 
+  VoiceGroup getVoiceGroup() const;
+
   // algorithm
   void writeDiff(Writer &writer, const GroupId &groupId, const PresetParameterGroup *other) const;
 
  protected:
   std::map<ParameterId, ParameterPtr> m_parameters;
+  VoiceGroup m_voiceGroup;
```
hhoegelo: m_voiceGroup is uninitialized, the default ctor does not initialize it either.

playground/src/groups/ParameterGroup.cpp
schlenkibus: I'm fine with how it is implemented after the m_voicegroup fixups

schlenkibus:
urgent don't merge yet, the added assertion in preset parameter group failed while testing. will investigate later

PR #1418, opened 2020-01-13
fixup missing parameters in preset to default

schlenkibus:
+ add log for future bug tracing

### DIFF
--- a/playground/src/groups/ParameterGroup.cpp
+++ b/playground/src/groups/ParameterGroup.cpp
@@ -98,7 +98,7 @@ void ParameterGroup::copyFrom(UNDO::Transaction *transaction, const PresetParame
       nltools::Log::warning("could not copy", myParameter->getID().toString(), " not present in other! defaulting...",
                             __LINE__);
       myParameter->setDefaultFromHwui(transaction);
-      if(auto mod = dynamic_cast<ModulateableParameter*>(g))
+      if(auto mod = dynamic_cast<ModulateableParameter*>(myParameter))
       {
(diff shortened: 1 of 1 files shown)


playground/src/groups/ParameterGroup.cpp
```diff
@@ -92,6 +94,23 @@ void ParameterGroup::copyFrom(UNDO::Transaction *transaction, const PresetParame
     {
       myParameter->copyFrom(transaction, otherParameter);
     }
+    else
+    {
```
hhoegelo: You can (and should) use Parameter::loadDefault which is overridden for modulateable parameters and macro parameters (and pedals and ribbons) already. Also it doesn't claim to be invoked by HWUI.

playground/src/groups/ParameterGroup.cpp
```diff
@@ -202,11 +221,28 @@ bool ParameterGroup::areAllParametersLocked() const
...
   {
-    if(auto c = other->findParameterByID({ g->getID().getNumber(), other->getVoiceGroup() }))
+    if(auto c = other->findParameterByID({ myParameter->getID().getNumber(), other->getVoiceGroup() }))
+    {
+      myParameter->copyFrom(transaction, c);
+    }
+    else
     {
-      g->copyFrom(transaction, c);
+      nltools::Log::warning("could not copy", myParameter->getID().toString(), " not present in other! defaulting...",
```
hhoegelo: please use Parameter::loadDefault()

PR #1441, opened 2020-01-30
1435

hhoegelo:


MatthiasSeeber:
Seems okay to me, compiles and runs as expected. However, shouldn't I get an error when starting the audio engine without the --heartbeat command line arg? (I tried with and without, both work, so it doesn't seem to make a difference)

MatthiasSeeber:
Just checked my Audio Interface MIDI LED while running the engine with and without --heartbeat arg. When using it, engine sends MIDI data. When omitting it, engine doesn't send. So, we just have to ensure that the engine always starts with the arg and everything should be ok.

schlenkibus:
> Just checked my Audio Interface MIDI LED while running the engine with and without --heartbeat arg. When using it, engine sends MIDI data. When omitting it, engine doesn't send. So, we just have to ensure that the engine always starts with the arg and everything should be ok.

If we always want to send the Midi Heartbeat do we need an passable argument then?

KlausStrohhaecker:
Checked on a system using the latest "MS1.7 NAMM Preview" code & system

PR #1457, opened 2020-02-06
Playground post namm fixes

schlenkibus:


playground/src/groups/OscillatorBGroup.cpp
```diff
@@ -72,9 +73,12 @@ void OscillatorBGroup::init()
   appendParameter(new Parameter(this, { 100, getVoiceGroup() }, ScaleConverter::get<Linear100PercentScaleConverter>(),
                                 0, 100, 1000));
 
-  appendParameter(new ModulateableParameter(this, { 302, getVoiceGroup() },
-                                            ScaleConverter::get<PhaseBipolar180DegreeScaleConverter>(), 0, 180, 1800));
+  appendParameter(new ModulateableParameterWithUnusualModUnit(
+      this, { 302, getVoiceGroup() }, ScaleConverter::get<PhaseBipolar180DegreeScaleConverter>(),
```
MatthiasSeeber: `ScaleConverter::get<PhaseBipolar180DegreeScaleConverter>()`
should be
`ScaleConverter::get<PhaseBipolar360DegreeScaleConverter>()`
instead, like used for Osc A

PR #1488, opened 2020-02-20
integrating lpc project into build system

hhoegelo:
Basically, I managed to integrate the LPC project into our build system. You have to install a cross compiler, though: sudo apt-get install gcc-arm-none-eabi.

@KlausStrohhaecker I had trouble with symbols defined twice: 

- USB0_IRQHandler is defined in nl_usb_host.c and in nl_usb_core.c
- sendAssertionToBB is defined in cpu/nl_assert.h/c and  dbg/nl_assert.h/c

Do you have an idea why this symbols exist twice and how your ide can deal with that?

Next step: moving the compiler into a docker, so the cross compiler doesn't have to be installed in the host system.


KlausStrohhaecker:
1) Namespace conflict. USB0_IRQHandler in nl_usb_host.c is local (not exported in *.h) but not declared static. As it turns out, it isn't even used -- dead code. LPC runs fine with it commented out, it seems.

2) sendAssertionToBB is used in an "NL_ASSERT" macro that seeminingly was used for debugging in the very early days. It's called inconsistently anyway, and bridge/pg don't do anything with the message sent. I removed the the whole thing and saw no problem.

Note that LPC code is compiled with "-fno-common" in the LPCXpresso setup for the M0 and M4 main modules (but not for the two instances of the libraries), so shared names don't actually result in shared physical objects with corresponding warnings, probably to hide some conflicts from less than perfect coding style.

As for next steps, once the things basically compiles we would need to export and apply the options from LPCXpresso project. And of course the biggest prerequisite is: do we get a raw binary output artifact (actually two, one for each core) that is compatible with the mxli progging tool? (additionally, do we get output in *.axf output format, which can be used inside LPCXpresso, for double-checking).
If so, it would be best to check if we get bitwise identical outputs when using the same compiler and project options. 
If true, we know we can still use LPCXpresso for development and notably debugging and use the automated build artefact safely as well, in production. If not, proper means of debugging, and more general, shifting the whole toolchain to Linux, must be checked. Unless these questions are answered first, it IMHO makes little sense to integrate the build process. Personally, I'm a bit reluctant to change current tool-chain for the moment, at least not until the current LPC topics are finished. Don't change horses in the middle of the stream ;-)

PR #1514, opened 2020-03-06
Audio engine new features

MatthiasSeeber:
- added two new parameters (voice group: fade from, fade range) - see ParameterList
- changed mono legato parameter (now possessing four distinct states) - see ParameterList and #1508 (btw, a conversion for the new param behavior is NOT required)
(these should to be implemented on playground-side)
- improved mono voice allocation accordingly
- envelopes now ready for bipolar level velocities
(audio engine still lacks proper consumption of preset/parameter messages for the two new ones - this can be implemented when the playground has integrated these)


projects/audio-engine/src/synth/c15-audio-engine/ae_poly_section.cpp
```diff
@@ -255,6 +261,67 @@ float PolySection::getVoiceGroupVolume()
...
 }
 
+void PolySection::evalVoiceFade(const float _from, const float _range)
+{
+  const int32_t start = m_fadeStart;
+  const int32_t fadeStart
+      = start + (m_fadeIncrement * (1 + (m_fadeStart + (m_fadeIncrement * static_cast<int32_t>(_from)))));
+  const int32_t fadeEnd = fadeStart + (m_fadeIncrement * static_cast<int32_t>(_range));
+  const float slope = 1.0f / (_range + 1.0f);
+  float fade = 1.0f, fadeValue;
```
hhoegelo: fadeValue is uninitialized,
also, it is only used inside the for loop ln 283ff. Consider declaring the variable locally in the for-scope

PR #1531, opened 2020-03-10
fixes #1516 & fixes #1512

### LINKED ISSUE #1512: BOLED: Preset Screen: Bank selektieren, Edit-Button drücken: Das Menü erscheint mit Verzögerung.

### LINKED ISSUE #1516: Available Update is not indicated
The current 'Master' does not show the presence of an update on USB, even though an update is executed once the C15 is restarted.

`UI: System Info -> Update Available NO`

schlenkibus:


projects/playground/src/proxies/hwui/panel-unit/boled/setup/UpdateAvailableView.cpp
```diff
@@ -29,5 +30,10 @@ bool UpdateAvailableView::poll()
 
 bool UpdateAvailableView::updateExists()
 {
-  return g_file_test(c_updateFileName, GFileTest::G_FILE_TEST_EXISTS);
+#ifdef _DEVELOPMENT_PC
+  return false;
+#endif
+  SpawnCommandLine cmd(
+      "ssh -o \"StrictHostKeyChecking=no\" root@192.168.10.11 'ls /mnt/usb-stick/nonlinear-c15-update.tar'");
```
hhoegelo: Either use c_updateFileName here or delete it above

projects/bbbb/src/io/USBManager.cpp
```diff
@@ -0,0 +1,38 @@
...
+        nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
+
+        std::this_thread::sleep_for(std::chrono::seconds(5));
+      }
+    } }
+{
+}
+
+USBManager::~USBManager()
+{
```
hhoegelo: So the destructor will add up to 5 seconds to the system shutdown time, right? How about having a std::condition_variable to wait on? It could be notified from destructor and waited on with a timeout?

Issue #1517, opened 2020-03-06
Friendlier OLED Messages


AntonSchmied:
The initial message could be friendlier than just 'initializing...'

How about Random Fortune Cookie Notes or Wise Quotes?
Or custom preset messages depending on the geographical location of the User?
 
Or simply something like "Welcome, those who love to play keys!" would do just fine!

hhoegelo:
@nonlinear-labs 
How about:

- Startup
 - Soled: "Starting C15."
 - Boled: "Welcome, those who love to play keys!"
- Shutdown:
 - Soled: "Shutting down C15."
 - Boled: "Bye Bye, those who love to play keys!"

AntonSchmied:
I like it! However I would suggest the Boled just shows "Bye, Bye!" on shut down. I feel it's a bit to much with "the keys".

nonlinear-labs:
I agree with Anton.

PR #1534, opened 2020-03-11
#1517

hhoegelo:


AntonSchmied:
- It would be correct to say/write "Shutting C15 down"

- I suggest messages which indicate a process end with "..." instead of ".", so
  "Starting C15..." and "Shutting C15 down..."

- Why is the startup message part of a service and the shut down message part of a script?

- And i trust you on the  double check in the c15-ota-update.sh

hhoegelo:
> * It would be correct to say/write "Shutting C15 down"

Okay

> * I suggest messages which indicate a process end with "..." instead of ".", so
>   "Starting C15..." and "Shutting C15 down..."

Done.

> * Why is the startup message part of a service and the shut down message part of a script?

I just could not find more symmetric places. 

> * And i trust you on the  double check in the c15-ota-update.sh

Yes, you can: First, I try to connect to the device without begging the developer for his password. Only if this fails, I will show a password prompt.

PR #1566, opened 2020-03-16
Master modulateables

MatthiasSeeber:
We re-classify the two Master parameters (247 - Volume and 248 - Tune) to be modulateable.
Tasks before merging:
- [x] param-db: re-classify Parameters
- [x] nltools: provide parameter structures
- [x] audio-engine: provide proper preset consumption
- [x] playground: implement re-classification (resolution and amount aspects are comparable to Part Group parameters: 358 - Volume and 360 - Tune)
- [x] nltools, audio-engine: testing and 
- [x] cleanup
- [x] playground: correction of Master Tune mod amount scale conversion (see below)


MatthiasSeeber:
@nonlinear-labs 
With this Pull Request, the two Master Volume and Tune parameters will become modulateable. However, there remains the question to what should happen with the Modulation Aspects, if a Single sound is loaded into a Dual Part. Shall the modulation remain on the Master parameter or instead be assigned to the Part parameter?

MatthiasSeeber:
Boled-Mon-Mar-16-16:36:55-2020
Boled-Mon-Mar-16-16:35:47-2020
@schlenkibus  the Master Tune parameter should possess (like Part Tune) a Mod Amount unit in semitones (st)

MatthiasSeeber:
Screenshot from 2020-03-16 17-13-35
@schlenkibus both proposals (if constexpr and range based for loop) seem to break my qt creator code model (which may be not up-to-date), which is inconvenient. Nevertheless, it compiles without issues and seems to work properly.

schlenkibus:
> Screenshot from 2020-03-16 17-13-35
> @schlenkibus both proposals (if constexpr and range based for loop) seem to break my qt creator code model (which may be not up-to-date), which is inconvenient. Nevertheless, it compiles without issues and seems to work properly.

Lets talk later about updating your toolchain to support the latest compiler / clang format

schlenkibus:
@MatthiasSeeber I updated the modulation amount for master tune

MatthiasSeeber:
Boled-Tue-Mar-17-12:24:47-2020
Boled-Tue-Mar-17-12:25:10-2020
We are almost there, but not quite. The Master Tune Amount is showing half its actual value (see Part Tune).

PR #1578, opened 2020-03-17
1568 - Animation Vs Action

schlenkibus:
@hhoegelo 

adresses #1568 

I changed implementation of animated generic item to execute the action only after the animation played, aborting the animation by eg changing focus will not execute the action. If this behaviour is wanted problems arise about open transactions and creating transactions inside the animation callback. EG: if i press convert to single and be quick about selecting a parameter the sound type is not changed. But if we wanted to execute the callback from the destructor of the animation then the open transaction from setFocusAndMode -> reset -> animation destructor will lead to an crash. This is not an trivial problem.


hhoegelo:
Let's skype about this. I don't understand the problem.

PR #1583, opened 2020-03-18
AudioEngine - dynamic SampleRate and DSP Optimization

MatthiasSeeber:
- fixes a Scale::BaseKey smoothing issue
- fixes an ancient GapFilter bug (HP Filter, left Channel)
- [x] init flexibly by sampleRate (48k, 96k only) -- static polyphony (24)
- [x] optimize poly soundgen fluctuation by providing parallel implementation
- [x] minimize smoothers: avoid conditional branching and redundant arithmetics
- [x] test experimental smoothers (see smoother_handle.h, proposals 2, 3a and 3b seem promising)
- [x] try to make smoothing buffers parallel
- [x] locate cause of buffer underruns: primarily denormal floats (ae_mono_section)
- [x] properly initialize poly/mono components and host/sections/utilities
- [x] get rid of denormal floats
- [x] found an issue within gap filter hp filter (left channel) - verified and approved
- [x] evaluate filter coefficient generation for -funsafe-math (poly and mono components)
- [x] integrate acceptance-tets - awaits approval
- [x] How to handle the three remaining voice loops in `PolyCombFilter::apply`? - maybe solve later
- [x] Integrate `NlToolbox::Constants::DNC_const` into `MonoReverb::apply` - done
- [x] cleanup

at last: necessity is currently unclear:

- [x] poly_soundgen: fluctuation: try to avoid rand_int overflow (implement 64-bit version - wrapping around 32-bit) - not necessary


MatthiasSeeber:
"normalized alternatives" was meant to be "un-normalized alternatives"
anyway, smoothers now are prepared for implementing more simplistic variation

MatthiasSeeber:
I implemented and tested a few alternatives for parameter smoothers, trying to reduce rendering arithmetics to a minimum.

The current implementation has the bonus that directions (raising / falling signals) are not relevant, due to the normalization of the transition. Getting rid of the normalization (and thus avoiding some calculations) requires paying attention to the signal direction, so new issues are introduced.

The first approach of trying to catch the point where the signal crosses its destination seems not possible (possibly due to wide-range variations of the signal growth), but I managed to find a solution (which is currently running and seems to work fine).

I am however not really sure if this really saves CPU in total:
- smoothers are now permanently running (so, it is expected that the mean consumption will raise)
- avoiding arithmetics in rendering procedure introduced conditionals (so consumption may have just been shifted)

Anyway, we should run some tests on that.

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -351,3 +351,35 @@ template <typename T, size_t size> inline T sumUp(const ParallelData<T, size> &i
 
   return ret;
 }
+
+template <typename T, size_t size>
+inline ParallelData<int32_t, size> phaseIsEdge(const ParallelData<T, size> &absolute_phase_difference)
```
hhoegelo: This function can be implemented like this:
return (absolute_phase_difference > 0.5f) * static_cast<uint32_t>(-1);

I tested for correctness (both implementations yield same results). Saves about 41% cpu.

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -351,3 +351,35 @@ template <typename T, size_t size> inline T sumUp(const ParallelData<T, size> &i
...
+    else
+    {
+      ret[i] = 0;
+    }
+  }
+
+  return ret;
+}
+
+template <typename TOut, typename T, size_t size> ParallelData<TOut, size> poly_cast(const ParallelData<T, size> &in)
```
hhoegelo: a cast operator already exists, so instead of writing 

poly_cast<float>(m_OscA_randVal_int)

you can just use

static_cast<PolyValue>(m_OscA_randVal_int)

MatthiasSeeber:
We now use smoothers that are still normalized (so, each sample we have to multiply with the segment difference), but the smoothers are state-less and therefore still simpler as the original implementation.

MatthiasSeeber:
- now, all member variables of audio components initialize to something
- resizing vectors initialize after resizing within init()

MatthiasSeeber:
The avalanche of X-Runs seems to have vanished now.

projects/audio-engine/src/synth/c15-audio-engine/ae_mono_gapfilter.cpp
```diff
@@ -97,132 +98,145 @@ void Engine::MonoGapFilter::set(MonoSignals &_signals)
...
   // biquad hp l
   float frequency = _signals.get(C15::Signals::Mono_Signals::Gap_Flt_HF_L);
-  tmpVar_1 = (frequency - m_freqScale_min) * (1.0f / (m_freqScale_max - m_freqScale_min));
-  tmpVar_1 = std::clamp(frequency, -1.0f, 1.0f);
+  tmpVar_1 = (frequency - m_freqScale_min) * m_freqScale_norm;
+  // tmpVar_1 = std::clamp(frequency, -1.0f, 1.0f);
+  // ??? this seems weird (other comparable calculations use tmpVar_1 instead of frequency)
+  // - further, it seems unlikely the frequency will ever fall below zero (expected freq range: [11.6 kHz ... nyquist])
+  // - anyway, shouldn't this be:
+  tmpVar_1 = std::clamp(tmpVar_1, -1.0f, 1.0f);  // (no noticeable differences found so far...)
```
MatthiasSeeber: @AntonSchmied 
It seems there was a well hidden bug in the Gap Filter (I also spotted this in the old **nlaudio** repo) - can you confirm this / do you approve of this "correction"?

projects/audio-engine/src/synth/c15-audio-engine/ae_mono_reverb.cpp
```diff
@@ -272,25 +273,25 @@ void Engine::MonoReverb::apply(MonoSignals &_signals, const float _rawSample_L,
...
       m_hp_omega_ramp += m_smooth_inc;
     }
-    m_hp_a0 = 1.f / (m_hp_omega + 1.f);
-    m_hp_a1 = m_hp_omega - 1.f;
+    m_hp_a0 = 1.0f / (m_hp_omega + 1.0f);
+    m_hp_a1 = m_hp_omega - 1.0f;
     tmpVar = m_lfo_stateVar_1 + m_lfo_omega_1;
     tmpVar = tmpVar - std::round(tmpVar);
-    m_lfo_stateVar_1 = tmpVar;
+    m_lfo_stateVar_1 = tmpVar;  // ??? lfo state vars without dnc_constant?
```
MatthiasSeeber: @AntonSchmied 
In contrast to the other Mono Effects, the Reverb filter-delay combos seem to lack usage of **NlToolbox::Constants::DNC_const**. Do you think they could be implemented?

MatthiasSeeber: Btw, the shown lfo_stateVars are a bad example. This is much more relevant further below, beginning at line 326 (the filter stateVars).

projects/audio-engine/src/synth/c15-audio-engine/ae_mono_reverb.cpp
```diff
@@ -323,13 +323,15 @@ void Engine::MonoReverb::apply(MonoSignals &_signals, const float _rawSample_L,
...
+  m_lp_stateVar_L = wetSample_L;
+  // ??? filter state vars without dnc_constant?
 
   wetSample_L = (wetSample_L + tmpVar) * m_lp_omega;
 
   wetSample_L = (wetSample_L - m_hp_stateVar_L * m_hp_a1) * m_hp_a0;
   tmpVar = m_hp_stateVar_L;
-  m_hp_stateVar_L = wetSample_L;  // ??? lfo state vars without dnc_constant?
+  m_hp_stateVar_L = wetSample_L;
+  // ??? filter state vars without dnc_constant?
```
MatthiasSeeber: @AntonSchmied 
Unlike other components, the Reverb doesn't use `NlToolbox::Constants::DNC_const`.
Do you think it's possible to implement that here as well?
(Currently, the Reverb isn't safe against _denormal_ floats.)

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
```diff
@@ -68,7 +68,7 @@ void Engine::PolyCombFilter::apply(PolySignals &_signals, const PolyValue &_samp
   m_apStateVar_1 = tmpOut;
   m_apStateVar_4 = m_apStateVar_3;
   m_apStateVar_3 = m_out;
-  // para d
+  // para d --- unfortunately still scalar, via voice loop (1 / 3)
   for(uint32_t v = 0; v < C15::Config::local_polyphony; v++)
   {
```
MatthiasSeeber: We may want to try getting rid of this remaining voice loop.

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
```diff
@@ -104,6 +104,7 @@ void Engine::PolyCombFilter::apply(PolySignals &_signals, const PolyValue &_samp
   tmpSmooth += (phaseMod * tmpSmooth);
   // delay
   auto holdsample = m_out;  // for Bypass
+  // delay buffer "write" --- unfortunately still scalar, via voice loop (2 / 3)
   for(uint32_t v = 0; v < C15::Config::local_polyphony; v++)
   {
     m_buffer[m_buffer_indx][v] = m_out[v];
```
MatthiasSeeber: We may want to try getting rid of this remaining voice loop.

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
```diff
@@ -124,8 +125,10 @@ void Engine::PolyCombFilter::apply(PolySignals &_signals, const PolyValue &_samp
...
   ind_tp2 &= m_buffer_sz_m1;
-  PolyValue fract, sample_tm1, sample_t0, sample_tp1, sample_tp2;
-  fract = tmpSmooth;
+  //  PolyValue fract, sample_tm1, sample_t0, sample_tp1, sample_tp2;
+  PolyValue sample_tm1, sample_t0, sample_tp1, sample_tp2;
+  // fract = tmpSmooth; // unused here
+  //  delay buffer "read" --- unfortunately still scalar, via voice loop (3 / 3)
   for(uint32_t i = 0; i < C15::Config::local_polyphony; i++)
   {
     sample_tm1[i] = m_buffer[ind_tm1[i]][i];
```
MatthiasSeeber: We may want to try getting rid of this remaining voice loop. Although I'm not entirely sure if this is achievable.

projects/audio-engine/src/synth/c15-audio-engine/ae_mono_gapfilter.cpp
AntonSchmied: aaaah clamping 'tmpVar_1' instead of 'frequency' you are! yeeeees! 
Approved!

projects/audio-engine/src/synth/c15-audio-engine/ae_mono_reverb.cpp
AntonSchmied: Oooooh snap! In the Ensemble, they are hidden in the [z^-1] makro ... forgot to put them in!  
So, yes, they should be there, same goes for the "Loop Filter R"!

image

AntonSchmied:
Concerning the _DNC_const_ in general: We have currently 259 _DNC_const_ usages, so about the same amount of additions each sample loop.
@hhoegelo suggested a proccess which resets the denormal floats to zeros, if I understood him correctly. Is this worth while looking upon, to optimzie the dsp even more?

hhoegelo:
I really don't know enough about DSP, but when it comes to optimization I can confirm that 259 x 24 additions take longer than 0 x 24 additions ;-)
So, if setting the flush-to-zero flag yields the same (or comparable) result, I strongly suggest to do so. If I remember correctly, the SIMD units don't care about denormals anyway, so adding the DNC_const on ParallelData<> is pure waste. But this has to be investigated, too.

PR #1598, opened 2020-03-25
first attempt to parallelize smoothing

hhoegelo:


hhoegelo:
@MatthiasSeeber  it is completely untested yet. I will test it right now.

projects/audio-engine/src/synth/c15-audio-engine/smoother_handle.h
```diff
@@ -147,63 +139,59 @@ template <class Sync, class Audio, class Fast, class Slow> class SmootherHandle
...
-      m_fast[i].reset();
-    }
-    for(uint32_t i = 0; i < static_cast<uint32_t>(Slow::_LENGTH_); i++)
-    {
-      m_slow[i].reset();
-    }
+    m_sync = 0.0f;
+    m_audio.reset();
+    m_fast.reset();
+    m_slow.reset();
```
MatthiasSeeber: Function reset() requires idx argument.
Furthermore, it doesn't seem to iterate through contained elements.

hhoegelo: Indeed. It still compiles, because reset() is never called. So I will remove the method.

PR #1611, opened 2020-03-31
fixes #1501 -> maybe i am missing more parameters where on/off is app…

### LINKED ISSUE #1501: WebUI - Slider behavior for "Switches"
In the Parameter Tab, editing switch-like Parameters (Osc A/B Reset, Mono Enable) feels a little exhausting, as the user has to drag over 70% of the Slider before changes become effective. (It could be more convenient, if the Slider is more sensitive when editing switches).

schlenkibus:
…licable?

### DIFF
--- a/projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
+++ b/projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
@@ -18,29 +18,28 @@ public class ValueUpdater extends Updater {
 		String bipolar = getChildText(root, "bipolar");
 		String scaling = getChildText(root, "scaling");
 		String defaultValue = getChildText(root, "default");
-		String isBoolean = getChildText(root, "boolean");
-
-		if(!isBoolean.isEmpty()) {
-			BooleanDataModelEntity bp = target.metaData.isBoolean;
-			bp.setValue(isBoolean.equals("1")? BooleanValues.on : BooleanValues.off);
-		}
 
 		if (!bipolar.isEmpty()) {
 			BooleanDataModelEntity bp = target.metaData.bipolar;
 			bp.setValue(bipolar.equals("1") ? BooleanValues.on : BooleanValues.off);
 		}
 
-		if (!coarse.isEmpty())
+		if (!coarse.isEmpty()) {
 			target.metaData.coarseDenominator.setValue((int) Double.parseDouble(coarse));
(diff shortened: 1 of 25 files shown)


projects/playground/src/parameters/ValueRange.h
```diff
@@ -36,6 +36,11 @@ template <typename tValue> class ValueRange
     return m_min == -m_max;
   }
 
+  bool isBoolean() const
```
hhoegelo, line 39: Hmmmm, I don't get his.

projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
```diff
@@ -18,6 +18,12 @@ public void doUpdate() {
 		String bipolar = getChildText(root, "bipolar");
 		String scaling = getChildText(root, "scaling");
 		String defaultValue = getChildText(root, "default");
+		String isBoolean = getChildText(root, "boolean");
```
hhoegelo, line 21: Couldn't we deduce the 'booleaness' from the number of steps?

projects/playground/src/parameters/scale-converters/EnvelopeAttackDecayTimeMSScaleConverter.cpp
```diff
@@ -43,9 +43,4 @@ Glib::ustring EnvelopeAttackDecayTimeMSScaleConverter::controlPositionToDisplayJ
 bool EnvelopeAttackDecayTimeMSScaleConverter::isBiPolar() const
```
hhoegelo, line 43: Is this method still needed?

projects/playground/src/parameters/value/RawValue.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "playground.h"
+#include <optional>
```
hhoegelo, line 4: can prolly be removed?

projects/playground/src/parameters/value/RawValue.cpp
```diff
@@ -79,12 +79,12 @@ bool RawValue::isBiPolar() const
 
 void RawValue::setIsBoolean(bool v)
```
hhoegelo, line 80: are this methods used?

schlenkibus: The Router Parameters use this function

PR #1612, opened 2020-03-31
tackling #1600

MatthiasSeeber:
Trying to tackle and make sense of Recall-Related Bugs (see #1600):
- fadepoint: sometimes fails to recover, staying muted
- voice reset: recalling between Layer Presets with the same Unison Voices resets Envelopes nevertheless (it shouldn't)
- wrong sound: sometimes, loaded Presets seem to fail to produce the right sound (observed with Layer <--> Layer, Layer <--> Single, Single <--> Single Presets)

### DIFF
--- a/projects/audio-engine/src/synth/c15-audio-engine/ae_fadepoint.cpp
+++ b/projects/audio-engine/src/synth/c15-audio-engine/ae_fadepoint.cpp
@@ -29,9 +29,24 @@ void atomic_fade_table::init(const float _samplerate)
   }
 }
 
-uint32_t atomic_fade_table::getTargetRampIndex() const
+bool atomic_fade_table::getTaskStatus()
 {
-  return m_muteTasks ? 0 : m_finalMuteRampIndex;
+  const uint32_t targetRampIndex = m_muteTasks ? 0 : m_finalMuteRampIndex;
+  if(m_currentMuteRampIndex != targetRampIndex)
+  {
+    if(m_currentMuteRampIndex > targetRampIndex)
+    {
+      // fade out
+      m_currentMuteRampIndex--;
+      return m_currentMuteRampIndex == targetRampIndex;  // return true if fade out completed
+    }
+    else
+    {
+      // fade in
+      m_currentMuteRampIndex++;
+    }
+  }
+  return false;
 }
 
 void atomic_fade_table::setTask(const MuteTask _task)
--- a/projects/audio-engine/src/synth/c15-audio-engine/ae_fadepoint.cpp
+++ b/projects/audio-engine/src/synth/c15-audio-engine/ae_fadepoint.cpp
@@ -13,69 +13,69 @@
 
 // current FadeTable implementation
 
-ae_fade_table::ae_fade_table()
-{
-}
+//ae_fade_table::ae_fade_table()
+//{
+//}
 
-void ae_fade_table::init(const float _samplerate)
-{
-  float fade_time = 1e-3f * C15::Config::fade_time_ms * _samplerate, fade_dx = 1.0f / (fade_time - 1.0f);
+//void ae_fade_table::init(const float _samplerate)
+//{
+//  float fade_time = 1e-3f * C15::Config::fade_time_ms * _samplerate, fade_dx = 1.0f / (fade_time - 1.0f);
 
-  m_table_index = 0;
-  m_trigger = false;
-  m_trigger_index = static_cast<uint32_t>(fade_time);
-  m_data.resize(2 * m_trigger_index);
-  m_event = FadeEvent::None;
-  m_value = 0.0f;
+//  m_table_index = 0;
+//  m_trigger = false;
+//  m_trigger_index = static_cast<uint32_t>(fade_time);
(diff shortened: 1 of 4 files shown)


MatthiasSeeber:
So far, I found out:
- The ALSA Buffer Size has an influence on these effects. With **48K** and **49** Samples, the effects are far more likely to occur, than **48K** and **64** Samples. (may be a hint on _threading problems_)
- The Fade Point indeed seems to have an "Achilles' heel": it possesses a _state_ mechanism (intended for preventing interruptions of fade processess). For some reason, this state mechanism appears to be connected to all mentioned effects of the Bug description (see above).

projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -577,6 +577,7 @@ void dsp_host_dual::onPresetMessage(const nltools::msg::SinglePresetMessage& _ms
   if(m_glitch_suppression)
   {
     // glitch suppression: start outputMute fade
+    m_fade.stop();
     m_fade.enable(FadeEvent::RecallMute, 0);  // enable fader with event
     m_output_mute.pick(0);                    // pickup fade-out
   }
```
MatthiasSeeber: When forcing the Fade Point to always start (interrupted or not), the described effects are not observable anymore.

MatthiasSeeber:
I intended to fix the mute bug with this, but as I described, the two other mentioned bugs seem to have vanished as well. It seems quite odd to me, despite the seemingly pleasant result.
I propose to carefully test this before merging/closing the issue.

hhoegelo:
There is still a problem with onSettingGlitchSuppr(). If switched off, you have to reset the m_fade object. Otherwise it will be stuck in whatever state it currently is.

MatthiasSeeber:
Right, I tried this twice. It made matters even worse. Although when I tried, I had not touched the Fade Point. So, I'll try once again.
Plus, a correction: the Voice Reset bug still occurs. So, the fix seems a little more real and a little less "magical". I'm on it.

MatthiasSeeber:
So here we have a new approach for the Fade Point. It should be thread-safe and can cope with several Mute Events occuring in unlikely small time intervals (less than 6 milliseconds) - still expected to be virtually impossible to provoke as a user.
It furthermore appears that Recall-related X-Runs occur a little less, as well as audible "clicks" when recalling. Recalled presets sounding wrong seems to be solved by this implementation, although more tests on that should be done.
There is still an issue with Layer-Sounds using Unison - they still reset held Keys, even when Unison Voices stays the same (in which case held Keys should remain audible). This is also not the case for Single Presets.

projects/audio-engine/src/synth/c15-audio-engine/ae_fadepoint.cpp
```diff
@@ -3,104 +3,43 @@
...
-  m_data[2] = &m_one;
-  m_index = m_current = 2;
-}
-
-void ae_fader::pick(const uint32_t _index)
+uint32_t atomic_fade_table::getTargetRampIndex() const
 {
-  m_index = _index;
-  m_current = 1;
+  return m_muteTasks ? 0 : m_finalMuteRampIndex;
```
hhoegelo, line 34: m_finalMuteRampIndex is kind of redundant as it can be calculated in no time from "m_data.size() - 1" - but I don't mind if you wan't to have it explicitly here.

projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -869,10 +868,9 @@ void dsp_host_dual::localUnisonVoicesChg(const nltools::msg::UnmodulateableParam
       nltools::Log::info("unison_voices_edit(layer:", layerId, ", pos:", param->m_position, ")");
     }
     // application now via fade point
-    m_fade.enable(FadeEvent::UnisonMute, 0);
-    m_output_mute.pick(0);
-    m_output_mute.m_preloaded_layerId = layerId;
-    m_output_mute.m_preloaded_position = param->m_position;
+    m_new_fade.setTask(MuteTask_Trigger_Unison);
```
hhoegelo, line 871: This a race: you have to set up the m_preload_xyz members before you set the task. The task might be executed by the audio thread before setTask returns in the UI thread.

projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -888,10 +886,9 @@ void dsp_host_dual::localMonoEnableChg(const nltools::msg::UnmodulateableParamet
     }
     param->m_scaled = scale(param->m_scaling, param->m_position);
     // application now via fade point
-    m_fade.enable(FadeEvent::MonoMute, 0);
-    m_output_mute.pick(0);
-    m_output_mute.m_preloaded_layerId = layerId;
-    m_output_mute.m_preloaded_position = param->m_scaled;
+    m_new_fade.setTask(MuteTask_Trigger_Mono);
```
hhoegelo, line 889: Same here: prepare the task before you set it

PR #1616, opened 2020-03-31
Issue 1586

### LINKED ISSUE #1586: Slider "Fade From II" und Splitpoint II von rechts fuellen
gillt fuer webui und boled

schlenkibus:
fixes #1586


projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/editBuffer/BasicParameterModel.java
```diff
@@ -19,6 +19,7 @@
 	public DoubleDataModelEntity originalValue = new DoubleDataModelEntity();
 	public BooleanDataModelEntity locked = new BooleanDataModelEntity();
 	public StringDataModelEntity info = new StringDataModelEntity();
+	public BooleanDataModelEntity enableFillFromRight = new BooleanDataModelEntity();
```
hhoegelo, line 22: I'd prefer to not spread this UI detail over the whole system.

I think for WebUI the ParameterPresenterProvider is the right place to add/set a boolean in the according presenter based on the parameterId (which already contains the voice group).

For boled, we can maybe utilize the Parameter::getVisualizationStyle() to customize the presentation?

projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/parameters/ParameterValueSlider.java
```diff
@@ -34,7 +39,24 @@ public void draw(Context2d ctx, int invalidationMask) {
 			if (isBiPolar)
 				drawCenterMark(ctx);
 
-			drawIndicator(ctx, value, isBiPolar);
+			boolean fillRight = false;
```
hhoegelo, line 42: Cannot we use presenter.fillFromRightEnabled here?

projects/playground/src/proxies/hwui/controls/BarSlider.cpp
```diff
@@ -72,3 +75,27 @@ bool BarSlider::drawCenterMark() const
 {
   return true;
 }
+
+bool BarSlider::shouldDrawParameterFromRight(Parameter *parameter)
```
hhoegelo, line 79: is this method still needed?

PR #1622, opened 2020-04-01
fixes #1585

### LINKED ISSUE #1585: Master Modulateables - integrate Mod Aspects into Dual mode conversions
- [x] when loading a Single preset into a Part: take Master Volume/Tune Mod Aspects and move them to Part Volume/Tune

- [x] when converting a Single preset into Split/Layer: take Master Volume/Tune Mod Aspects and move them to both Parts Volume/Tune (seems the most consistent to me)

- [x] when loading a Split/Layer Preset Part into a Part: ignore Master Volume/Tune completely (as the Part should already have its proper conversions/assignments)

- [x] when converting a Split/Layer Preset back to Single: as far as I understand, the selected Part will be chosen for the resulting Single preset. So, take Part Volume/Tune Mod Aspects (of the selected Part) and move them to Master Volume/Tune

- [x] @nonlinear-labs please verify.

schlenkibus:
added tests for new/existing behaviour, removed error message because it was wrong and not providing value

closes #1585


projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -763,6 +758,10 @@ void EditBuffer::copyAndInitGlobalMasterGroupToPartMasterGroups(UNDO::Transactio
 
   for(auto &parameter : global->getParameters())
   {
+    if(auto modParam = dynamic_cast<ModulateableParameter *>(parameter))
```
hhoegelo, line 761: maybe you can just use loadDefault() here?

schlenkibus: loadDefault is not virtual and will not set modualtion-aspects to default values

schlenkibus: making loadDefault virtual will break the curretn to default functionality. I would prefer to default the parameters this way

hhoegelo: virtual void Parameter::loadDefault(UNDO::Transaction *transaction);

is indeed virtual

Issue #1509, opened 2020-03-04
Settings.xml is not ported from BBB to ePC correctly


AntonSchmied:
When an update is carried out to the current 'Master', which involves moving the Settings.xml from BBB to ePC /persistent/ the settings are lost/overwritten after the update is done. 
- Setting "SSID" is not existent 
- Device Name receives a random Number   

UPDATE:
It has been observed, that the settings.xml is copied correctly from BBB to ePC, when updating from 1.5 to 1.7. However, it is overwritten when the ePC is rebooted.

AntonSchmied:
If someone points me in the right direction, I could take a look at this! 
Cause this bug is still there..

AntonSchmied:
I'll fix it!

PR #1640, opened 2020-04-03
BBB update: Playground service is stopped before the files are moved …

AntonSchmied:
…from BBB to ePC, to avoid overwriting of settings.xml and presets
This resolves Issue #1509


AntonSchmied:
unless of course it makes more sense NOT to abort the update procedure completely, if we cannot stop the playground service.

hhoegelo:
@AntonSchmied  I agree - the update process should NOT abort if stopping the playground fails.

schlenkibus:
when will systemctl return an error? If it returns an error if the service is not running then i would prefer not aborting the update.

AntonSchmied:
alright! let me just pull the latest from the master into the branch!

AntonSchmied:
After a walk I have had a further thought:
in bbb_update.sh we're are currently checking the connection to the ePC just for the sake of eventually moving files, which might not even be there. This check has nothing to do with rsync-ing rootfs of the BBB. 
Suggestion: 
BBB to ePC connection + /persistent mount checks + moving files should only be carried out if the settings.xml is still present on the BBB under root? Is that a sufficient criterion?

AntonSchmied:
@hhoegelo, Shlenki and I looked at the code and are happy with the procedure logic behind it. Could you check the code for sexiness please?

PR #1659, opened 2020-04-06
cleanup

schlenkibus:


projects/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/MacroControlParameterLayouts.cpp
```diff
@@ -257,9 +283,22 @@ MacroControlParameterSelectLayout2::MacroControlParameterSelectLayout2()
...
+      setButtonAText("I / II");
+    }
+    else
+    {
+      setButtonAText("HW Pos");
+    }
+  }
+  else if(getMode() == Mode::MacroControlValue)
+  {
+    setButtonAText("");
```
hhoegelo, line 300: this should be unnecessary, no?

MatthiasSeeber: Not sure about that but the application does but the Screen displays and behaves correctly.

schlenkibus: Agreed this is uneccessary as the button is initialized to empty

PR #1660, opened 2020-04-06
fixes hw amount boolean slider

### LINKED ISSUE #1588: WebUI: Die HW Amounts werden, auch wenn sie Schalter sind, in der Bauchbinde als kontinuierlich dargestellt/bedient.

schlenkibus:
closes #1588

### DIFF
--- a/projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
+++ b/projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
@@ -18,6 +18,12 @@ public class ValueUpdater extends Updater {
 		String bipolar = getChildText(root, "bipolar");
 		String scaling = getChildText(root, "scaling");
 		String defaultValue = getChildText(root, "default");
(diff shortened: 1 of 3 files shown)


projects/NonMaps/src/main/java/com/nonlinearlabs/client/dataModel/ValueUpdater.java
```diff
@@ -18,6 +18,12 @@ public void doUpdate() {
 		String bipolar = getChildText(root, "bipolar");
 		String scaling = getChildText(root, "scaling");
 		String defaultValue = getChildText(root, "default");
+		String bool = getChildText(root, "boolean");
+
+		if(!bool.isEmpty()) {
```
hhoegelo, line 23: why is the number of steps in coarse/fineDenominator not sufficient?

schlenkibus: The Scale Converter of Hardware Amounts is not boolean if non returning, checking for steps is not sufficent for these parameters

PR #1664, opened 2020-04-06
floating FadeRanges - Addendum

MatthiasSeeber:
Consider a Layer Sound in Part 1: FadeFrom = 2, FadeRange = 0.1
Subsequent Fade Levels for Keys are:
InterpolatedFloatingFadeRanges
- red was original approach (fade ranges were integers and had no fractional parts)
- green is current approach (can produce quite small fade values when fade range has small fractional)
- blue is addendum approach (using interpolation for better fade values when fade range has small fractional)

Here: FadeFrom = 2, FadeRange = 1.0
InterpolatedFloatingFadeRanges2

(all approaches produce equal fade values for fade ranges with a fractional part of zero)


MatthiasSeeber:
Also, I disabled Sample Counting by default. (This can and should be enabled when working within **acceptance-tests**)

MatthiasSeeber:
Now, Fade From and Fade Range possess parameter infos as well (they were the last two params without infos). Infos for HW Sources now use standard Quotes by escape sequences.

The file used for the two fade images is the following. Feel free to download it, unzip, open in it in your browser, and play with the "Range" parameter to get a feel for the implementation of this PR.

[FloatingFadeRanges.zip](https://github.com/nonlinear-labs-dev/C15/files/4443243/FloatingFadeRanges.zip)

PR #1690, opened 2020-04-15
Issue 1684

### LINKED ISSUE #1684: Dual Mode Conversion and (Load to Part Rules)
[ModeConversionRules-Rev2.pdf](https://github.com/nonlinear-labs-dev/C15/files/4473896/ModeConversionRules-Rev2.pdf)

schlenkibus:
closes #1684


projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -676,29 +677,84 @@ void EditBuffer::undoableConvertDualToSingle(UNDO::Transaction *transaction, Voi
...
   if(copyFrom != VoiceGroup::I)
     copyVoiceGroup(transaction, copyFrom, VoiceGroup::I);
 
+  initToFX(transaction);
+  initFadeFrom(transaction, SoundType::Single);
+  initSplitPoint(transaction);
+
+  forEachParameter(VoiceGroup::II, [&](Parameter *p) { p->loadDefault(transaction); });
+
+  {
```
hhoegelo, line 745: Why do we have the scope here?

projects/playground/src/parameters/scale-converters/LinearBipolar48StScaleConverter.cpp
```diff
@@ -9,3 +9,8 @@ LinearBipolar48StScaleConverter::LinearBipolar48StScaleConverter()
 LinearBipolar48StScaleConverter::~LinearBipolar48StScaleConverter()
 {
 }
+
+tControlPositionValue LinearBipolar48StScaleConverter::displayToControlPosition(const tDisplayValue& displayValue) const
+{
+  return displayValue / getControlPositionRange().getRangeWidth();
```
hhoegelo, line 15: I am not 100% sure, but this looks wrong to me.
getControlPositionRange() should be [-1 ... 1] for bipolar parameters, so range width is (1 - -1) = 2
DisplayValue should be - according to the name - [-48 St ... 48 St]. So the method will return values in the range [-24 ... 24], but it should yield values in the range [-1 ... 1].
Base class already implements it the way that I think is correct - what's the reason for overriding?

MatthiasSeeber:
- regarding Conversions: Cross FB and Fade Range:
Conversion_2

MatthiasSeeber:
Conversion_3

This is my review regarding Conversion so far.
There is an especially weird Unison Situation in Layer Sounds, as only on Part instead of two will sound as a unison cluster. I checked within the Audio Engine, but could not find a possible culprate so far..

projects/playground/src/parameters/scale-converters/LinearBipolar48StScaleConverter.cpp
schlenkibus: removed completely, the method was unused

projects/playground/src/presets/EditBuffer.cpp
schlenkibus: done

projects/playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/AppendOverwriteInsertButtonMenu.h
```diff
@@ -34,7 +34,7 @@ class AppendOverwriteInsertButtonMenu : public ButtonMenu
   void insertPreset(Bank* bank, size_t pos, bool modified);
   void appendPreset(Bank* bank, bool modified);
   void pushRenameScreen();
-  bool animate();
+  static bool animate();
```
hhoegelo, line 37: so far, I cannot see a 'static' user of this method - does it have to be static at all?

projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -1159,3 +1160,12 @@ void EditBuffer::undoableConvertSplitToLayer(UNDO::Transaction *transaction)
   calculateFadeParamsFromSplitPoint(transaction);
   undoableUnisonMonoLoadDefaults(transaction, VoiceGroup::II);
 }
+
+void EditBuffer::calculateSplitPointFromFadeParams(UNDO::Transaction *transaction)
+{
+  const auto f1 = findParameterByID({ 396, VoiceGroup::I })->getControlPositionValue();
+  const auto f2 = findParameterByID({ 396, VoiceGroup::II })->getControlPositionValue();
+  const auto meanFadeFrom = (f1 + f2) / 2.0;
```
MatthiasSeeber, line 1170: This seems totally appropriate, but is the possibility considered, that the ControlPosition of VoiceGroup::II may be inverted? (At least UI-wise, the Parameter Control Positions seem inverted, so maybe this inversion should be reflected here as well...)

MatthiasSeeber:
Convert_6
- Functionality seems okay now

PR #1717, opened 2020-04-21
expand tests to include all load to part aspects

### LINKED ISSUE #1716: Implement and Test Load To Part Rules
according to this rule set: 
[ModeConversionRules_Rev6.pdf](https://github.com/nonlinear-labs-dev/C15/files/4508841/ModeConversionRules_Rev6.pdf)

schlenkibus:
closes #1716


MatthiasSeeber:
LoadTests-2

schlenkibus:
@MatthiasSeeber Load to Part Aspects, Load to Part of Single Presets into Dual Edit-Buffer, has been reworked and found issues have been fixed.

MatthiasSeeber:
Load-Tests-03

:/

MatthiasSeeber:
Sorry, the Part Volume/Tune was wrongfully marked as flawed. Here are the real results:

Load-Tests-04

MatthiasSeeber:
Here is the latest version of my Test Bank.
- I re-ordered the Bank (Targets now below Sources) and re-named Presets for a better overview

[Load To Part Tests.zip](https://github.com/nonlinear-labs-dev/C15/files/4518244/Load.To.Part.Tests.zip)

MatthiasSeeber:
I just checked the two Single to Part of Split/Layer cases, using commit 1b7ad3486bea4ccea16c0a674daec98ee1776930

It tested Cross FB, To FX, Unison and Mono again. Here's my result:
Load-Tests-06

I'm afraid the current Mode Conversion Rules and the Checklist are not that helpful. Would it be worth approaching the issues entirely by pseudo-code?

projects/playground/src/parameters/ScopedLock.cpp
```diff
@@ -0,0 +1,73 @@
...
+  auto eb = Application::get().getPresetManager()->getEditBuffer();
+  if(auto param = eb->findParameterByID(id))
+  {
+    addLock(param);
+  }
+}
+
+void ScopedLock::addLock(Parameter* param)
+{
+  if(param)
```
hhoegelo, line 48: I suggest adding "! param->isLocked()" here: If locked, do nothing. If unlocked, call undoableLock() and add to member list or vector. So you don't have to store and care for the old lock-state. The unlocking in destructor can be unconditional, you can use a vector or list instead of the - rather expensive - map. Also, the container and the loops will become smaller.

projects/playground/src/parameters/ScopedLock.h
```diff
@@ -0,0 +1,24 @@
...
+  ~ScopedLock();
+
+  void addLock(const ParameterId& id);
+  void addLock(Parameter* param);
+  void addGroupLock(const GroupId& id);
+
+  void addLocks(const std::vector<Parameter *>& params);
+ private:
+  UNDO::Transaction* m_transaction;
+  std::vector<ParameterId> m_oldLockedParameters;
```
hhoegelo, line 23: Sorry for grumble again, but I'd prefer to store Parameter pointers here. As far as I can see, it would save quite some EditBuffer::findParameterById() calls without adding more risk.

projects/playground/src/parameters/Parameter.cpp
```diff
@@ -576,9 +576,11 @@ void Parameter::undoableRecallFromPreset(UNDO::Transaction *transaction)
 
 void Parameter::copyFrom(UNDO::Transaction *transaction, const Parameter *other)
 {
-  nltools_assertOnDevPC(other->getVoiceGroup() != getVoiceGroup());
-
-  setCpValue(transaction, Initiator::INDIRECT, other->getControlPositionValue(), false);
+  if(!isLocked())
+  {
+    nltools_assertOnDevPC(other->getVoiceGroup() != getVoiceGroup());
```
hhoegelo, line 581: Hmm, why is it not allowed that both parameters live in the same voice group?

schlenkibus: i cant think of a reason why this is here, i guess it can be removed

PR #1722, opened 2020-04-21
AudioEngine (Single) Glide Fix

### LINKED ISSUE #1719: AudioEngine - (Single) Mono Glide vs Unison Voices > 12
Glide Time seems to only affect one Voice Group - ParameterChanged and Recall.

The problem seems to arise with the current Key Event handling. Glides are only possible, when the Unison Index is zero, which would not be the case for the other Voice Group. The Test for starting a (Mono) Glide in the Poly Section needs another conditional variable (once and only first per Voice Group).

MatthiasSeeber:
closes #1719

### DIFF
--- a/projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
+++ b/projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
@@ -1165,6 +1165,7 @@ uint32_t dsp_host_dual::onSettingToneToggle()
 
 void dsp_host_dual::render()
 {
+  m_sample_counter += SAMPLE_COUNT;
   m_fade.evalTaskStatus();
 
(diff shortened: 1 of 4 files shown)


projects/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1165,6 +1165,7 @@ uint32_t dsp_host_dual::onSettingToneToggle()
 
 void dsp_host_dual::render()
 {
+  m_sample_counter += SAMPLE_COUNT;
```
MatthiasSeeber: I originally removed this line because it was present twice and thought, counting the sample at the end would make more sense. However, it then would not be counted if the Engine is muted.
Now, it always counts and is present only once, at the start.

PR #1729, opened 2020-04-23
In order to measure the cpu load our ePC is capable to handle,

hhoegelo:
this patch adds a CpuBurningSynth. If commandline parameter "-u N"
is used, the CpuBuringSynth will be installed instead of the C15Synth.
The synth will then create a signal of N stereo sines (= 2 x N sine
computations per sample frame). By experimenting with "-u"-parameter,
developers can measure, how buffer sizes (latency),
buffer underruns, and CPU load (CPU load can be monitored in the shell)
depend on each other.

### DIFF
--- a/projects/audio-engine/src/synth/CPUBurningSynth.cpp
+++ b/projects/audio-engine/src/synth/CPUBurningSynth.cpp
@@ -1,0 +1,32 @@
+#include "CPUBurningSynth.h"
+#include "AudioEngineOptions.h"
+#include <math.h>
+
+CPUBurningSynth::CPUBurningSynth(const AudioEngineOptions *options)
+    : Synth(options)
+{
+}
+
+void CPUBurningSynth::doMidi(const MidiEvent &)
+{
+}
+
+void CPUBurningSynth::doAudio(SampleFrame *target, size_t numFrames)
+{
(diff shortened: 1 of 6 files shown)


projects/audio-engine/src/synth/CPUBurningSynth.cpp
```diff
@@ -0,0 +1,32 @@
...
+    {
+      target[i].left += 0.000001f * std::sin(m_phase++ / 1000);
+      target[i].right += 0.000001f * std::sin(m_phase++ / 1000);
+
+      if(m_phase > 1000)
+        m_phase = 0;
+    }
+
+    target[i].left *= 0.000001f;
+    target[i].right *= 0.000001f;
```
MatthiasSeeber, line 30: I'm not sure if this is intended, but in regards to "synthesizing" it seems a little odd to increment the m_phase twice. But as this isn't supposed to create pleasant sounds, I think we're good here.

hhoegelo: I wanted to prevent the compiler from optimizing and instead calculate sine only once and re-use the value.

MatthiasSeeber: Seems legit.

PR #1730, opened 2020-04-23
Audio Engine - potential improvements

MatthiasSeeber:
- [x] **Comb Filter**: provide `ParallelData`-based approach for the 3 remaining Voice Loops
- [x] **Denormal Canceling** (DNC): since a precompiled flag in `C15Synth::doAudio()` should reliably suppress any occuring _denormal float_, many additions of 
`NlToolbox::Constants::DNC_const` should be redundant and can therefore be discarded
- [x] **ParallelData**: debugging assembler revealed that not all `ParallelData` operations are SIMD compliant, the goal here is to reduce them (if possible, get rid of them), potential flaws:
`namespace std { abs, min, max, clamp, round }`
- [x] **ParallelData** further flaws, outside namespace: `keepFractional, sinP3_wrap, sinP3_noWrap, threeRanges`
- [x] **Numeric Tests**: all potential improvements should produce values equal to the previous implementation, so for good measure, this should be verified

The mentioned improvements focus on primarily reducing rendering cpu consumption (although optimized Time Handling could affect cpu consumption during parameter events as well).

_**other, possibly way more complicated further improvements could be**_:
_(delegated to Milestone Beyond 1.7)_
- **Envelopes**: provide `ParallelData`-based approach (well, maybe that isn't even possible..)
- **Transition handling**: try to reduce long switch statements for transition events (as this is the backbone of the whole parameter handling in the Audio Engine, this seems to be a big deal of "surgery"..)
- **Time Handling**: time-related parameter smoothers (Envelope times, Macro times, Glide time) can be optimized by rendering transitions of incremental ("dx") signals instead of pure times (in ms), avoiding cpu-intensive time conversions during "clock ticks", reducing conversions only to event-related triggers
- **Cleanup**: once a dsp strategy for the EPC has been found, solidify proposed implementations, remove defines and testing
- **funsafe-math**: suspected to already be used within `ParallelData`, but may be additionally helpful for all scalar DSP (mostly in Mono Section)


projects/audio-engine/src/synth/c15-audio-engine/ae_potential_improvements.h
```diff
@@ -0,0 +1,34 @@
...
+
+#define POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_1 __POTENTIAL_IMPROVEMENT_DISABLED__
+#define POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_2 __POTENTIAL_IMPROVEMENT_DISABLED__
+#define POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_3 __POTENTIAL_IMPROVEMENT_DISABLED__
+// - three options, potentially reducing voice loops within the Comb Filter to zero
+
+#define POTENTIAL_IMPROVEMENT_DNC_CONST_OF_ZERO __POTENTIAL_IMPROVEMENT_DISABLED__
+#define POTENTIAL_IMPROVEMENT_DNC_OMIT_POLYPHONIC __POTENTIAL_IMPROVEMENT_DISABLED__
+#define POTENTIAL_IMPROVEMENT_DNC_OMIT_MONOPHONIC __POTENTIAL_IMPROVEMENT_DISABLED__
+// - first option just for proof of concept, two further options potentially saving many additions
```
MatthiasSeeber, line 34: So far, using improvement in lines 26, 27, 28, 32, 33 (no more Comb Voice Loops, no more Denormal cancelling) and observing the mean CPU consumption while using the Audio Engine by playing, loading Presets and such, there seems to be a performance gain by around 10%.
_(I intend to leave everything disabled within this PR, though. This corresponds to the Audio Engine using no improvements.)_

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
```diff
@@ -63,11 +71,29 @@ void Engine::PolyCombFilter::apply(PolySignals &_signals, const PolyValue &_samp
...
   m_out -= (m_apStateVar_4 * m_apCoeff_2);
+#if POTENTIAL_IMPROVEMENT_DNC_OMIT_POLYPHONIC
+#else
   m_out += NlToolbox::Constants::DNC_const;
+#endif
   m_apStateVar_2 = m_apStateVar_1;
   m_apStateVar_1 = tmpOut;
   m_apStateVar_4 = m_apStateVar_3;
   m_apStateVar_3 = m_out;
+#if POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_1
```
hhoegelo, line 82: 👏👏👏

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
```diff
@@ -126,6 +159,15 @@ void Engine::PolyCombFilter::apply(PolySignals &_signals, const PolyValue &_samp
   ind_tp1 &= m_buffer_sz_m1;
   ind_tp2 &= m_buffer_sz_m1;
   PolyValue sample_tm1, sample_t0, sample_tp1, sample_tp2;
+#if POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_3
+  // POTENTIAL_IMPROVEMENT_COMB_REDUCE_VOICE_LOOP_3: provided by parallel template
+  sample_tm1 = polyVectorIndex(m_buffer, ind_tm1);
+  sample_t0 = polyVectorIndex(m_buffer, ind_t0);
+  sample_tp1 = polyVectorIndex(m_buffer, ind_tp1);
+  sample_tp2 = polyVectorIndex(m_buffer, ind_tp2);
+  // seems to work as well, although I'm not sure if passing the whole buffer (even as reference) is a good idea..?
```
hhoegelo, line 168: References are just syntactic sugar for immutable pointers - so no reason to worry about this. I like the way it is implemented, but I don't believe the polyVectorIndex()-method can be SIMD'ed by the compiler.

hhoegelo:
So far - so great!

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
MatthiasSeeber: I fear that could be the case, too. Is there a way to verify?

hhoegelo: I can look at the assembler. Will do and report here.

hhoegelo: Ok, this loop is indeed not compiled into SIMD instructions:

Block 4:
movsxdl  (%r10,%rsi,4), %rdi
shl $0x6, %rdi
add %r13, %rdi
vmovssl  (%rdi,%rsi,4), %xmm0
vmovssl  %xmm0, (%rbx,%rsi,4)
inc %rsi
cmp $0xc, %rsi
jnz 0x687a0 <Block 4>

MatthiasSeeber: I see, thanks for the research.
So we technically don't really need `polyVectorIndex()`, but since you expressed approval of this implementation, and since this avoids "ugly" voice loops, we can still keep it, right?

hhoegelo:
I measured CPU usage with intel vTune. There are basically 2 hot spots: PolySoundGenerator and MonoReverb.

PolySoundGenerator (uses twice as much CPU as the second one (MonoReverb)) uses sinP3_noWrap, which loops over the vector and does scalar computations. I didn't measure or test, but the function could be implemented with SIMD like this, I think:

```
template <typename T, size_t size> inline ParallelData<T, size> sinP3_noWrap(ParallelData<T, size> _x)
{
  _x += _x;
  _x = std::abs(_x);
  _x = 0.5f - _x;

  auto x_square = _x * _x;
  return _x * ((2.26548f * x_square - 5.13274f) * x_square + 3.14159f);
}
```

MatthiasSeeber:
Alright, I will implement this, also by define.
Time Handling on the other hand, reveals itself to be a bit tricky - no breakthrough so far..

hhoegelo:
As for threeRanges, maybe something like this could be faster? 
(don't know if I understood and translated the algorithm correctly, though)

```
template <typename T, size_t size>
inline ParallelData<T, size> threeRanges(const ParallelData<T, size> &sample, const ParallelData<T, size> &ctrlSample,
                                         const float &foldAmnt)
{
  ParallelData<T, size> ret;

  auto tooSmallF = std::abs(static_cast<ParallelData<float, size>>(ctrlSample < -0.25f));
  auto tooBigF = std::abs(static_cast<ParallelData<float, size>>(ctrlSample > 0.25f));

  auto lower = (sample + 1.f) * foldAmnt - 1.f;
  auto upper = (sample - 1.f) * foldAmnt + 1.f;
  ret = tooSmallF * lower + (1.0f - tooSmallF) * sample;
  ret = tooBigF * lower + (1.0f - tooBigF) * sample;

  return ret;
}
```

MatthiasSeeber:
> As for threeRanges, maybe something like this could be faster?
> (don't know if I understood and translated the algorithm correctly, though)
> 
> ```
> template <typename T, size_t size>
> inline ParallelData<T, size> threeRanges(const ParallelData<T, size> &sample, const ParallelData<T, size> &ctrlSample,
>                                          const float &foldAmnt)
> {
>   ParallelData<T, size> ret;
> 
>   auto tooSmallF = std::abs(static_cast<ParallelData<float, size>>(ctrlSample < -0.25f));
>   auto tooBigF = std::abs(static_cast<ParallelData<float, size>>(ctrlSample > 0.25f));
> 
>   auto lower = (sample + 1.f) * foldAmnt - 1.f;
>   auto upper = (sample - 1.f) * foldAmnt + 1.f;
>   ret = tooSmallF * lower + (1.0f - tooSmallF) * sample;
>   ret = tooBigF * lower + (1.0f - tooBigF) * sample;
> 
>   return ret;
> }
> ```

Overwriting `ret` twice before returning something seems inadequate to me. I think this could be done like the `para_d` in the Comb Filter, using an abs() and sign based approach.

MatthiasSeeber:
Here a comparison. I klimpered the whole NL Mallets 12 bank top-to-bottom and screenshotted the Engine terminal messages afterwards.

1) without potential improvements

Test-Improvements-none

2) with all potential improvements

Test-Improvements-all

I think we're on the right track here.

MatthiasSeeber:
So far, I did not notice any changes in the sound, but I would like to be certain here.
Maybe, we also should include numerical tests of potential improvements?

MatthiasSeeber:
We decided that it would be acceptable for a while to leave the potential optimizations (which can be enabled via defines) in the code, in order to have complete freedom when deciding what to do regarding current audio glitches. This means, that Code Cleanup is also postponed, and this PR can be merged upon approval.

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -186,65 +192,162 @@ namespace std
...
     for(size_t i = 0; i < size; i++)
       ret[i] = std::max(in[i], a);
 
     return ret;
+#endif
   }
 
   template <typename T, size_t size> inline ParallelData<T, size> clamp(const ParallelData<T, size> &in, T min, T max)
   {
+#if POTENTIAL_IMPROVEMENT_PARALLEL_DATA_STD_MINMAX_FLOAT
```
hhoegelo, line 273: As wo already have min- and max-methods in parallel flavor, a parallelized clamp could be easily - and with improved readability - composed as:
```
return std::max(std::min(in, _max), _min);
```

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -186,65 +192,162 @@ namespace std
...
     return ret;
+#endif
   }
 
   template <typename TOut, typename T, size_t size>
   inline ParallelData<TOut, size> round(const ParallelData<T, size> &in)
   {
+#if POTENTIAL_IMPROVEMENT_PARALLEL_DATA_STD_ROUND
+    // assuming that the ROUNDING paradigm should fit nltoolbox.h[line 359]::float2int()
+    const ParallelData<T, size> wrap = 0.5f
```
hhoegelo, line 337: If this method relies on T to be float, please consider to add a static_assert.

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -257,38 +360,79 @@ inline ParallelData<T, size> unipolarCrossFade(const ParallelData<T, size> &_sam
 
 template <typename T, size_t size> inline ParallelData<T, size> keepFractional(const ParallelData<T, size> &in)
 {
+#if POTENTIAL_IMPROVEMENT_PARALLEL_DATA_KEEP_FRACTIONAL
+  const ParallelData<T, size> wrap = 0.5f
```
hhoegelo, line 364: If this method relies on T to be float, please consider to add an according "static_assert".

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
```diff
@@ -257,38 +360,79 @@ inline ParallelData<T, size> unipolarCrossFade(const ParallelData<T, size> &_sam
...
 
   return ret;
+#endif
 }
 
 template <typename T, size_t size> inline ParallelData<T, size> sinP3_wrap(ParallelData<T, size> _x)
 {
+#if POTENTIAL_IMPROVEMENT_PARALLEL_DATA_SINP3_WRAP
+  _x += -0.25f;
+  const ParallelData<T, size> wrap = 0.5f
```
hhoegelo, line 384: If this method relies on T to be float, please consider to add an according "static_assert".

hhoegelo: Maybe we should group all the algorithms relying on T == float into template function specilizations - I will think about it....

projects/audio-engine/src/synth/c15-audio-engine/ae_poly_comb.cpp
hhoegelo: Right!

projects/audio-engine/src/synth/c15-audio-engine/ParallelData.h
hhoegelo: Ok, even better: All methods that do only work for floats should have no T template parameter, for example:
```
template <size_t size> inline ParallelData<float, size> keepFractional(const ParallelData<float, size> &in) {}
```
Thus, the method won't even exist for ParallelData<OtherType, size>.

Other methods, like max and min, can be expressed in a general way for arbitrary types and specialized for floats:

```
// the general method working for any T, std::min() is defined for
 template <typename T, size_t size> inline ParallelData<T, size> min(const ParallelData<T, size> &in, T a)
  {
    ParallelData<T, size> ret;

    for(size_t i = 0; i < size; i++)
      ret[i] = std::min(in[i], a);

    return ret;
  }

// another overload for std::min, that is preferred over the more general (less specific) min()-implementation above
// see https://en.cppreference.com/w/cpp/language/overload_resolution
 template <size_t size> inline ParallelData<float, size> min(const ParallelData<float, size> &in, float a)
  {
    ParallelData<float, size> ret;
    _mm_intrinsics_voodoo_only_working_with_float32();
    return ret;
  }
```

I suggest using this pattern for min,max, round, clamp.

MatthiasSeeber: The suggested std::min() will be generalized, but not SIMD compliant..

MatthiasSeeber:
- I reduced available types of parallel operations to floats and simplified clamp where possible - as proposed. (clamp(parallel, scalar, parallel) however still requires at least half-intrinsic approach)
- The availability of the potential improvements remains bound to the define in ae_potential_improvements.h (and is set to disabled by default) --> ...
... I would still suggest to somehow integrate this as a flag into the build process, which enables this for "user" builds. This way, this would be instantly clear to the Developer, and no extra code manipulation would be necessary

PR #1771, opened 2020-05-04
Parameter-DB Ammendments

MatthiasSeeber:
This is a first attemtpt to further unify parameter definitions shared by different applications.
In addition to the so far generated headers (used by Playground and AudioEngine), I also provided definitions for Java (WebUI) and JavaScript (MC View).


projects/parameter-db/generated/Test_ProtoParameterFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+package	com.nonlinearlabs.client.dataModel.editBuffer;
+
+import java.util.Arrays;
+import java.util.NoSuchElementException;
+import java.util.HashMap;
+
+import com.nonlinearlabs.client.dataModel.editBuffer.EditBufferModel.SoundType;
+import com.nonlinearlabs.client.dataModel.editBuffer.EditBufferModel.VoiceGroup;
+import com.nonlinearlabs.client.world.overlay.html.soundSettings.SoundSettingsDialog;
```
MatthiasSeeber, line 18: As I am completely new to Java, I don't know whether "package management" would work here, especially with this test file having a completely different directory to the WebUI java files. I just put that here in order to provide a file similar to the original ParameterFactory.java.

projects/parameter-db/generated/Test_ProtoParameterFactory.java
```diff
@@ -0,0 +1,89 @@
...
+
+  static public HashMap<SoundType, int[]> disabledParametersBySoundType = new HashMap<SoundType, int[]>();
+  static {
+    final int[] singleDisabled = { 346 };
+    disabledParametersBySoundType.put(SoundType.Single, singleDisabled);
+    final int[] splitDisabled =  { 346 };
+    disabledParametersBySoundType.put(SoundType.Split, splitDisabled);
+    final int[] layerDisabled = {};
+    disabledParametersBySoundType.put(SoundType.Layer, layerDisabled);
+  }
```
MatthiasSeeber, line 87: I managed to provide all required sortings from the (updated) spreadsheet.

However, I am not convinced of this approach here. Like the parameter-db headers, I would much prefer to just provide definitions here, no actual code: the hidden and disabled ID arrays are fine, the HashMaps should be elsewhere (in a derived class in a separate file, for example).

Also, after a little research I found that enums are possible in Java. Maybe, usage of them would make the code more readable here.

projects/parameter-db/generated/test_MacroIDs.js
```diff
@@ -0,0 +1,19 @@
...
+const MacroIDs = {
+    get MC_A() { return 243; },
+    get MC_B() { return 244; },
+    get MC_C() { return 245; },
+    get MC_D() { return 246; },
+    get MC_E() { return 369; },
+    get MC_F() { return 371; }
+};
+
+console.log("usage: MacroIDs.MC_A = ", MacroIDs.MC_A);
```
MatthiasSeeber, line 19: The provided MC ID set is intended to replace any usage of Macro IDs within the "new.js" main file of the MC Screen. This file can simply be loaded as a separate script (before the main file) in the corresponding HTML file.

PR #1785, opened 2020-05-06
LPC firmware major overhaul

KlausStrohhaecker:
- now using common shared definitions in projects/shared for different projects like the LPC firmware and the helper tools, also for future use by playground and audio-engine.

Helper tools (running natively on the BBB):
- "lpc-read" tool for inspecting messages send by the LPC, with flexible viewing options.
- "lpc" tool for general control of the LPC, notably for the new debug/service/profiling functions.
- "ehc" tool for full-featured EHC ("pedals") setup.
- "ehc-preset" tool to apply (hard-coded) presets for pedals, basically a mockup/blueprint for the corresponding playground code.

LPC firmware:
- unified linker skripts for M0 and M4 cores to avoid duplicate data, with explicit shared memory region to avoid uncontrolled memory clobbering.
- M0/M4 : rewrite / speed optimization of the inter-process communication to use the shared memory region with all data packed into one single shared variable.
- M0/M4 : build artefacts now contain a human-readable signature denoting which firmware version it is, and which core, for easy version identification.
- M0 : complete rewrite / speed optimization of the main process loop, the keybed scanner now runs inside the timer interrupt while the ADC scanning happens running open-loop, controlled by a feedback mechanism to adjust cycle time to 781us (16x per 12.5ms). Timer interrupt now happens every ~1us as compared to 62.5us. ESPI clock-speed of ADC access increased from 16MHz to 20MHz.
- M0 : complete rewrite / speed optimization of the keybed scanner to afford a 16us cycle time (vs. 125us), increasing velocity timing information resolution to 16us. NOTE: pin definitions from boards/emphase_v5.* are not used anymore, rather direct and inlined GPIO access with hard-coded addresses is used for speed.
- M4 : main module task time-slicing is now correctly set up to avoid more than one task at at time (except for fast processes which have been combined in one single task). SPI heartbeat signal (toggling every 125us, going to BBB) moved to interrupt for less jitter. ADC processing stuff has been split into sections to distribute peak loads.
- M4 : added watchdog to take care of the occasional hang-up of any core (for ESD or other reasons), also to implement a software-controlled reset.
- M4 : LEDs processing now uses its own task to avoid task overruns and to implement multiple accesses to the same LED correctly.
- M4 : heartbeat LPC-->PG now runs independant from incoming audio-engine heartbeat.
- M4 : added fail-safe EEPROM store & recall for EHC setup and ribbon calibration data, using shadow regions and CRC checking.
- M4 : complete new code for a flexible and powerful implementation of External Hardware Controllers (EHCs, formerly known as pedals), with many new features, including robustness from user perspective ("it just works"). Uses sophisticated multi-level averaging/smoothing (and apparent resolution increase), FIR filtering, adaptive auto-ranging with dead-zones at the outer ends, and an adjustable auto-hold ("de-noising") for the output values.
- M4 : calculate all the floating point stuff for the curve data (velocity, bender, aftertouch) at boot time rather than at runtime on every edit buffer load.
- M4 : lots of helper/service/debug functions added (accessible via LPC messages), notably for profiling.
- M4 : lots of general cleanup/refactoring/commenting, removal of dead code / unused modules, etc.


projects/realtime-system/nl_lib/nl_lib_com/sys/eeprom_18xx_43xx.c
```diff
@@ -0,0 +1,115 @@
+/*
+ * @brief LPC18xx/43xx EEPROM driver
```
hhoegelo, line 2: is this comment still truth or just clutter?

projects/realtime-system/doc/V206-features.txt
```diff
@@ -106,3 +106,17 @@ By default, the function is disabled to avoid unneccessary traffic.
 It can be enabled/disabled by new BB message "BB_MSG_TYPE_SETTING, SETTING_ID_SEND_RAW_SENSOR_DATA"
 
 
+
+
+6) Complete rewrite for External Hardware Controllers (formerly knwon as Pedal).
```
hhoegelo, line 111: typo knwon

projects/realtime-system/nl_lib/nl_lib_com/sys/eeprom_18xx_43xx.c
KlausStrohhaecker: still correct legal info

projects/realtime-system/doc/V206-features.txt
KlausStrohhaecker: fixed

KlausStrohhaecker:
> 
> 
> Regarding documentation:
> I'd prefer to check in only a source document per topic and then let 'make' generate the translated documents. For example:
> check in: "ExternalHardwareControllers_UI_Implementation_V1.0.odt"
> let "make doc" generate a "ExternalHardwareControllers_UI_Implementation_V1.0.pdf" by use of "soffice --headless --convert-to pdf ....doc"
> 
> If you agree, I would add those make rules.

I'd strongly prefer to make the PDF's myself (one-click operation) and always upload source and pdf to the repo in sync, rather than relying on someone executing a make script at an unknown time in the future, plus having no control over it (including options for the generation, like image compression, bookmarks, etc).

Or is that 'make" triggered automatically on upload? If so, then it might be OK to automate it this way, of course... then again, I will always generate a PDF copy locally, for proof reading and getting a 1:1 output of the actual visual representation, so why not just use it?

hhoegelo:
> I'd strongly prefer to make the PDF's myself (one-click operation) and always upload source and pdf to the repo in sync

Ich stimme nicht mit Dir ueberein, dass es ein Argument FUER die Erzeugung von redundanten Dokumenten ist, dass sie so leicht zu erzeugen sind. Vielmehr halte ich es fuer ein potentielles Problem, dass nun Code, Quelle der Doku und Output der Doku in sync zu halten sind.

hhoegelo:
> Or is that 'make" triggered automatically on upload? 

Auf keinen Fall - das ist ja genau das Gegenteil von dem, was wir haben wollen: Wir erschaffen redundante Information und speichern sie. Welchen Sinn sollte das haben?

KlausStrohhaecker:
Der Sinn eines (an sich redundanten) PDFs ist mE der, das Dokument einfach und vor allem *barrierefrei* zur Verfügung zu stellen statt in einem den meisten Leuten unbekannten Spezialfomat, das ist ja gerade der Sinn von PDF (Portable Dokument Format). 
*.pdf, *.doc und *.xls sind lange etablierte "Weltstandards".

Dass der Ersteller dann geringfügig mehr Arbeit hat, von seinem (letzlich beliebigen) Quellformat in PDF zu konvertieren, ist gegenüber diesem Nutzen mE vollkommen irrelevant. Und man kann soviel Selbstdisziplin erwarten, dass das PDF aktuell gehalten wird wenn man die Quelle ändert und uploaded.

PR #1788, opened 2020-05-06
Lpc pedal ui rework

schlenkibus:


projects/playground/src/proxies/lpc/LPCProxy.cpp
```diff
@@ -325,6 +327,23 @@ void LPCProxy::sendSetting(uint16_t key, uint16_t value)
   DebugLevel::info("sending setting", key, "=", value);
 }
 
+void LPCProxy::sendPedalSetting(uint16_t pedal, uint16_t pedalType)
+{
+#warning continue here
```
schlenkibus, line 332: @KlausStrohhaecker Does this look like a valid staring point, to implement the new API, to you?

KlausStrohhaecker: If I could only parse C++ better ;-)
But basically, yes. Set up a message word by word, each word transferred via `*cmp << data;`. Repeat for all the required messages (3...7).
The message length (second word of raw message) is set up by `queueToLPC(cmp);` ?

projects/shared/lpc-converters.h
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <EHC-pedal-presets.h>
```
KlausStrohhaecker, line 3: Why do we need EHC-pedal-presets.h here?
And if so, aren't user includes set up with "" whereas system includes use <> ?

projects/playground/src/proxies/hwui/panel-unit/boled/setup/PedalEditor.cpp
```diff
@@ -3,28 +3,161 @@
...
+class PedalMenuEditorEntry : public MenuEditorEntry
+{
+ public:
+  PedalMenuEditorEntry(PedalEditor* parent, int i)
+      : m_parent(parent)
+      , m_index { i }
+  {
+  }
+
+  bool redraw(FrameBuffer& fb) override
```
hhoegelo, line 17: Can this override be removed then?

projects/playground/src/proxies/lpc/MessageParser.h
```diff
@@ -18,6 +18,7 @@ class MessageParser
     EDIT_CONTROL = 0x0500,
     MORPH_POS = 0x0600,
     SETTING = 0x0700,
+    EHC_CONFIG = 0x0F00,
```
hhoegelo, line 21: Please sort by value.

projects/shared/lpc-converters.h
schlenkibus: `EHC_ControllerConfig_T` must be known inside lpc-converters.h

projects/NonMaps/src/main/java/com/nonlinearlabs/client/presenters/DeviceSettings.java
```diff
@@ -19,7 +19,10 @@
 	static public class Pedal {
 		public String displayValue = "---";
 		public double sliderPosition = 0;
-		static public String[] options = { "Pot Tip Active", "Pot Ring Active", "Switch Closing", "Switch Opening" };
+
+		static public String[] options = { "Pot Tip Active", "Pot Ring Active", "Pot Tip Active Reverse", "Pot Ring Active Reverse", "Resistor", "Resitor Reversed", 
+										   "Switch Closing", "Switch Opening", "CV 0..5V", "CV 0..5V Auto Range", "OFF", "Boss EV 30", "Doepfer FP 5", "Fractal EV 2",
+										   "Korg DS1H", "Korg EXP2", "Lead Foot LFX1", "MAudio EXP", "Moog EP3", "Roland DP10", "Roland EV5", "Yamaha FC3A", "Yamaha FC7" };
```
KlausStrohhaecker, line 25: Please use "---OFF---" for the "off" setting which also doubles up as a delimiter between generic and specific settings.
Further, there are typos ("Resitor"), and spelling of several items is not correct, please use the strings from shared/EHC-pedal-presets.c literally (with all the commas, hyphens, etc), unless there are good reasons to change them (like not fitting in the display -- even after making the region for that as wide as possible which is not the case in the original UI layout).
For abbreviations like "Rev." it is of course possible to use the non-abbreviated word when space allows.
```
"Pot, Tip-Active"
"Pot, Ring-Active"
"Pot, Tip-Act. Rev."
"Pot, Ring-Act. Rev."
"Resistor"
"Resistor, Rev."
"Switch, Closing"
"Switch, Opening"
"CV, 0..5V"
"CV, 0..5V(Auto-Range)"
"---OFF---"
"E:Boss EV-30"
"E:Boss FV-500L"
"E:Doepfer FP5"
"E:Fractal EV-2"
"D:Korg DS-1H"
"E:Korg EXP-2"
"E:Lead Foot LFX-1"
"E:M-Audio EX-P (M.)"
"E:Moog EP-3 (std.)"
"D:Roland DP-10 (cont.)"
"E:Roland EV-5"
"D:Yamaha FC3A"
"E:Yamaha FC7"
```
The longest string is "D:Roland DP-10 (cont.)" with 22 chars, which is 7 more than legacy "Pot Ring Active" with 15. From eye-balling at the display of my C15, there is space for at least 25...30 chars give or take, accounting for the variable character size, so IHMO all string should fit.

projects/playground/src/proxies/lpc/LPCProxy.cpp
schlenkibus: Can you elaborate on: `required messages (3...7)`? What messages are we talking about. The resetting of the adc 0/1 for pedal 1 kind of messages?

projects/playground/src/device-settings/PedalType.h
```diff
@@ -1,14 +1,19 @@
...
-{
-  PotTipActive = 0,
-  PotRingActive = 1,
-  SwitchClosing = 2,
-  SwitchOpening = 3,
-};
+ENUM(PedalTypes, int, PotTipActive = EHC_PRESET_Pot_Tip_Active, PotRingActive = EHC_PRESET_Pot_Ring_Active,
+     PotTipActiveReverse = EHC_PRESET_Pot_Tip_Act_Rev, PotRingActiveReverse = EHC_PRESET_Pot_Ring_Act_Rev,
+     Resitor = EHC_PRESET_Resistor, ResistorReversed = EHC_PRESET_Resistor_Rev,
+     SwitchClosing = EHC_PRESET_Switch_Closing, SwitchOpening = EHC_PRESET_Switch_Opening, CV0to5V = EHC_PRESET_CV_0_5V,
```
KlausStrohhaecker, line 10: Typo "Resitor"

projects/playground/src/device-settings/PedalType.cpp
```diff
@@ -1,35 +1,73 @@
...
-{
-}
+PedalType::~PedalType() = default;
 
-void PedalType::sendToLPC() const
+void PedalType::sendToLPC(SendReason reason) const
 {
-  Application::get().getLPCProxy()->sendSetting(m_lpcKey, (uint16_t)(get()));
+  auto reset = reason == SendReason::SettingChanged;
+  Application::get().getLPCProxy()->sendPedalSetting(m_lpcKey, get(), reset);
```
KlausStrohhaecker, line 21: "Reason" looks wrong (and isn't working).

"Reset" must be set (and only be set) when the user presses Enter and especially even when the setting has NOT changed because this is the exact reason why "reset" was introduced in the first place, the user wanting to reset the pedal's auto-ranging limits (among other settings) which are kept across power-cycling (stored in LPC EERPOM).

"Reset" of course also must be set when the Setting was actually changed, which will be catched automatically when looking after the explicit "enter" button (HWUI and WebUI).

OTOH, "reset must" always be cleared when sendToLPC() was NOT invoked by user action, that is, when it is send as a part of the settings used to set-up the LPC inititially (smth. like "onEditBufferChange", currently). The LPC automatically resets the pedal if required, in that case.

projects/playground/src/proxies/lpc/LPCProxy.cpp
```diff
@@ -325,6 +327,25 @@ void LPCProxy::sendSetting(uint16_t key, uint16_t value)
...
 }
 
+void LPCProxy::sendPedalSetting(uint16_t pedal, PedalTypes pedalType, bool reset)
+{
+  auto len = EHC_GetLPCMessageLength();
+  uint8_t buffer[len];
+
+  if(auto written = EHC_ComposeLPCSetupMessageById(static_cast<EHC_PRESET_ID>(pedalType), pedal,
+                                                   reset ? EHC_NORESET : EHC_RESET, buffer))
+  {
```
KlausStrohhaecker, line 337: There is a global "off-by-one error", though not necessarily here.
The setting actually written to the LPC is the one from the *previous* position in the EHC_presets[] array, so the used ID seems to be wrong, being too low by 1.

KlausStrohhaecker: With settings.xml =
```
 <setting id="Pedal1Type">PotRingActive</setting>
 <setting id="Pedal2Type">PotTipActive</setting>
 <setting id="Pedal3Type">PotTipActive</setting>
 <setting id="Pedal4Type">PotTipActive</setting>
```
PG logs an error when re-transmitting the edit buffer after a heartbeat reset:
```
May 23 19:34:18 apl playground[7615]: Warning:  Could not compose pedal preset 2 = 0
May 23 19:34:18 apl playground[7615]: Warning:  Could not compose pedal preset 3 = 0
May 23 19:34:18 apl playground[7615]: Warning:  Could not compose pedal preset 4 = 0
```
"PotTipActive" is wrongly translated to ID 0 (instead of ID 1) which then gets de-asserted by EHC_ComposeLPCSetupMessageById()

PR #1792, opened 2020-05-07
Lpc henry

hhoegelo:
With this PR merged, make bbb-rootfs (and all the update targets) will produce a bbb image containing  all the bbb/lpc supplement tools installed into /usr/bin (resources (bbmsg) are installed into /usr/share.


KlausStrohhaecker:
So I'm gonna try compile this under WSL and check if all output executables still work as intended?

Some things don't make too much sense, though, like projects/bbb-tools/inhouse-ribbon-calibration/ribbon-readout-helper-tool/readout-ribbon-adcs.sh is a script intended to be run on a linux PC connection to a C15 via SSH.
This script, as well as the associated "ribbon-filter" tool will be soon defunct anyway as the ribbon cal thing (with its custom-tailored output) will move to the "lpc-read" tool, in use together with the "lpc" tool, and this again will be replaced by a intelligent true in-system calibration tool later on (either in PG or outside).

For the development of all those BBB-tools for now I will keep sticking to my local Windoze cross-compiler etc as this is so much faster and more way convenient than the "official" build process via WSL (7 minutes minimum turnaround time, you know :-).

projects/bbb-tools/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(bbb-tools)
+
+add_subdirectory(ehc)
+add_subdirectory(ehc-preset)
+add_subdirectory(inhouse-ribbon-calibration)
+add_subdirectory(lpc)
```
KlausStrohhaecker, line 8: Build errors, no CmakeLists.txt in "make-cal-bin" and "ribbon-filter".
And compiling/collecting stuff from ribbon-readout-helper-tool doen't make much sense for the target system.
==>temporarily removed "add_subdirectory(inhouse-ribbon-calibration)"

PR #1828, opened 2020-05-14
implement feature 1826

hhoegelo:


projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/maps/parameters/Parameter.java
```diff
@@ -192,8 +189,14 @@ public Control pinch(Position eventPoint, double touchDist, TouchPinch pinch) {
 		return super.pinch(eventPoint, touchDist, pinch);
 	}
 
+	public boolean forceFine(boolean in) {
```
schlenkibus, line 192: what does this method do?

PR #1845, opened 2020-05-18
changed load to part to be independent of focus and mode && reset load to part on dual preset load, closes #1833 and closes #1837

### LINKED ISSUE #1833: BOLED: Preset-Screen im "To Part"-Modus, Parameter selektieren: "To Part" sollte aktiv bleiben. Vorbild: WebUI
- [ ] "To Part" darf auch nicht ausgehen, wenn man auf den Preset-Button drückt. Das sollte den Parameter-Screen öffnen (oder was sonst vorher offen war).

### LINKED ISSUE #1837: BOLED: Dual-Preset laden, "To Part" einschalten: statt des eigenen Parts wird ein Preset/Part von früher selektiert. Vorbild: WebUI
Das Laden eines Dual-Presets sollte alle alten Selektionen für "To Part" zurücksetzen.

In der WebUI funtioniert es korrekt: Beim Einschalten von "To Part wird zuerst der selektierte Part des geladenen Presets vorgeschlagen (Quelle = Ziel).

schlenkibus:
closes #1833
closes #1837


hhoegelo:
I am struggling here: It feels as if "LoadToPart" is something that shouldn't be owned buy the HWUI, but I am not sure. Is the mode connected to the WebUI? Should it be a setting? If I remember correctly, it once was a setting.
Maybe we can skype about it and you can explain to me, what exactly the "load to part" is and why it is so different from other states in the system.

projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -1591,6 +1591,11 @@ void EditBuffer::cleanupParameterSelection(UNDO::Transaction *transaction, Sound
 
     hwui->setCurrentVoiceGroup(VoiceGroup::I);
   }
+
+  if(newType == SoundType::Single)
```
schlenkibus, line 1595: subscribe to type changed from hwui

projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -39,6 +39,7 @@
 #include <parameters/ScopedLock.h>
 #include <tools/StringTools.h>
 #include <parameter_declarations.h>
+#include <proxies/hwui/panel-unit/boled/preset-screens/PresetManagerLayout.h>
```
hhoegelo, line 42: unused include

projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -873,6 +877,7 @@ void EditBuffer::undoableSetType(UNDO::Transaction *transaction, SoundType type)
     transaction->addSimpleCommand([=](auto state) {
       swap->swapWith(m_type);
       m_signalTypeChanged.send();
```
hhoegelo, line 879: can we get rid of the duplicate signal then and just ignore the sound type on receivers side?

schlenkibus: sure

PR #1849, opened 2020-05-18
reset load to part selections on dual preset load, closes #1837

### LINKED ISSUE #1837: BOLED: Dual-Preset laden, "To Part" einschalten: statt des eigenen Parts wird ein Preset/Part von früher selektiert. Vorbild: WebUI

schlenkibus:
closes #1837

### DIFF
--- a/projects/playground/src/presets/EditBuffer.cpp
+++ b/projects/playground/src/presets/EditBuffer.cpp
@@ -39,6 +39,7 @@
 #include <parameters/ScopedLock.h>
 #include <tools/StringTools.h>
 #include <parameter_declarations.h>
+#include <proxies/hwui/panel-unit/boled/preset-screens/PresetManagerLayout.h>
 
 EditBuffer::EditBuffer(PresetManager *parent)
     : ParameterDualGroupSet(parent)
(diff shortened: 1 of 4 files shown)


projects/playground/src/presets/EditBuffer.cpp
```diff
@@ -1591,6 +1592,15 @@ void EditBuffer::cleanupParameterSelection(UNDO::Transaction *transaction, Sound
 
     hwui->setCurrentVoiceGroup(VoiceGroup::I);
   }
+
+  if(auto layout = Application::get().getHWUI()->getPanelUnit().getEditPanel().getBoled().getBaseLayout())
```
hhoegelo, line 1596: I am not happy with this control direction. If avoidable, we should not control the UI directly from edit buffer. I don't have better solution at hand right now (as I don't understand the problem you try to solve) - maybe we should have a skype call and try to come up with a better decoupled solution.

PR #1855, opened 2020-05-19
Supervisor power cycling feature

KlausStrohhaecker:


projects/supervisor/src/led.cpp
```diff
@@ -238,4 +124,111 @@ void Led_t::Show_Hardware_and_Firmware_ID(void)  // handle special case
   fw_blinks = FW5_Version_ID + 1;  // set blink counter according to firmware revision
 }
 
+static int8_t showPowerCyclingOverride(void)
+{
+  static const int8_t *pSeq = SEQ_PWR_CYCLING;
+  int8_t               c    = *pSeq++;
+  if (!c)
+    pSeq = SEQ_PWR_CYCLING;
```
hhoegelo, line 132: Shouldn't c be reassigned to *pSeq at this place?

projects/supervisor/src/led.cpp
```diff
@@ -238,4 +124,111 @@ void Led_t::Show_Hardware_and_Firmware_ID(void)  // handle special case
   fw_blinks = FW5_Version_ID + 1;  // set blink counter according to firmware revision
 }
 
+static int8_t showPowerCyclingOverride(void)
+{
+  static const int8_t *pSeq = SEQ_PWR_CYCLING;
+  int8_t               c    = *pSeq++;
+  if (!c)
```
hhoegelo, line 131: I guess, you either want to test the advanced *pSeq here instead of 'c' to find the terminator, or you should re-assign c to the wrapped sequence value, no?

projects/supervisor/src/led.cpp
```diff
@@ -238,4 +124,111 @@ void Led_t::Show_Hardware_and_Firmware_ID(void)  // handle special case
...
+  int8_t c = getNextCmd();
+  displayCmd(c);
+  applyWait(c);
+}
+
+static int8_t getNextCmd(void)
+{
+  int8_t c = *pSeq++;
+  if (!c)
+    pSeq = pSeq_start;
```
hhoegelo, line 200: pointer is reset, but 'terminator' is returned - is this intended?

projects/supervisor/src/led.cpp
KlausStrohhaecker: The /0 terminator is a legal value and is returned to the caller. It is (currently) discarded later in displayCmd() and applyWait().
The outer functions may want to know whether a sequence has reached its and then, for example may want to switch to another/previous sequence (for a one-shot type of pattern) which would be impossible when the inner function doesn't report the end of the sequence.

projects/supervisor/src/led.cpp
KlausStrohhaecker: .... see above

projects/supervisor/src/led.cpp
hhoegelo: ok, thanks for the explanation

Issue #1542, opened 2020-03-12
Simple automatic backup system


hhoegelo:
@KlausStrohhaecker used to install a BeagleBone shell script for automated backups of presets on shut down. The backups could have been used by technicians to restore user content after fatal system damage. 
As playground moved to ePC it becomes more complicated to backup user generated content on bbb, especially when systems are shutting down independently. Also, the backups are not available to the user without opening the housing.

@nonlinear-labs agreed on the following idea:
The playground creates a backup copy of the preset manager on every startup. It keeps up to 100 backups. The backups can be downloaded via well known URLs:

http://192.168.8.2/backup/0

will download the most recent backup, created just before the current playground instance started.

http://192.168.8.2/backup/99

will download the most outdated available backup.

There is no UI to list or manage backups, neither HWUI nor WebUI. Restoring of backups is done via HWUI or WebUI with the existing feature set.

KlausStrohhaecker:
Looks ok to me, preferably if it's a non-blocking parallel task, not delaying startup.
The physical medium is the ePC's removable SSD, so there are chances to read that SSD even if the ePC gave up the ghost, right?
Do we also store other user settings and device information in those backups that would be needed to fully "clone" a C15 onto another hardware instance quickly with a seamless transition?

KlausStrohhaecker:
What is the physical path to where the backups are stored?
In the emergency case, the SSD will have to installed into a new ePC which should then be booted from an USB stick with a live system, mounting the SSD read-only to avoid any further corruption.

PR #1866, opened 2020-05-22
Update preconditions

AntonSchmied:
I have added the discussed precondition checks for our update procedure. However, I was not yet able to run a test or two on a C15, resulting this to be a draft atm.

If you have further remarks or questions. let's hear them!

Quick overview:

- [x] chmod +x rootfs tools, apperently these are not always executable
- [x] Pre-Update Checks: Can we access the ePC from BBB, abort if Windows is the OS
- [x] Forced manual reboot after the update procedure (LPC needs a hard reset)
- [x] Friendlier OLED messages, El Jefe says it s enough to print out the steps without further description
- [x] Abort BBB Update if settings/presets/calibration data is present but conditions are bad (Critical when updating from 1.5 to 1.7)
- [x] Make fixCPUBinding and fixOverlayOrder part of the update
- [x] check if windows desktop is mountable (preconditions)
- [x] ~~check if LPC is updatable @KlausStrohhaecker, wie denn?~~

Should we create a backup directory on the BBB, which is untouched by rsync?


KlausStrohhaecker:
> Abort BBB Update if settings/presets/calibration data is present but conditions are bad (Critical when updating from 1.5 to 1.7)

I'm unclear what you mean by this...

KlausStrohhaecker:
> Should we create a backup directory on the BBB, which is untouched by rsync?

A backup of non-recoverable data in case of failure always is a good idea. So this would include preset, settings and calibration.
Presets and settings are very important also for a seamless and complete migration of a working C15 to new/other underlying hardware (in case the old hardware breaks). The removable medium (SD-card) is the medium of choice because it can easlily be salvaged from a broken C15 and chances are high it still contains valid data.

For that reasons, MS1.7 should also make regular backups of the critical user data to the SD-card of the BBB (as we currently don't populate the SD-card slot on the newer ePCs), even though the SSD drive of the ePC is removable itself (but might die, and SSDs die in an instant once the cells have worn out, whereas the SD-card, when used only for backups might have a much longer service life).

AntonSchmied:
> > Abort BBB Update if settings/presets/calibration data is present but conditions are bad (Critical when updating from 1.5 to 1.7)
> 
> I'm unclear what you mean by this...

On version 1.5 settings/presets/calibration data is kept on the BBB
So in order to push these to the ePC (update to 1.7) the following conditions have to be fulfilled, otherwise the update is aborted:
- is the ePC IP address correct
- can the ePC be pinged
- can we logon onto the ePC
- is /persistent mounted

KlausStrohhaecker:
Understood. In that case a backup of the user data and calib sure is sort of mandatory anyway. Maybe we could even copy a copy of the backup on the USB-stick when the update starts from there....

KlausStrohhaecker:
> check if LPC is updatable @KlausStrohhaecker, wie denn?

A misunderstanding, probably. LPC can always be updated unless we have a gross hardware failure (GPIOs on BBB not working -- with one BBB revision exactly having that problem, as we know, or programming connenction via USB<-->serial chip broken). I haven't seen a failure of this kind on a formerly working Uniboard. For additional safety we could create a firmware download setup for the MXLI which tries to download the current firmware(s) from the chip and look if that succeeds but IHMO this is overkill. MXLI programming is very robust and reliable IME, much more robust than programming the chip with the progging adapter. Just that sometimes, for yet unknown reasons, the chip doesn't come out of reset even when pulling the reset line, requiring a power-cycle to recover.

Or did you mean Supervisor instead? If so, I'm just writing a Uniboard revision readout tool from which we can see if it is a hardware < 7.1 which cannot update the Supervisor.

KlausStrohhaecker:
And a suggestion for 7.1 Hardware:
We could hold the Supervisor in reset during the whole update, notably when Supervisor is also updated (as the first action). By this, accidental or intentional user action to shut down the C15 during update via power button could be inhibited. HW 7.1 is designed so that system power remains up even when the Sup is in reset (and/or re-programmed) but requires some specific care by the BBB.
A topic of only medium priority IMHO, not required for the initial release of the new update mechanism

AntonSchmied:
> Or did you mean Supervisor instead? If so, I'm just writing a Uniboard revision readout tool from which we can see if it is a hardware < 7.1 which cannot update the Supervisor.

no i meant LPC. I mean we have not so far but could build in a check if we can reach the LPC, so we can proceed with an update. But if that is unneeded then we can leave it out

AntonSchmied:
> A topic of only medium priority IMHO, not required for the initial release of the new update mechanism

I like the idea, but what do we do with the older C15s? Maybe create a ticket for it with target "beyond 1.7"?

AntonSchmied:
the checklist is done. @hhoegelo and @schlenkibus what are your thoughts on backups for presets/settings/calibration data? Especially when moving from 1.5 (BBB) to 1.7 (ePC).

AntonSchmied:
So, @schlenkibus and I discussed the backup topic
- The Idea to do recurring backups is already discussed here #1542
- If we talk about the Update 1.5 to 1.7 then the BBB will keep everything till the update is successful. So in this case the critical data is stored safely and is left untouched, if the conditions are bad, as i pointed out earlier

> * is the ePC IP address correct
> * can the ePC be pinged
> * can we logon onto the ePC
> * is /persistent mounted

build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -187,22 +228,12 @@ main() {
...
-        pretty "" "Rebooting System..." "" "Rebooting System..." ""
-        sleep 2
+        pretty "" "$MSG_UPDATING_C15 $MSG_FAILED" "$MSG_CHECK_LOG" "$MSG_UPDATING_C15 $MSG_FAILED" "$MSG_CHECK_LOG"
+        freeze
     fi
 
-    [ $UPDATE_EPC == 1 -a $UPDATE_BBB == 1 ] && rebootEPC
-    [ $UPDATE_BBB == 1 ] && rebootBBB
+    pretty "" "$MSG_UPDATING_C15 $MSG_DONE" "$MSG_RESTART" "$MSG_UPDATING_C15 $MSG_DONE" "$MSG_RESTART"
+    freeze
```
hhoegelo, line 236: I don't understand, why we have to freeze the device here. All my other devices (PC, radio, mobile phone) are just rebooting after successful update.

AntonSchmied: @KlausStrohhaecker ... ?

AntonSchmied: El Jefe sagt, dass ihm das ganz gut gefällt, wenn das Gerät von alleine startet nach einem erfolgreichen Update.

KlausStrohhaecker: Ich bleibe dagegen, stelle mich aber nicht in den Weg und bin dann halt auch nicht schuld.

AntonSchmied: @KlausStrohhaecker, und wie war ist es mit dem harten LPC reset?

KlausStrohhaecker: Es kommt (selten) vor, dass der LPC ein power-cycling braucht nach dem progging, ein Reset allein reicht nicht. Einer der Gründe für ein power-cycling (das halt leider nur der User intiieren kann, derzeit). Zumindest solange wir keine Diagnose und entspr. User-Meldungen machen zum "allgemeinen Systemzustand".
Gerade nach einem Update stiftet ein dann nicht mehr funktionierendes Gerät richtig Verwirrung auch wenn es nur der temporäre LPC-glitch wäre, was der User aber nicht wissen kann.

AntonSchmied: El Jefe and I discussed this just now and we decided to keep the 'freeze'-state after a successful update for the upcoming release, forcing the user to restart the C15. (Windows Updates for example, instruct the user to restart the computer to complete the update, is one of the examples we thought of).

We can think about changing the approach, once we can control the Supervisor to actually provoke a Power-Cycle, which I believe, we are not able to do atm.

And if the restarting gets annoying for us developers, then maybe we can set a flag to not Power-Cycle, just a thought.

KlausStrohhaecker: With the newer switch-equipped C15's (HW V71), manual power-cycling is extremely simple, just switch off and immediately on again, no need to wait until the shutdown has finished. Only with the older button-equipped models you have to wait until shutdown has completed.

And correct, atm we cannot instruct the supervisor to initiate a power cycle, current supervisor code (5.1 and 5.2) does not implement it, and there is neither user code nor driver on BBB side that serve the serial interface.

Regardless of simple soft reboots or full power-cycle, we can always easily implement a feature scanning the TopCover Buttons to inhibit power-cycling or reboot, or to force it or select between them, whatever. Could be hidden (no screen message) or public (reboot? y/n), the future is wide open....

hhoegelo:
So can we finally merge the PR?

AntonSchmied:
I have nothing left to add to this PR

PR #1869, opened 2020-05-25
Protect journal

AntonSchmied:
Prevents the update from clearing journal on the BBB

### DIFF
--- a/projects/scripts/bbb/install-update.sh
+++ b/projects/scripts/bbb/install-update.sh
@@ -1,15 +1,15 @@
 #!/bin/sh
 
 if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]; then
-    rm -rf /update \
-    && mkdir -p /update \
-    && cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
-    && cd /update \
-    && tar xf nonlinear-c15-update.tar
-
-    if [ $? -eq 0 ]; then
-        mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
-        chmod +x /update/run.sh
-        /bin/sh /update/run.sh
+    if rm -rf /update \
+        && mkdir -p /update \
+        && cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
+        && cd /update \
+        && tar xf nonlinear-c15-update.tar; then
+            mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
+            if ! ls -l /update/BBB/ | grep bbb_update_1811a.sh && ! ls -l /update/BBB/ | grep bbb_update_1905a.sh; then
+                chmod +x /update/run.sh
+                /bin/sh /update/run.sh
+            fi
     fi
 fi
--- a/projects/scripts/bbb/install-update.sh
+++ b/projects/scripts/bbb/install-update.sh
@@ -1,10 +1,9 @@
 #!/bin/sh
 
-rm -rf /update
-mkdir -p /update
-
 if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]; then
-    cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
+    rm -rf /update \
+    && mkdir -p /update \
+    && cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
     && cd /update \
     && tar xf nonlinear-c15-update.tar
 
@@ -14,4 +13,3 @@ if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]; then
         /bin/sh /update/run.sh
     fi
 fi
-
--- a/projects/scripts/bbb/install-update.sh
+++ b/projects/scripts/bbb/install-update.sh
@@ -1,13 +1,17 @@
 #!/bin/sh
 
-if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]
-then
-    rm -rf /update/*
-    cp /mnt/usb-stick/nonlinear-c15-update.tar /update
-    cd /update
-    tar xf nonlinear-c15-update.tar
-    mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
-    chmod +x /update/run.sh
-    /bin/sh /update/run.sh
+rm -rf /update
+mkdir -p /update
+
+if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]; then
+    cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
+    && cd /update \
+    && tar xf nonlinear-c15-update.tar
+
+    if [ $? -eq 0 ]; then
+        mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
+        chmod +x /update/run.sh
+        /bin/sh /update/run.sh
+    fi
 fi
 
--- a/projects/scripts/bbb/install-update.sh
+++ b/projects/scripts/bbb/install-update.sh
@@ -7,7 +7,9 @@ if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]; then
         && cd /update \
         && tar xf nonlinear-c15-update.tar; then
             mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
-            if ! ls -l /update/BBB/ | grep bbb_update_1811a.sh && ! ls -l /update/BBB/ | grep bbb_update_1905a.sh; then
+            if ! ls -l /update/BBB/ | grep bbb_update_1811a.sh \
+                && ! ls -l /update/BBB/ | grep bbb_update_1905a.sh \
+                && ! ls -l /update/BBB/ | grep playground-RELEASE-2019-05-1.5-706569a; then
                 chmod +x /update/run.sh
                 /bin/sh /update/run.sh
             fi
--- a/projects/scripts/bbb/install-update.sh
+++ b/projects/scripts/bbb/install-update.sh
@@ -1,15 +1,13 @@
 #!/bin/sh
 
-rm -rf /update
-mkdir -p /update
-
 if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]
 then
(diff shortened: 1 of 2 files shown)


AntonSchmied:
ADDITIONALY: These are our journalctl settings. Should we alter anything while we are at it?

```
# cat /etc/systemd/journald.conf
#  This file is part of systemd.
#
#  systemd is free software; you can redistribute it and/or modify it
#  under the terms of the GNU Lesser General Public License as published by
#  the Free Software Foundation; either version 2.1 of the License, or
#  (at your option) any later version.
#
# See journald.conf(5) for details

[Journal]
Storage=persistent
#Compress=yes
#Seal=yes
#SplitMode=uid
#SyncIntervalSec=5m
#RateLimitInterval=30s
#RateLimitBurst=1000
#SystemMaxUse=
#SystemKeepFree=
#SystemMaxFileSize=
#RuntimeMaxUse=
#RuntimeKeepFree=
#RuntimeMaxFileSize=
#MaxRetentionSec=
#MaxFileSec=1month
#ForwardToSyslog=no
#ForwardToKMsg=no
#ForwardToConsole=no
#ForwardToWall=yes
#TTYPath=/dev/console
#MaxLevelStore=debug
#MaxLevelSyslog=debug
#MaxLevelKMsg=notice
#MaxLevelConsole=info
#MaxLevelWall=emerg
```

projects/scripts/bbb/install-update.sh
```diff
@@ -1,15 +1,13 @@
...
 if [ -e /mnt/usb-stick/nonlinear-c15-update.tar ]
 then
-	cp /mnt/usb-stick/nonlinear-c15-update.tar /update		
-	cd /update
-	tar xf nonlinear-c15-update.tar
-	mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
-	chmod +x /update/run.sh
-	/bin/sh /update/run.sh
+    rm -rf /update/*
+    cp /mnt/usb-stick/nonlinear-c15-update.tar /update
```
KlausStrohhaecker, line 6: Wouldn't it be safer to emit a `mkdir -p /update` like before, just in case /update doesn't exist?

And in rare cases, the SD-card might be corrupted and then might be mounted read-only, then the update would have no chance to execute (and the cp statement should be checked for a fail, then). One cound try copy to RAM instead (/tmp or any other place with a tempfs file-system), and try reformat and remount the SD-card before coying the the new root-fs.

AntonSchmied: my bad, I thought /update is part of the rootfs build! will fix it and add the checks!

KlausStrohhaecker:
According to https://www.freedesktop.org/software/systemd/man/journald.conf.html
by default the files might take up to 10% total disk space, that would be 1.6G in case of 16G SD-Cards. Also, the indivdual files grow as large as1/8th of that size (200M), for seven rotated files in the history.

Using the orignal 16G SD-Card from my test-C15 (was in heavy use but isn't old exactly, as we introduced the 16G card not that long ago, and there is a MS1.5 system running) I get a used space of 406M, currently.
On the second SD-card, running an MS1.7 installed just recently, used space for journal is 56M.

I suggest we check the currently used spaces on all C15's we can access, with `du -h /var/log/journal` to get a feeling for the used sizes. Of course only the candidates which have not been updated with the current mechanism deleting the journals... probably not much units left, I fear. We could also use the power-cycling to force a lot of journal data in short time.

Appending data to a compressed large file might be a slow process increasing boot times as the system ages, so maybe we should check how much space is needed to cover, say, at least, the last 20..30 boots and normal use cases (unit on for several hours, the occasional warnings from PG/AE etc), with compression maybe set to off and then restrict total and individual file sizes accordingly, something like
> #SystemMaxUse=500k
#SystemKeepFree=
#SystemMaxFileSize=100k
#RuntimeMaxUse=500k
#RuntimeKeepFree=
#RuntimeMaxFileSize=100k

I have no gut feeling right now what would be reasonable sizes, though.

PS. We have to check the journal quotas on the ePC as well

projects/scripts/bbb/install-update.sh
AntonSchmied: done! 
however i left out the reformating/remounting of the SD-Card out for now. It seems to me like a little bit of a bigger chapter. But if we all agree that we need this, maybe make a ticket out of it?

KlausStrohhaecker: OK, but the rm -rf ec still after the 'if', so /update gets not deleted until another valid update is started. Which is quite convenient for testing (simple to re-run it) and also as a dignostic so we can see what whas the last update and how it went.

For the same reason we should not delete stuff in /upadte at other places, like the `rm -rf /update/BBB/rootfs/*` etc. in bbb_update.sh?

projects/scripts/bbb/install-update.sh
```diff
@@ -1,15 +1,13 @@
...
+    cd /update
+    tar xf nonlinear-c15-update.tar
+    mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
+    chmod +x /update/run.sh
+    /bin/sh /update/run.sh
 fi
```
KlausStrohhaecker, line 16: There is no safeguard to prevent the user from installing very old updates. We had introduced the check for an existing run_v2.sh (post 1811) and run_v3.sh (post MS1.5) for that reason, do you remember?

projects/scripts/bbb/install-update.sh
AntonSchmied: /update/BBB/rootfs is a temp. directory. the rootfs will still be there as a '.tar.gz'

projects/scripts/bbb/install-update.sh
AntonSchmied: @hhoegelo and I discussed this and came to a conclusion, that adding further pointer scripts could get confusing. We thought on gathering the checks within one main script, the run.sh

KlausStrohhaecker: https://github.com/nonlinear-labs-dev/C15/blob/master/build-tools/create-c15-update/create-c15-update.sh
installs those file in the update, so that current generation 'install-update.sh' won't fail, but the newly deployed 'install-update.sh' doesn't check for them. Actually the counter should be increased to run_v4.sh to prevent users from installing pre-MS.17 updates once the first update to MS1.7 took place

KlausStrohhaecker: >@hhoegelo and I discussed this and came to a conclusion, that adding further pointer scripts could get confusing. We thought on gathering the checks within one main script, the run.sh

No, you can't prevent the exection of an old run.sh belonging to an old update, so this won't work inside (a new run.h). The only way to prevent it is either the check for the dummy files or a direct analysis of the contents of the update after unpacking, as a part of "install-update.sh"

AntonSchmied: yes, I see what you mean. need a moment to think about this though.

KlausStrohhaecker: Probably something like version control file instead of those dummy scripts would be nice.
With a major and a minor version number, and a new update can only be installed when the major number is the same or greater in the update than currently installed.
Whereas 'recent' updates with lower minor numbers than current could still be installed.

Most cetainly @hhoegelo will have a nice solution for this ;-)

At any rate the check must be inside the deployed new "install-update.sh"

AntonSchmied:
A quick and dirty update flow chart, maybe it will help see the unseen

https://docs.google.com/document/d/1anIS3YssSXhYHn20x_CRuWBNn4Yiks9alsepbdo6tJI/edit

projects/scripts/bbb/install-update.sh
AntonSchmied: @schlenkibus and I looked into what might happen, when different updates are deployed on different running versions, and thanks to your checks in the previous run.sh's @KlausStrohhaecker it seems like we're safe. I tried to gather everything here:
https://drive.google.com/open?id=1ZM3OnGrG_riIvAecdooKNiawM4KndQtD

Once we're on 1.7, we think that updating back and forth between the 1.7 versions should be allowed or rather does not seem to be critical. If the User is unhappy with an update, he could always go back f.e.

KlausStrohhaecker: Looks like not all scenarios are analyzed correctly (as checked on a running MS1.5)
- MS1.5 --> 1811 and older : won't execute as MS1.5 checks for run_v3.sh and1811 only provides run_v2.sh, older only run.sh.
- MS1.5 --> 1905 (which is a minor wrt 1811): won't exceute, for the same reason
Since we had set up most 1.5 systems manually, it's not clear if all have the proper install-update.sh which does that check, though. Then, the update would fail at the secodary checks

But those are errors on the goog side, so we should have double protection ;-)

Further notes:
On MS1.7, updates which would happen to execute and are relying on text2soled being located at the lecacy position will crash from the missing text2soled. Applies to (most) 1811 and older. 
1905 installs a new text2soled in the legacy position and won't crash.
Newer1811 update packages have been updated to install their own text2soled after the issues popped up (text2soled clearing the display on every call, a bug we re-introduced lately ;-) but I don't know if they ever have been shipped.

Still I feel it would be better to discard older updates in MS1.7 systematically rather than relying on them to fail internally and generate irritating messages (or even crash from the missing text2soled).

AntonSchmied:
so ... the install_update.sh will check whether the unpacked update has old files under /update/BBB
- bbb_update_1905a.sh
- bbb_update_1811a.sh
- playground-RELEASE-2019-05-1.5-706569a
It will protect the 1.7 from windows and 1.5 (jic) related updates.

projects/scripts/bbb/install-update.sh
```diff
@@ -1,15 +1,15 @@
...
-        chmod +x /update/run.sh
-        /bin/sh /update/run.sh
+    if rm -rf /update \
+        && mkdir -p /update \
+        && cp /mnt/usb-stick/nonlinear-c15-update.tar /update \
+        && cd /update \
+        && tar xf nonlinear-c15-update.tar; then
+            mv /mnt/usb-stick/nonlinear-c15-update.tar /mnt/usb-stick/nonlinear-c15-update.tar-copied
+            if ! ls -l /update/BBB/ | grep bbb_update_1811a.sh && ! ls -l /update/BBB/ | grep bbb_update_1905a.sh; then
+                chmod +x /update/run.sh
```
KlausStrohhaecker, line 13: check wouldn't catch very old updates prior to 1811 (I archived updates going back to 2017-08) but that's probably worth the risk?

AntonSchmied: jesus ... there are more??? :D I focused on the updates i know of, so the ones we offered or handed out explicitly. but it's your call @KlausStrohhaecker

PR #1871, opened 2020-05-25
start bbbb after install update

schlenkibus:
fixes bug where epc starts loading banks and displays them via bbbb even though the install-update-from-usb.service yet has to run -> bbbb should start after install-update completed

@hhoegelo or was there a particular reason to remove that dependency in the first place?

### DIFF
--- a/projects/bbbb/systemd/bbbb.service.in
+++ b/projects/bbbb/systemd/bbbb.service.in
@@ -1,6 +1,6 @@
 [Unit]
 Description=Nonlinear-Labs BeagleBoneBlackBridge
-After=syslog.target network.target systemd-modules-load.service
+After=syslog.target network.target systemd-modules-load.service install-update-from-usb.service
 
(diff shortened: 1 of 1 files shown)


KlausStrohhaecker:
Yes, that dependency has always been there since 1811a (both the then playground and bridge services had to wait for 'install-update.sh' to complete, I too wonder why it has been removed?

And when 'install-update-sh' actually finds and excecutes an update, it (or the called run.sh) must not return. Rather it must display success/failure message(s), prompt the user and wait in an infinite loop until the user power-cycles the C15. This is the only safe way IMHO and common practice, I've never seen anythings else, no fancy auto-reboot of parts of the system -- also, the update of the LPC sometimes requires a power-cycle for the LPC to get it out of reset.

KlausStrohhaecker:
Ah sorry, the reboot stuff already addressed in #1866

PR #1881, opened 2020-05-26
added boled and soled download via webui

### LINKED ISSUE #1875: WebUI Developer Options: Export Boled/Soled
add developer option to download soled / boled screens via browser

schlenkibus:
closes #1875 

TODO: change bbbb and exported color pallete to match real colors on HW displays (white instead of yellow)

### DIFF
--- a/projects/playground/src/http/ServedFile.cpp
+++ b/projects/playground/src/http/ServedFile.cpp
@@ -18,10 +18,10 @@ ServedFile::~ServedFile()
 void ServedFile::startServing()
 {
   Glib::ustring filePath;
-  if(m_request->getPath().find("/tmp/") != 0)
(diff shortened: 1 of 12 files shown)


projects/playground/src/http/ServedFile.cpp
```diff
@@ -17,7 +17,12 @@ ServedFile::~ServedFile()
 
 void ServedFile::startServing()
 {
-  Glib::ustring filePath = getFilePathFromMessagePath();
+  Glib::ustring filePath;
+  if(m_request->getPath().find("/tmp/") != 0)
```
hhoegelo, line 21: better test for '== 0', so that /foo/bar/tmp/baz will not match - and of course invert the logic behind

PR #1910, opened 2020-06-03
issue 1909

MatthiasSeeber:


MatthiasSeeber:
It is decided that we instead tackle the problem from the UI-side.

Issue #1538, opened 2020-03-11
text2soled - drawing rectangles


AntonSchmied:
If I try to draw a rectangle with "horizontal-line" and "vertical-line" on the sOLED, the previously drawn line disappears. 
- Using `/usr/C15/text2soled/text2soled` 
- Master [41f3f37] --> too old?

PLUS:
Can we please bring the "TEST OLEDs" Feature back to the Developer Options! Was really handy for production! --> new ticket?

KlausStrohhaecker:
I double-checked this and indeed the text2soled version seems to be from the stone age where both screens are always cleared by default before plotting.

We first ran into this issue when updating to "system 1811a" which introduced a periodic monitor script "check_system.sh" to check if ePC can be ping'd and if on-board flash is mounted (via displaying dots on SOLED when these tests fail), allowing some remote diagnosis. On older systems (early V6 era) with the auto-clearing text2soled, this causes issues of blanking the screen when either error condition appeared. For the same reason, running the update scripts caused garbled display messages from this behavior.

This was fixed in the "PG1905a" update (and hotfixed for the 1811a deploy) by also copying a new text2soled executable to the system.

schlenkibus:
@AntonSchmied please check on the next clean deploy if you encounter an old version of text2soled again, if this is the case this ticket is still relevant. If a deployment process leads to the correct version of text2soled this ticket can be closed.

AntonSchmied:
So, pulled the latest master, deployed the Update on Stephan's C15,
the problem remains, that I cannot draw a rectangle using `/usr/C15/text2soled/text2soled`.
Each new line deletes the previously drawn.

Used the following lines of code to draw the rectangle:

```
# /usr/C15/text2soled/text2soled clear
# /usr/C15/text2soled/text2soled horizontal-line 0 0 256
# /usr/C15/text2soled/text2soled horizontal-line 0 63 256
# /usr/C15/text2soled/text2soled horizontal-line 0 64 256
# /usr/C15/text2soled/text2soled horizontal-line 0 95 256
# /usr/C15/text2soled/text2soled vertical-line 0 0 96
# /usr/C15/text2soled/text2soled vertical-line 255 0 96
# /usr/C15/text2soled/text2soled vertical-line 127 64 32

```

PR #1924, opened 2020-06-04
fixed text2soled: uncontrolled clearing of the display

KlausStrohhaecker:
- move clear() call from multitext constructor to execute()
- remove default clear() before trying to parse command line when no named command was found
==>
Display is cleared only with "multitext" command and explicit "clear" command.
Now compatible with legacy behaviour


KlausStrohhaecker:
closes issue #1538

PR #1928, opened 2020-06-05
add mxli to /update/utilities/ and closes #1736

### LINKED ISSUE #1736: LPC update
lpc_update.sh relies on /nonlinear/scripst/mxli. @KlausStrohhaecker suggests it should be part of the update.tar stashed under /LPC/.

I can change the path, in the lpc_update.sh script. However, I would not know how to deploy the mxli with cmake into the right directory.

AntonSchmied:


build-tools/create-c15-update/create-c15-update.sh
```diff
@@ -133,6 +133,12 @@ get_tools_from_rootfs() {
         fi
     done
 
+    # once mxli is part of /usr/bin this can be done in the loop above
+    if ! cp $(find $BINARY_DIR/build-tools/bbb/rootfs/ -type f -name "mxli") $OUT_DIRECTORY/utilities/ & ! chmod +x $OUT_DIRECTORY/utilities/mxli; then
```
hhoegelo, line 137: I am sure you've tested it, but the 

```
if ! A & !b; then
fi
```

doesn't look right to me. First, I would have expected && instead of &. Even if I just don't get it right and everything works fine here, I would read the & as 'cp in background' so I'd suggest to rewrite the operations in a more readable way, for example by splitting both operation into separate blocks.

AntonSchmied: done!

PR #1933, opened 2020-06-05
nonllinear-c15-update.tar is only renamed if update was successful. 

### LINKED ISSUE #1515: Renaming nonlinear-c15-update.tar before reboot
Renaming 'nonlinear-c15-update.tar' should be a sign of a successful update and should happen before rebooting the BBB and ePC

Solution: passing the name of the update to run.sh and renaming it there:
`/bin/sh /update/run.sh /mnt/usb-stick/nonlinear-c15-update.tar`

AntonSchmied:
also closes #1515

have to run a quick check on Monday to see if it all works.

### DIFF
--- a/build-tools/create-c15-update/update_scripts/run.sh
+++ b/build-tools/create-c15-update/update_scripts/run.sh
@@ -1,9 +1,4 @@
 #!/bin/sh
-#
-#
-# Name:         Anton Schmied
-# Date:         2020.05.22
-# Version:      2.1
 
 EPC_IP=192.168.10.10
 BBB_IP=192.168.10.11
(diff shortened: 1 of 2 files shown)


schlenkibus:
@AntonSchmied looks good to me, if the test on monday succeeds, feel free to merge this PR

AntonSchmied:
So, I've tested it again and I can't find any flaws.

- When updating from 1.5 the update with the old install-update.sh is carried out. The OLEDs will tell the user if the update was successful or not and ask him/her to restart the C15. The install-update.sh is then replaced by the new version.
- Once the C15 is running a 1.7 Release it might find an update on the stick, but will not execute anything old (1.5/ 1811/ 1905). The C15 will start up the BBBB and run as if nothing happened!
- If it does find a 1.7 update the update will be carried out. The OLEDs will tell the user if the update was successful or not and ask him/her to restart the C15.
- I also added a timed check in `check_preconditions()` which waits for the playground on the ePC for 10 seconds. A couple of times the check came way to soon and nothing was done.

PR #1936, opened 2020-06-08
Kstr pedal ui

schlenkibus:


KlausStrohhaecker:
This PR overwrote the same PR I had just opened, killing my content....

### Checklist:
### Pedals:
- [ ] WebUI : 
grafik
- [ ] PG startt with wrong/unknown Pedal-Type in settings.xml : must call the Pedal-Init with "reset" in this case

tbc

projects/realtime-system/doc/Compiler_Hinweise.txt
```diff
@@ -0,0 +1,33 @@
+Compiliert nur mit LPCXpresso Version 8.2.2_650 !!
```
hhoegelo, line 1: Is this file still needed / its content valid? Looks like clutter to me.

projects/realtime-system/doc/V206-features.txt
```diff
@@ -1,10 +1,9 @@
-LPC firmware Version 205 (for Milestone 1.7) features:
```
hhoegelo, line 1: Do you actually use this text file? To me, it doesn't look like a file that should be versioned, but more like personal notes.

projects/realtime-system/doc/Compiler_Hinweise.txt
KlausStrohhaecker: Yes, valid and up-to-date and very required, it actually is THE guide to succesfully compile with LPCxpresso.

projects/realtime-system/doc/V206-features.txt
KlausStrohhaecker: It is a summary of the current features, part of the usual project documentation. I could move its contents into "realtime-system\src\shared\sys\nl_version.h" where it belongs to, conceptually, ... but I fear nobody will find quickly when it is buried that deep...

KlausStrohhaecker:
> 
> 
> I cannot comment on doc / odt files - there is at least one 'draft' file. Does it mean, that it can be removed now?

Removed: no (contains information I don't want to loose), moved:yes (to private archive). Also for ExternalHardwareControllers_UI_Implementation_V1.0.odt
Done.

KlausStrohhaecker:
.... pinging @hhoegelo on Slack...

projects/playground/src/proxies/hwui/panel-unit/boled/setup/SetupLayout.cpp
```diff
@@ -540,10 +554,10 @@ namespace NavTree
       children.emplace_back(new WebUIAdress(this));
       children.emplace_back(new FreeInternalMemory(this));
       children.emplace_back(new UISoftwareVersion(this));
-      children.emplace_back(new RTSoftwareVersion(this));
-      children.emplace_back(new OSVersion(this));
       children.emplace_back(new DateTime(this));
       children.emplace_back(new UpdateAvailable(this));
+      children.emplace_back(new RibbonCalibration(this, true));
```
hhoegelo, line 559: Consider using an enum for upper/lower as true/false can not be understood by the reader without diving into the code

PR #1958, opened 2020-06-10
Issue 1953 : missing LPC reset after power-up

### LINKED ISSUE #1953: LPC dead on start
After updating a C15 with HW 6 with master - 381eb19 the LPC does not start as usual. One has to reset it manually. Similar behavior has been observed on a C15 with HW 7.1 when updating to KSTR-PedalUI. One had to restart the C15 twice after the update. Otherwise there was no sound.
 
I am "standing on the hose" atm. ... the only difference I see, is the reduction of LPC blobs to one.

KlausStrohhaecker:
fixes #1953
also fixes #1975 and some minor glitches (see files).

General:
- fix new LPC code to initialize clocks & PLLs reliably
- unrelated but required anyway: reset LPC after power-up on older motherboards where the supervisor can't reset the LPC. Required because of slightly wrong hardware wrt reset and supply ramping.
- this might fully settle the occasional LPC hang-up issues and hence it should be safe now, even when decide not to force a power-cycle after an update
- other features : added "key counters" (scanner level and TCD level) as diagnostic tool to track down missed note on/off events

Topics:
- [x] fix LPC code
- [x] extend the "lpc" tool for a "managed reset" function : reset with with several retries until LPC is definitely up. 
- [x] extend lpc tools and code for monitoring keybed events
- [x] while we're at it, integrate the four helper LPC tools in the build chain
- [x] add a one-shot boot-time BBB service for "managed reset"
- [x] add check via "managed reset" also for LPC update process
- [ ] testing (I can confirm success on both V6 and V7.1 hardware platforms, so far)

### IMPORTANT: !! this PR must be merged in unison with PR https://github.com/nonlinear-labs-dev/nonlinux/pull/12 !!


AntonSchmied:
I am afraid, the reset does not help on V6.
I can run the script manually `./reset-lpc.sh` and receive a positive return, but the LEDs remain dark and the C15 silent ...

KlausStrohhaecker:
Looks like a simple reset is not always enough, sometimes doesn't work at all, on some C15's for , so this is still WIP -- converted to draft

AntonSchmied:
I can confirm that a single reset does not seem to solve the problem.
I have furthermore updated the LPC with 2 old blobs (M0/M4) and after that the LPC starts well, without any noticeable dropouts.

build-tools/bbb/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 SET(USER $ENV{USER})
 SET(DOCKERNAME nl-cross-build-environment)
+SET(NONLINUX nonlinear_2020.04--lpc-tools)
```
KlausStrohhaecker, line 4: # IMPORTANT: !! change back to `nonlinear_2020.04` before merging !!

AntonSchmied:
I have built an update and deployed it on a C15 (V7.1).
Now everytime I start the C15, right after the banks are loaded, the C15 is frozen for a short while - Pushing buttons or touching the ribbons does not have any effect. I started the C15 5 times, and each time I observed a similar behavior.

KlausStrohhaecker:
Hhm, that sounds strange. I've done a number of updates on my C15 and no problem at all.
Maybe we can double check with a build from the current master whether this is actually related to the changes introduced here or not....
Note: I found that for a reliable build it is advisable to start with a new, clean build directory which sadly takes several hours....

hhoegelo:
Probably related to fsync calls I've just added... I'll take care of it tomorrow.

AntonSchmied:
@KlausStrohhaecker, Henry fixed the issue, with the last PR. Could you merge the master into this PR? Then i'd build it anew and test it, cheers!

AntonSchmied:
Why do you change the version.h? I can see the branch and the commit Head in the Sub-Menu.
Would this mean, you would have to change it before merging into master?

Plus I have not noticed an Audio-Engine dropout after a C15-Update-Restart

KlausStrohhaecker:
> Why do you change the version.h? I can see the branch and the commit Head in the Sub-Menu.
> Would this mean, you would have to change it before merging into master?

I just want to keep track of the changes in an easy way. Notably when we plan to distribute beta builds among colleages for testing it is nice to have a quick human-readable and catchy reference and keep it up to date (the commit ID isn't that human readable).

Master shall never contain a legal version ID as it is a (confirmed) work-in-progress state by definition.
That is, the version shall be the place-holder "YY-WW" or even "not defined" or something.

The consequence is that anything else like true temporary test builds (of master) should be branched off from master in case of a release build, or whatever branch for a beta, and have the versio ID set before build and deploy.
This one responsibility left for the person doing the build that cannot be "automated away".

That is at least my view what the handling of the version ID should be

> Plus I have not noticed an Audio-Engine dropout after a C15-Update-Restart
Good to hear

KlausStrohhaecker:
Checklist:
- [x] https://github.com/nonlinear-labs-dev/C15/issues/1859 (".not-mounted")
- [ ] https://github.com/nonlinear-labs-dev/C15/issues/1978 ("Unison - Voices verschieden in WebUI und BOLED"
- [x] https://github.com/nonlinear-labs-dev/C15/issues/1974 ("Update procedure: breaking detection of ePC OS type")
- [x] https://github.com/nonlinear-labs-dev/C15/issues/1980 ("Ribbon Touch Threshold (left side of ribbon) wrong with default calibration")
- [x] https://github.com/nonlinear-labs-dev/C15/issues/1953 ("LPC dead on start")
- [x] https://github.com/nonlinear-labs-dev/C15/issues/1897 ("HWUI: Nach dem Hochfahren wird die MC-Position nicht in den Ribbon-LEDs angezeigt")

projects/lpc-tools/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+add_executable(ehc        src/ehc.c)
+add_executable(ehc-preset src/ehc-preset.c  ${CMAKE_SOURCE_DIR}/projects/shared/EHC-pedal-presets.c)
+
+
+include_directories(src)
+include_directories(${CMAKE_SOURCE_DIR}/projects/)
+
+
+# INSTALL
+install(TARGETS lpc         DESTINATION ${C15_INSTALL_PATH}/${PROJECT_NAME})
```
hhoegelo, line 23: if you want to have this binaries installed system wide, you could replace ${C15_INSTALL_PATH} with /usr/bin.

KlausStrohhaecker: OK, will use litaral path

PR #1964, opened 2020-06-11
Update message and stop service

AntonSchmied:
Informs the User, that the Update will start (a bit friendlier) and stops the services without a BBB_Update dependency.


KlausStrohhaecker:
Not directly related, but for some reason the bbbb.service doesn't seem to be blocked by install-update.service.
I always see the PG starting before the update starts, which means bbbb is already running.

Looks like the following fixes this:
- Make install-update.service type Oneshot (which it should have been anyway)
- Add dependancy "Before=bbbb.service" to it

KlausStrohhaecker:
checked proper function when update is present and also when it is not present.
approved.

PR #1975, opened 2020-06-12
Update procedure: breaking detection of ePC OS type?

### LINKED ISSUE #1974: Update procedure: breaking detection of ePC OS type?
Issue detected:

- Check_preconditions() calls wait4playground() which seems to be intended to detect whether we have a NLL linux system on the ePC.
- This will only succeed when MS1.7 is already installed on the ePC.
- A MS1.5 ePC has no playground.service and thus this check will fail.
- Therefore, Check_preconditions() will try to reach the ePC assuming it is Windows system, which will fail also.

```
wait4playground() {
    for COUNTER in $(seq 1 $TIMEOUT); do
        echo "awaiting playground ... $COUNTER/$TIMEOUT"
        sleep 1
        executeAsRoot "systemctl status playground" && return 0
    done
    return 1
}

check_preconditions() {
    if ! wait4playground; then
        if [ -z "$EPC_IP" ]; then report "" "E81: Usage: $EPC_IP <IP-of-ePC> wrong ..." "Please retry update!" && return 1; fi
        if ! ping -c1 $EPC_IP 1>&2 > /dev/null; then  report "" "E82: Cannot ping ePC on $EPC_IP ..." "Please retry update!" && return 1; fi
        if executeOnWin "mountvol p: /s & p: & DIR P:\nonlinear"; then
            report "" "E84: Upgrade OS first!" "Please contact NLL!" && return 1
        else
            if mount.cifs //192.168.10.10/update /mnt/windows -o user=TEST,password=TEST \
                && ls -l /mnt/windows/ | grep Phase22Renderer.ens; then
                    report "" "E85: OS too old for update!" "Please contact NLL!" && return 1
            fi
        fi
        report "" "Something went wrong!" "Please retry update!" && return 1
    fi
    return 0
}
```

KlausStrohhaecker:
check for NLL linux by testing for known directories rather that checking if there exists a playground.service (which doesn't on a MS1.5 system)

fixes #1974


build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -73,17 +73,18 @@ executeOnWin() {
 
 TIMEOUT=10
 
-wait4playground() {
+isNllLinuxSystem() {
```
AntonSchmied, line 76: While I agree, that waiting for Playground blocks 1.5 (And Windows!!) and was unintentional, this function is only supposed to wait for the ePC to show up on the Ethernet horizon. Therefor a simple ping should do. The following checks should be enough to tell the user if something is not right.

build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -92,9 +93,10 @@ check_preconditions() {
             if mount.cifs //192.168.10.10/update /mnt/windows -o user=TEST,password=TEST \
                 && ls -l /mnt/windows/ | grep Phase22Renderer.ens; then
                     report "" "E85: OS too old for update!" "Please contact NLL!" && return 1
+            else
+              report "" "E86: Can't detect OS!" "Please contact NLL!" && return 1
```
AntonSchmied, line 97: why should the OS not be detectable?

KlausStrohhaecker:
I suggest a re-write for two-step check:
1 ping, if it fails throw message & exit (unlikely and if so, a typically recoverable error)
2 detect OS type, as explicitly as needed:
- check if we can issue any simple linux type command, then we have some linux (MS1.5++)
- else check the two possible windows variants and handle a missing mount.cifs gracefully. The point is, after a BBB udpate to MS1.5++ there is no samba + mount.cifs anymore and the check cannot be uses (would fail in the unlikely case that a old windows ePC is attached after one did a MS1.5++ update on the BBB).

KlausStrohhaecker:
Ping would fail with the common eth0 hang-up, so not that unlikely

build-tools/create-c15-update/update_scripts/run.sh
KlausStrohhaecker: I would assume there are possible error scenarios where the ePC can be pinged but we don't manage to detect the OS type (see below), some type of OS screw-ups, what have you. Would ba hard fault (unit needs true hand-on repair, probably)

AntonSchmied:
1) `executeAsRoot "exit" && return 0` this way we check if we can logon to MS 1.5++ 
2)  Check Connection and the two windows variants.

If all the checks fail, I decided to tell the User to retry the update. However this really should not happen and can be avoided with the checks. In case we for some reason cannot detect the OS even after a (several) retry(-ies) ... I assume the User will tell us about it.

So I would do it like so
```
wait4epc() {
    for COUNTER in $(seq 1 $TIMEOUT); do
        echo "awaiting playground ... $COUNTER/$TIMEOUT"
        sleep 1
        executeAsRoot "exit" && return 0
    done
    return 1
}

check_preconditions() {
    if ! wait4epc; then
        if [ -z "$EPC_IP" ]; then report "" "E81: Usage: $EPC_IP <IP-of-ePC> wrong ..." "Please retry update!" && return 1; fi
        if ! ping -c1 $EPC_IP 1>&2 > /dev/null; then  report "" "E82: Cannot ping ePC on $EPC_IP ..." "Please retry update!" && return 1; fi
        if executeOnWin "mountvol p: /s & p: & DIR P:\nonlinear"; then
            report "" "E84: Upgrade OS first!" "Please contact NLL!" && return 1
        else
            if mount.cifs //192.168.10.10/update /mnt/windows -o user=TEST,password=TEST \
                && ls -l /mnt/windows/ | grep Phase22Renderer.ens; then
                    report "" "E85: OS too old for update!" "Please contact NLL!" && return 1
            fi
        fi
        report "" "Something went wrong!" "Please retry update!" && return 1
    fi
    return 0
}
```

KlausStrohhaecker:
looks good, except the `echo "awaiting playground ... $COUNTER/$TIMEOUT"`
The ping detection is still upside down logic but if it works, don't bother ;-)

AntonSchmied:
yes true ... we're "waiting for the OS" to be correct.
However I would disagree that the detection is upside down. If we start Ping-ing first, the OS might still not yet be ready to be overwritten. executeAsRoot 'exit' does both things for us. The ping test afterwards would only verify that the Ethernet connection is down, if it is.

KlausStrohhaecker:
No worries ;-)

KlausStrohhaecker:
will be fixed via #1958

PR #1977, opened 2020-06-17
system call 'sync' hangs up to 10 seconds when called immediately

hhoegelo:
after booting the bbb.

Now, bbb's sync is called only if there is actually a chance that
the write operations target is the usb stick.

Also, syncing is done in background thread without blocking the main
thread anymore.


hhoegelo:
@reviewers: It is probably enough if one of you could review the PR.

PR #1982, opened 2020-06-17
For an unknown reason, audio-engine.path systemd component

hhoegelo:
does not work reliable on KSTR's system right after update.
To not run any risk, we decided for pumbing our own "wait for audio hardware"
mechanism, based on spinning loop.


KlausStrohhaecker:
Looks fine, checked on my system, both standard power-ups (right after update as well as lateron) and LPC-resets.

KlausStrohhaecker:
Looks fine, checked on my system, both standard power-ups (right after update as well as lateron) and LPC-resets.

PR #1985, opened 2020-06-17
Fixing this in a clean way is nearly impossible: buildroot just does

hhoegelo:
not support cleaning the target dir. Thus, old artifacts that
we want to be sure to wipe away, have to be removed in the build rule.
Otherwise, switching between branches while re-using the pre-built buildroot
in fs.ext4 will create zombies in the final builds. (#1859)

### DIFF
--- a/build-tools/bbb/CMakeLists.txt
+++ b/build-tools/bbb/CMakeLists.txt
@@ -106,6 +106,7 @@ add_custom_command(
         "make lpc_bb_driver-rebuild"
         "make espi_driver-clean-for-reconfigure"
         "make espi_driver-rebuild"
+        "rm -rf /fs/nonlinux/output/target/mnt/.not-mounted"
         "make"
         "cp /fs/nonlinux/output/images/rootfs.tar.gz /workdir/"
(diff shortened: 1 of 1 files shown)


KlausStrohhaecker:
So the implications is (please correct in case it's wrong):
Whenever building "something serious", start from scratch (delete the build-tools\bbb\fs.ext4 (or is there a better way, remove the .filesystem-created marker?)

hhoegelo:
> So the implications is (please correct in case it's wrong):
> Whenever building "something serious", start from scratch (delete the build-tools\bbb\fs.ext4 (or is there a better way, remove the .filesystem-created marker?)

I tried all combinations of targets and stamp removals without success.
Maybe I should add a "clean" target, that removes the build dir and target dir inside fs.ext4, but keeps the ccache and dowloads! ... Later.

KlausStrohhaecker:
> Maybe I should add a "clean" target, that removes the build dir and target dir inside fs.ext4, but keeps the ccache and dowloads! ... Later.

Sounds like an efficient way to handle the zombie problems

PR #1994, opened 2020-06-18
OTA updates should reboot the device afterwards

hhoegelo:


build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -203,6 +203,7 @@ main() {
     rm -f /mnt/usb-stick/nonlinear-c15-update.log.txt
     rm -f /update/errors.log
     touch /update/errors.log
+    chmod +x /update/utilities/*
```
KlausStrohhaecker, line 206: Had to insert `chmod` for update to work when copied to target with Windows tools

PR #1995, opened 2020-06-19
added sda4 fix if broken

AntonSchmied:
Apparently /dev/sda4 is not always properly installed on the ePC. (I have not yet found out why). Coming from 1.5 to 1.7 this can be crucial, since /persistent is not always mounted and we cannot transfer data from the BBB.

Worked manually, have to check in context of an update on Monday.

### DIFF
--- a/build-tools/create-c15-update/update_scripts/epc_fix.sh
+++ b/build-tools/create-c15-update/update_scripts/epc_fix.sh
@@ -42,6 +42,7 @@ epc_fix() {
     fi
 
     if ! fsck /dev/sda4 -n; then
+        umount /dev/sda4
         if ! mkfs.ext4 /dev/sda4; then
             printf "E48 ePC update: fixing sda4 failed" >> /tmp/fix_error.log && ((fix_errors++))
         fi
@@ -55,6 +56,7 @@ epc_fix() {
 
 main() {
     epc_fix
+    return $?
 }
 
 main
--- a/build-tools/create-c15-update/update_scripts/epc_fix.sh
+++ b/build-tools/create-c15-update/update_scripts/epc_fix.sh
@@ -41,6 +41,12 @@ epc_fix() {
         fi
     fi
 
+    if ! fsck /dev/sda4 -n; then
+        if ! mkfs.ext4 /dev/sda4; then
+            printf "E48 ePC update: fixing sda4 failed" >> /tmp/fix_error.log && ((fix_errors++))
+        fi
+    fi
+
     if [ $fix_errors -gt 0 ]; then
          return 48
     fi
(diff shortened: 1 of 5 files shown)


KlausStrohhaecker:
How can we create a forced case to check?

AntonSchmied:
Well ... What I do is, install a "naked" NonLinux on the ePC from a USB Stick (see https://github.com/nonlinear-labs-dev/Audiophile2NonLinux). This installation is sometimes clean and sometimes /dev/sda4 is un-mountable:
```
$ fsck /dev/sda4 -n

fsck from util-linux 2.34
e2fsck 1.45.3 (14-Jul-2019)
ext2fs_open2: Bad magic number in super-block
fsck.ext2: Superblock invalid, trying backup blocks...
Superblock has an invalid journal (inode 8).
Clear? no

Journal superblock is corrupt.
Fix? no

fsck.ext2: The journal superblock is corrupt while checking journal for /dev/sda4
e2fsck: Cannot proceed with file system check

/dev/sda4: ********** WARNING: Filesystem still has errors **********
``` 
One could try to deliberately corrupt the magic number of the super-block ... or something similar. I have not yet found a way to that.

build-tools/create-c15-update/update_scripts/epc_fix.sh
```diff
@@ -41,6 +41,12 @@ epc_fix() {
         fi
     fi
 
+    if ! fsck /dev/sda4 -n; then
```
KlausStrohhaecker, line 44: Could we just fake an error here to see if all goes well, script-wise etc?

AntonSchmied: I have not come up with anything deliberate yet ... But i am looking into how to damage a partition atm.

build-tools/create-c15-update/update_scripts/epc_fix.sh
```diff
@@ -41,6 +41,12 @@ epc_fix() {
...
+        if ! mkfs.ext4 /dev/sda4; then
```
KlausStrohhaecker, line 45: /dev/sda4 must be unmounted first, otherwise mkfs.ext4 will bail out

KlausStrohhaecker:
```
executeAsRoot() {
    echo "sscl" | /update/utilities/sshpass -p 'sscl' ssh -o ConnectionAttempts=1 -o ConnectTimeout=1 -o StrictHostKeyChecking=no sscl@$EPC_IP "sudo -S /bin/bash -c '$1' 1>&2 > /dev/null"
    return $?
}
```
The suffix  `1>&2 > /dev/null` seems to spoil the redirection of the output as used in `epc_fix()`, line 119, erroneously forcing empty output.
` executeAsRoot "cat /tmp/fix_error.log" >> /update/errors.log && return $result`

Same potential issue in `executeOnWin()`

AntonSchmied:
I used `dd if=/dev/zero of=/dev/sda4 bs=1k count=1024` on the ePC to deliberately "damage" he partition. 
To try out fail report comment following lines in "epc_fix.sh": 44,45,49 (/dev/sd4 should be mounted on /persistent)

PR #1999, opened 2020-06-19
2020.06 LPC fix midi buffer overflow ("Beta 20-25 RC03")

### LINKED ISSUE #1991: hanging notes
@nonlinear-labs reported several occasions of hanging notes during extensive play.

I've added diagnostic features in the LPC code and helper tools (https://github.com/nonlinear-labs-dev/C15/tree/issue-1953) to track down the error if it happens at or below TCD message level. 

When it happens, do the following:
- log on to the C15 with a terminal (putty)
- stop bridge: `systemctl stop bbbb`
- start the lpc-read tool `lpc-read -h`
- log on to the C15 with another terminal (putty)
- don't have any keys pressed on the keybed from now on!
- fetch LPC diagnostic data with `lpc req status`, results are displayed by lpc-read in the other window. It should look like this:
```
0000000000 SYSTEM STATUS:
cleared with 'lpc req clear-status':
  M4 ticker      :    3129724 ticks since last call
  Scheduler      :     0 task overruns
  Scheduler      :     2 tasks max. per time-slice
  Scheduler      :    60us max. task runtime
  Scheduler      :    60us max. time-slice
  Midi Buffers   :     0 dropped
  BBB Msg        :     0 buffer overruns / ESPI send fails
  TCD Msg        :     0 buffer overruns / USB send fails
cleared after every call:
  M0 ADC Scanner :   815us max. round trip time
  M0 Key Scanner :     0 (overrun flag)
cleared only by reset:
  Keybed:        :     0 missed events (Scanner)
  Keybed:        :     0 missed events (TCD)
```
The last two lines indicate whether there is a mismatch between total count of note-on vs note-off events, for two indepedant stages of processing (low-level keybed scanner and TDC message level).
Further, you should have `TCD Msg        :     0 buffer overruns / USB send fails`, if not, TCD messages may have not been succesfully send over MIDI.

To analyze deeper when missed events have been indicated, run `lpc req key-counters` which will output the counters for each key, again both on the scanner level (hardware key numbers 0...60) and TCD level (MIDI key numbers 0..127):
```
0000000000 Key Counters
 Scanner:
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .
 TCD:
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .     .     .     .     .
    .     .     .     .     .     .     .     .
```
Every position that doesn't show the dot (==0) marks a key with missed events. Positive numbers tell that there have been more note-ons than note-offs, negative means more note-offs than note-ons.

KlausStrohhaecker:
closes #1991

### DIFF
--- a/projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
+++ b/projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
@@ -14,13 +14,15 @@
 #include "tcd/nl_tcd_adc_work.h"
 
 #include "shared/lpc-defs.h"
+#include "shared/globals.h"
 
 /******************************************************************************/
 /*	module local defines														  */
 /******************************************************************************/
 
-#define BUFFER_SIZE       (64)
-#define NUMBER_OF_BUFFERS (8)
+#define BUFFER_SIZE            (256)  // space for 256 / 4 = 64 TCD events per buffer
+#define NUMBER_OF_BUFFERS      (8)    // number of buffers, must be 2^N !
+#define NUMBER_OF_BUFFERS_MASK (NUMBER_OF_BUFFERS - 1)
 
 /******************************************************************************/
 /*	modul local variables													  */
@@ -67,10 +69,29 @@ static void LogError(void)
     NL_systemStatus.TCD_usbJams++;
 }
 
+static inline void FillBuffer(uint8_t const b1, uint8_t const b2, uint16_t const w)
+{
+  if (bufHeadIndex >= BUFFER_SIZE)
+  {  // current head is full, switch to next one
+    uint16_t tmpBufHead = (bufHead + 1) & NUMBER_OF_BUFFERS_MASK;
+    if (tmpBufHead == bufTail)  // no space left ?
+    {
+      LogError();  // ring of buffers is full
+      return;      // discard incoming data
+    }
+    bufHead      = tmpBufHead;
+    bufHeadIndex = 0;
+  }
+  buff[bufHead][bufHeadIndex++] = b1;
+  buff[bufHead][bufHeadIndex++] = b2;
+  buff[bufHead][bufHeadIndex++] = w >> 7;    // first 7 bits
+  buff[bufHead][bufHeadIndex++] = w & 0x7F;  // second 7 bits
+}
+
 /******************************************************************************/
 /**	@brief  SendMidiBuffer - sends the USB bulk
 *******************************************************************************/
-void MSG_SendMidiBuffer(void)
+void MSG_Process(void)
 {
   if (dropMessages)
   {  // discard all messages
@@ -78,14 +99,6 @@ void MSG_SendMidiBuffer(void)
     bufTail      = 0;
     bufHeadIndex = 0;
     return;
-  }
-
-  if (bufHeadIndex >= BUFFER_SIZE)
-  {  // current head is full, switch to next one
-    bufHead      = (bufHead + 1) % NUMBER_OF_BUFFERS;
-    bufHeadIndex = 0;
-    if (bufHead == bufTail)
-      LogError();  // ring of buffers is full
   }
 
   if (bufHead == bufTail && bufHeadIndex == 0)
@@ -98,21 +111,24 @@ void MSG_SendMidiBuffer(void)
   {  // send stashed buffers first
     if (!USB_MIDI_Send(buff[bufTail], BUFFER_SIZE, 0))
       LogError();  // send failed
-    bufTail = (bufTail + 1) % NUMBER_OF_BUFFERS;
+    bufTail = (bufTail + 1) & NUMBER_OF_BUFFERS_MASK;
     return;
   }
   if (bufHeadIndex != 0)
   {  // send current buffer
     if (!USB_MIDI_Send(buff[bufHead], bufHeadIndex, 0))
       LogError();  // send failed
-    bufHead      = (bufHead + 1) % NUMBER_OF_BUFFERS;
+    bufHead      = (bufHead + 1) & NUMBER_OF_BUFFERS_MASK;
     bufHeadIndex = 0;
     bufTail      = bufHead;
   }
 }
 
+#if LPC_KEYBED_DIAG
 int16_t         TCD_keyOnOffCntr[128];
 static uint16_t keyOnOffIndex;
+#endif
+
 /*****************************************************************************
 *	@brief  MSG_KeyPosition
 *	Sends the index of the selected key before sending the On/Off velocity
@@ -122,13 +138,10 @@ void MSG_KeyPosition(uint32_t key)
 {
   if ((key < 36) || (key > 96))
     return;  /// assertion
-  buff[bufHead][bufHeadIndex++] = AE_TCD_WRAPPER;
-  buff[bufHead][bufHeadIndex++] = AE_TCD_KEY_POS;
-  buff[bufHead][bufHeadIndex++] = key >> 7;    // first 7 bits
-  buff[bufHead][bufHeadIndex++] = key & 0x7F;  // second 7 bits
-
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_KEY_POS, key);
+#if LPC_KEYBED_DIAG
   keyOnOffIndex = key;
-  MSG_SendMidiBuffer();
+#endif
 }
 
 /*****************************************************************************
@@ -138,13 +151,10 @@ void MSG_KeyPosition(uint32_t key)
 ******************************************************************************/
 void MSG_KeyDown(uint32_t vel)
 {
-  buff[bufHead][bufHeadIndex++] = AE_TCD_WRAPPER;
-  buff[bufHead][bufHeadIndex++] = AE_TCD_KEY_DOWN;
-  buff[bufHead][bufHeadIndex++] = vel >> 7;    // first 7 bits
-  buff[bufHead][bufHeadIndex++] = vel & 0x7F;  // second 7 bits
-
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_KEY_DOWN, vel);
+#if LPC_KEYBED_DIAG
   TCD_keyOnOffCntr[keyOnOffIndex]++;
-  MSG_SendMidiBuffer();
+#endif
 }
 
 /*****************************************************************************
@@ -154,13 +164,10 @@ void MSG_KeyDown(uint32_t vel)
 ******************************************************************************/
 void MSG_KeyUp(uint32_t vel)
 {
-  buff[bufHead][bufHeadIndex++] = AE_TCD_WRAPPER;
-  buff[bufHead][bufHeadIndex++] = AE_TCD_KEY_UP;
-  buff[bufHead][bufHeadIndex++] = vel >> 7;    // first 7 bits
-  buff[bufHead][bufHeadIndex++] = vel & 0x7F;  // second 7 bits
-
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_KEY_UP, vel);
+#if LPC_KEYBED_DIAG
   TCD_keyOnOffCntr[keyOnOffIndex]--;
-  MSG_SendMidiBuffer();
+#endif
 }
 
 /*****************************************************************************
@@ -173,11 +180,7 @@ void MSG_HWSourceUpdate(uint32_t source, uint32_t position)
 {
   if ((source >= NUM_HW_REAL_SOURCES) || (position > 16000))
     return;  /// assertion
-  buff[bufHead][bufHeadIndex++] = AE_TCD_WRAPPER;
-  buff[bufHead][bufHeadIndex++] = AE_TCD_HW_POS | source;
-  buff[bufHead][bufHeadIndex++] = position >> 7;    // first 7 bits
-  buff[bufHead][bufHeadIndex++] = position & 0x7F;  // second 7 bits
-  MSG_SendMidiBuffer();
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_HW_POS | source, position);
 }
 
 /*****************************************************************************
@@ -187,11 +190,7 @@ void MSG_HWSourceUpdate(uint32_t source, uint32_t position)
 ******************************************************************************/
 void MSG_SendAEDevelopperCmd(uint32_t cmd)
 {
-  buff[bufHead][bufHeadIndex++] = AE_TCD_WRAPPER;
-  buff[bufHead][bufHeadIndex++] = AE_TCD_DEVELOPPER_CMD;
-  buff[bufHead][bufHeadIndex++] = cmd >> 7;    // first 7 bits
-  buff[bufHead][bufHeadIndex++] = cmd & 0x7F;  // second 7 bits
-  MSG_SendMidiBuffer();
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_DEVELOPPER_CMD, cmd);
 }
 
 /*****************************************************************************
@@ -201,10 +200,8 @@ void MSG_SendAEDevelopperCmd(uint32_t cmd)
 void MSG_SendActiveSensing(void)
 {
 #ifdef __IMPLEMENT_ACTIVE_SENSING
-  buff[bufHead][bufHeadIndex++] = 0x0F;  // cable 0, packet type "single byte"
-  buff[bufHead][bufHeadIndex++] = 0xFE;  // MIDI real-time command "active sensing"
-  buff[bufHead][bufHeadIndex++] = 0x00;
-  buff[bufHead][bufHeadIndex++] = 0x00;
-  MSG_SendMidiBuffer();
+  FillBuffer(0x0F,  // cable 0, packet type "single byte"
+             0xFE,  // MIDI real-time command "active sensing"
+             0);
 #endif
 }
--- a/projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
+++ b/projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
@@ -1,7 +1,7 @@
 /******************************************************************************/
 /** @file		nl_tcd_msg.c
-    @date		2013-04-23
-    @version	0.02
+    @date		2020-06-19 KSTR
+    @version	1.0
     @author		Stephan Schmitt [2012-06-15]
     @brief		<tbd>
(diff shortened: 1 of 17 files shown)


KlausStrohhaecker:
Currently, there are 8 buffer 64 bytes (16 TCD messages) each = 128 TCD mesages max.
I tested down to 2 buffer 4 bytes, and above 4 x 4 buffers there were no obvious audible hiccups anymore (and none in the LPC status as well), so 8 x 64 should be safe even with more TCD traffic from HW sources etc.
Tested with manual playing and/or with key emulation test script using clustered key events (21 notes):
```
#!/bin/sh
while true; do
 lpc key 40 1; lpc key 41 1; lpc key 42 1; lpc key 43 1
 lpc key 44 1; lpc key 45 1; lpc key 46 1; lpc key 47 1
 lpc key 48 1; lpc key 49 1; lpc key 50 1; lpc key 51 1
 lpc key 52 1; lpc key 53 1; lpc key 54 1; lpc key 55 1
 lpc key 56 1; lpc key 57 1; lpc key 58 1; lpc key 59 1
 lpc key 60 1;
 sleep 0.4
 lpc key 40 -1; lpc key 41 -1; lpc key 42 -1; lpc key 43 -1
 lpc key 44 -1; lpc key 45 -1; lpc key 46 -1; lpc key 47 -1
 lpc key 48 -1; lpc key 49 -1; lpc key 50 -1; lpc key 51 -1
 lpc key 52 -1; lpc key 53 -1; lpc key 54 -1; lpc key 55 -1
 lpc key 56 -1; lpc key 57 -1; lpc key 58 -1; lpc key 59 -1
 lpc key 60 -1
 sleep 1
done
```
---------
They keybed scanner round-trip time has been increased to 24us as I found the occasional keybed IRQ overrun event. Overruns don't corrupt scanner data but add jitter/uncertainity to the timing, also for the ADC scanner timing. 24us is still 5x higher timing resolution than the legacy 125us round-trip.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -16,19 +16,21 @@
 #include "shared/lpc-defs.h"
 
 /******************************************************************************/
-/*	modul local defines														  */
+/*	module local defines														  */
 /******************************************************************************/
 
-#define BUFFER_SIZE 512
+#define BUFFER_SIZE       (64)
```
hhoegelo, line 22: I do not understand, why we have two levels of buffering here. From what I can see in the code, each message is only 4 bytes, so I'd imagine a ring buffer of N messages. IMHO, there is no need for bufHeadIndex - but maybe I just don't see it.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -47,55 +49,65 @@ void MSG_CheckUSB(void)  // every 200 ms
...
+    bufHead      = (bufHead + 1) % NUMBER_OF_BUFFERS;
+    bufHeadIndex = 0;
+    if (bufHead == bufTail)
+      LogError();  // ring of buffers is full
+  }
+
+  if (bufHead == bufTail && bufHeadIndex == 0)
+    return;  // nothing to send
+
+  if (USB_MIDI_BytesToSend() != 0)
```
hhoegelo, line 94: Who takes care about retries? If there are still some bytes in flight, the function will just return and only the next incoming event will finally send the data, no?

KlausStrohhaecker: Yes, the function keeps buffering the data until the next USB transfer can be started. But the buffer logic is still somewhat inefficient and needs changes, see next commits.

In case of true USB fails there are no retries (and have never been), as well as when all buffers finally overrun (dto).
When USB fails (intermittent cable contact, etc), the AE restarts anyway.

hhoegelo: I see the buffering, but I don't see the retry. Can you please point me to the place where the retry in response to free space in the usb stack actually happens?

KlausStrohhaecker: There never was a retry on the high protocol level. If there is anything inside the USB-stack I'm not aware of it.

KlausStrohhaecker: FYI, Nemanjas legacy code didn't even recover after a USB fail / dropout

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -47,60 +51,84 @@ void MSG_CheckUSB(void)  // every 200 ms
...
-        NL_systemStatus.TCD_usbJams++;
-    }
+  if (bufHead == bufTail && bufHeadIndex == 0)
+    return;  // nothing to send
+
+  if (USB_MIDI_BytesToSend() != 0)
+    return;  // last transfer still in progress
 
-    buf = 0;  // this discard the messages in case of error, no re-send attempts
+  if (bufHead != bufTail)
```
hhoegelo, line 110: Maybe this should be a while() ? Probably its not needed, but from correctness point of view, as long as USB_MIDI_Send accepts data, we don't need to add a delay here.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -109,21 +137,11 @@ static uint16_t keyOnOffIndex;
...
-  {
     return;  /// assertion
-  }
-
-  buff[writeBuffer][buf++] = AE_TCD_WRAPPER;
-  buff[writeBuffer][buf++] = AE_TCD_KEY_POS;
-  buff[writeBuffer][buf++] = key >> 7;    // first 7 bits
-  buff[writeBuffer][buf++] = key & 0x7F;  // second 7 bits
-
+  FillBuffer(AE_TCD_WRAPPER, AE_TCD_KEY_POS, key);
```
hhoegelo, line 141: This is much better than before!

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -14,21 +14,25 @@
...
+#include "shared/globals.h"
 
 /******************************************************************************/
-/*	modul local defines														  */
+/*	module local defines														  */
 /******************************************************************************/
 
-#define BUFFER_SIZE 512
+#define BUFFER_SIZE            (256)  // space for 256 / 4 = 64 TCD events per buffer
+#define NUMBER_OF_BUFFERS      (8)    // number of buffers, must be 2^N !
```
hhoegelo, line 24: I still don't get why a "buffer of buffers of bytes" is better than a "buffer of bytes" here. I'd even prefer a "Buffer of MidiMessages". But maybe there is a reason that I just can't see.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -47,60 +51,84 @@ void MSG_CheckUSB(void)  // every 200 ms
...
+static void LogError(void)
+{
+  DBG_Led_Error_TimedOn(10);
+  if (NL_systemStatus.TCD_usbJams < 0xFFFF)
+    NL_systemStatus.TCD_usbJams++;
+}
+
+static inline void FillBuffer(uint8_t const b1, uint8_t const b2, uint16_t const w)
+{
+  if (bufHeadIndex >= BUFFER_SIZE)
```
hhoegelo, line 74: should actually be (BUFFER_SIZE - 4) as you will eat up 4 bytes later

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -14,21 +14,25 @@
...
 /*	modul local variables													  */
 /******************************************************************************/
 
-static uint8_t  buff[2][BUFFER_SIZE] = {};  // Two buffers for MIDI sending, alternately used
-static uint32_t buf                  = 0;   // Counts the used bytes of the bulk buffer
-
-static uint32_t writeBuffer = 0;
+// ringbuffer of buffers
+static uint8_t  buff[NUMBER_OF_BUFFERS][BUFFER_SIZE];
+static uint16_t bufHead      = 0;  // 0..(NUMBER_OF_BUFFERS-1)
```
hhoegelo, line 33: Consider writeBufferIdx and readBufferIdx instead of bufHead and bufTail.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -14,21 +14,25 @@
...
+static uint16_t bufTail      = 0;  // 0..(NUMBER_OF_BUFFERS-1)
+static uint16_t bufHeadIndex = 0;  // used bytes in the current head (front) buffer
```
hhoegelo, line 35: Consider IdxInWriteBuffer or something similar.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -47,60 +51,84 @@ void MSG_CheckUSB(void)  // every 200 ms
...
+  if (USB_MIDI_BytesToSend() != 0)
+    return;  // last transfer still in progress
 
-    buf = 0;  // this discard the messages in case of error, no re-send attempts
+  if (bufHead != bufTail)
+  {  // send stashed buffers first
+    if (!USB_MIDI_Send(buff[bufTail], BUFFER_SIZE, 0))
+      LogError();  // send failed
+    bufTail = (bufTail + 1) & NUMBER_OF_BUFFERS_MASK;
+    return;
```
hhoegelo, line 115: As long as USB_MIDI_Send accepts data, we do not need to return here, do we?

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
```diff
@@ -47,60 +51,84 @@ void MSG_CheckUSB(void)  // every 200 ms
...
+    return;  // nothing to send
+
+  if (USB_MIDI_BytesToSend() != 0)
+    return;  // last transfer still in progress
 
-    buf = 0;  // this discard the messages in case of error, no re-send attempts
+  if (bufHead != bufTail)
+  {  // send stashed buffers first
+    if (!USB_MIDI_Send(buff[bufTail], BUFFER_SIZE, 0))
+      LogError();  // send failed
```
hhoegelo, line 113: Is it really an error if USB_MIDI_send does not accept data right now? I suggest:
USB_MIDI_Send returns true: data accepted
USB_MIDI_Send returns false: try again later (will be done anyway)

If there is really an error in sending usb midi, USB_MIDI_Send should log that internally - probably it can even explain better, what exactly isthe problem.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: Assume one packet is under way, USB_MIDI_BytesToSend() will continue to report != 0 (not done) and for some reason it takes longer than 1 time-slice to finish or, alternatively, MSG_Process() is called more than once (this was the trigger for the hanging-notes bug to appear, btw). 
With this, you have only capabilities for one pending call, then you run out of buffers (even though the buffers may contain only a single TCD message).

With 8 buffers, there is backlog of transfers that then complete with a short delay, and the re-fill of the buffers is completetly unaffected, unless you really run of fully filled buffer space.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: I deliberaty choose names NOT containing index and read/write cues to avoid confusion with the actual byte write process and its associated index. Buffer Head/Tail is widely used nomencature, isn't it?

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: Assume buffer size is 4 (a legal size). Then checking for >= (BUFFER_SIZE - 4) will always be true. The current buffer, while empty (containg stale data),  will be judged as full and will be stashed for later transfer and the next buffer will be selected ==> data salat. With larger buffers, likewise, the last four bytes will never be filled but will be transfered.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: Ähm, LPC is all real-time step-chain style single-thread coding (apart from some DMA), no loops possible unless it is exactly known how long the function takes and how many times it must be polled ;-)

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: I took over the logic from the original coder, which was : 
1 - poll USB_MIDI_BytesToSend() until it signal "done/ready"
2 - start USB transfer with USB_MIDI_Send(), which is and must be a non-blocking function, it starts the DMA in the end. Back to 1.
Strictly a one message at a time strategy (usually a good choice)

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
KlausStrohhaecker: I don't know if it is good idea to try to complety overhaul the various bufferings and their interplay at the moment, and see above, one message block at a time, if it works lets keep it.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
hhoegelo: What I actually mean is: in case of, for example, bufHeadIndex == (BUFFER_SIZE - 1), condition in 74 will be false, so in line 87 the function will overwrite foreign memory.

hhoegelo: Now, that I look deeper into the code, I wonder why there is a ring buffer in the tcd midi layer at all. USB_MIDI_Send already puts data into a backlog buffer, if it can not be sent right now. 
Maybe we could remove all buffering in nl_tcd_msg and simply rely on proper buffering in nl_usb_midi? This would reduce complexity, save some copy operations and even remove the delay between FillBuffer and MSG_Process-calls. Of course, USB_MIDI_Send() needs to handle the backlog before trying to use the usb device.

projects/realtime-system/src/main/src/tcd/nl_tcd_msg.c
hhoegelo: I just tell you my thoughts and findings. It is not to be understood as "Please change this!" but more as "Could be improved" or "This is not obvious to the reader.". In the end, **you** have to gauge risks and code quality and costs of refactoring.

KlausStrohhaecker:
OK, I've set up a hardcore torture test to check if the buffering is robust, see [nl_tcd_poly.c](https://github.com/nonlinear-labs-dev/C15/blob/4763dc296c13102634669a0818881a7d9d822e81/projects/realtime-system/src/main/src/tcd/nl_tcd_poly.c), lines 197ff.

Short form: generate 61 key events at once, within a single time slice of 125us. on/off type of events is changed after every call, otherwise the AE doesn't output sound when sending the note-offs in the same go (too fast for, or considered by, the AE, it seems).

Let the LPC tool trigger this as fast as possible:
` while true; do lpc key 0 0; sleep 0; lpc key 0 0; sleep 0; done` (the `sleep 0` seems to be required because without it I sometimes got no ouput from the AE when using single shot-test sequences (`lpc key 0 0; lpc key 0 0`), probably still to fast -- without any other signs of errors, though).

I let this run for about 5 minutes while using the C15 normally, switching presets, use the hardware sources, additional manual playing. Depending on the playload on the Bridge, an increased pause time between burst was noticed (still only fractions of a seconds, though) which is OK for obvious reasons (bridge may eat some 40++% of CPU, according to `top)`

Then I checked if there are any errors as indicated by the LPC status data, also checking BBB system log with `dmesg` (where any LPC<-->BBB communcation errors are logged) and _everything was OK, no buffer overruns or other critical events:_
```
# lpc-read -h -r
Output from '/dev/lpc_bb_driver' :
0000000000 SYSTEM STATUS:
cleared with 'lpc req clear-status':
  M4 ticker      :   64003725 ticks since last call
  Scheduler      :     0 task overruns
  Scheduler      :     2 tasks max. per time-slice
  Scheduler      :    85us max. task runtime
  Scheduler      :    95us max. time-slice
  Midi Buffers   :     0 dropped
  BBB Msg        :     0 buffer overruns / ESPI send fails
  TCD Msg        :     0 buffer overruns / USB send fails
cleared after every call:
  M0 ADC Scanner :   800us max. round trip time
  M0 Key Scanner :     0 (overrun flag)
```
-------

Further, I used `while true; do lpc key 0 0; sleep 0; lpc key 0 0; sleep 0.5; done` to send 61 note-on, followed by 61-note offs immediatly, then pausing a noticable amount of time.

With this, _hanging notes_ would be readily apparent acoustically but did never occur in more than 20mins of testing. I also listened for uniformity of that ultra-high-speed sort of glissando sound and _it was also very OK_ (though not perfect), with exception of the occasional click/glitch when reverb was present on a preset.

Sometimes one could hear that this very fast note sequence was a bit slower or sort of fractioned, I would assume this rather happens on the AE side (getting less than normal CPU percentage sometimes) than in the LPC, or it is something USB related....

---------
#### Others may want to double check this but I would think the MIDI-buffer issue can be considered settled now, code really behaved well. Nonetheless, polishing up the code might be worth it but not at this time, I'd suggest

PR #2002, opened 2020-06-23
moved preset select/load into mouseUp

### LINKED ISSUE #2001: WebUI: Preset select and load on Pointer Up instead of down
reacting on pointer down leads to problems when trying to drag presets with direct load enabled

schlenkibus:
closes #2001


majjam:
Does that mean that dragging will be a two-click action ?
1st click and release for selecting the preset.
2nd click and hold for dragging ?

PR #2015, opened 2020-06-24
move to aws

hhoegelo:


KlausStrohhaecker:
Links look OK.
I'll do an overnight fresh build and report back any issue.

PR #2033, opened 2020-06-30
added webui infos

schlenkibus:


KlausStrohhaecker:
grafik

I suggest moving the date/time field right after "free memory" (which might be better called "free disk space"), before the delimiter line, and name it "system date/time".
Then add "commit date" field where system date/time previously was

schlenkibus:
@KlausStrohhaecker updated WebUI according to your comment

PR #2066, opened 2020-07-09
Update cosmetic repairs

AntonSchmied:
Sorry for the late PR.
This one includes minimal changes, which improve the update behavior!   
Chose the wrong branch for the PR first ...

### DIFF
--- a/build-tools/create-c15-update/update_scripts/bbb_update.sh
+++ b/build-tools/create-c15-update/update_scripts/bbb_update.sh
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-#
-# Author: Anton Schmied
 # version : 2.0
 #
 # ----------- install a BBB Updates from USB-Stick ---------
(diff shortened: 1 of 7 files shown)


KlausStrohhaecker:
Hh, previous "install-update.sh" always freezes, new one never does, and run.sh, in a subsequent update, will return in case of success and thus system will continue after the update but will be inoperable (bridge stopped). Net effect is the same though, "restart" message will be displayed and C15 is stalled in either case and waiting for the the user to power-cycle.

Wouldn't the freeze() collide with the "reboot" option for OTA? That is, if there is any tiny error the OTA will probably leave the C15 bricked for further remote access?

AntonSchmied:
let me have a think for a moment ...

KlausStrohhaecker:
Most likely this are non-issues for the current release and things can be postponed.

AntonSchmied:
I will have a think about what you mentioned! and then i'll get back to the PR

AntonSchmied:
so, i still have to run a test or two just to be sure, but what do you think of this solution @KlausStrohhaecker?
concerning the ota <-> freeze: so far a manual restart of the C15 is the only solution, even if something goes south, yes .
I would suggest we think about this later on, since atm. it does not concern the users out there?

AntonSchmied:
Tested: 
1905 -> 1.7: run.sh freezes on fail (ePC unpingable) instead of exiting to install-update.sh
1.5 -> 1.7: run.sh freezes on fail (ePC unpingable) instead of exiting to install-update.sh
1.7 -> 1.7: run.sh exits with 1 on fail (ePC unpingable) and install-update.sh freezes

KlausStrohhaecker:
> so, i still have to run a test or two just to be sure, but what do you think of this solution @KlausStrohhaecker?
concerning the ota <-> freeze: so far a manual restart of the C15 is the only solution, even if something goes south, yes .
I would suggest we think about this later on, since atm. it does not concern the users out there?

We will have the occasional BBB network failure during update causing a stall/freeze but that is probably tolerable, we just need to tell the users that they must not worry and simply try again (power-cycle).

PR #2079, opened 2020-07-20
fixes #2078

### LINKED ISSUE #2078: WebUI: Bank Info Changes does not change current WebUI
e.g. Bank Position changed via WebUI Bank Info moves the bank in the boled and other WebUI's but not in the WebUI the change was performed

schlenkibus:
hotfixes #2078


schlenkibus:
> this fixes #2078 but:
> shouldn't this PR be targeted at `RELEASE-2020-06` instead of `20-29`? (as 20-29 already is final)

agreed, i changed the branch! We should talk about that branching issue again soon

PR #2086, opened 2020-07-22
rework preset drag behavior and add open tab behavior on preset drag

### LINKED ISSUE #2068: Preset Drag&Drop into Sound Tab vs auto Panning in Main Area
When dragging a Preset into the Sound Tab, your movement has to be sufficiently fast, otherwise the dragged Preset will stick to the Sound Tab border and downward auto-panning will be activated.

Once in panning mode, it can be difficult and take a while to stop.

Due to the panning, the visible part of the Main Area may completely change, which is quite inconvenient when in the middle of assembling dual Presets the "fast" way.

Maybe, auto-panning of dragged Presets should only activate, when definitely hitting a true corner, but not the Sound Tab border?

schlenkibus:
added drag handler for belt preset/sound button -> enables opening / switching to belt preset/sound tab for dropping into that list / loading to part

fixes #2068

### DIFF
--- a/projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/PresetBeltButton.java
+++ b/projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/PresetBeltButton.java
@@ -1,13 +1,19 @@
 package com.nonlinearlabs.client.world.overlay.belt;
 
 import com.nonlinearlabs.client.world.Control;
+import com.nonlinearlabs.client.world.IPreset;
 import com.nonlinearlabs.client.world.Position;
+import com.nonlinearlabs.client.world.overlay.DragProxy;
 import com.nonlinearlabs.client.world.overlay.OverlayLayout;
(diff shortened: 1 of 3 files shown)


projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/PresetBeltButton.java
```diff
@@ -1,13 +1,19 @@
...
 import com.nonlinearlabs.client.world.Control;
+import com.nonlinearlabs.client.world.IPreset;
 import com.nonlinearlabs.client.world.Position;
+import com.nonlinearlabs.client.world.overlay.DragProxy;
 import com.nonlinearlabs.client.world.overlay.OverlayLayout;
+import com.nonlinearlabs.client.world.overlay.belt.Belt.BeltTab;
 
 public class PresetBeltButton extends BeltButton {
 
+	private Belt m_belt;
```
hhoegelo, line 12: BeltButton already has a belt ptr

PR #2118, opened 2020-08-11
added not received placeholder for RTVersion

### LINKED ISSUE #2117: RT version sometimes not fetched/displayed (never works in 20-32-RC3)
In earlier versions than 20-32-RC3, we see that the RT version number sometimes is not fetched and then displays "0" (which is still a bug, btw, it should display "-" or "n/a").

In 20-32-RC3, the LPC version number is never fetched. Fetching the version number manually with the LPC tools works as expected, so it seems to be purely PG-related. Another hint is that restarting the PG, which should initiate the fetch, also just does not do that.

schlenkibus:
closes #2117

### DIFF
--- a/projects/playground/src/proxies/lpc/LPCProxy.cpp
+++ b/projects/playground/src/proxies/lpc/LPCProxy.cpp
@@ -129,7 +129,7 @@ void LPCProxy::onNotificationMessageReceived(const MessageParser::NLMessage &msg
   uint16_t id = msg.params[0];
   uint16_t value = msg.params[1];
 
-  nltools::Log::warning("received ID:", id, "from LPC with value:", value);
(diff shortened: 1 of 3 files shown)


projects/playground/src/proxies/lpc/LPCProxy.cpp
```diff
@@ -129,6 +129,8 @@ void LPCProxy::onNotificationMessageReceived(const MessageParser::NLMessage &msg
   uint16_t id = msg.params[0];
   uint16_t value = msg.params[1];
 
+  nltools::Log::info("received ID:", id, "from LPC with value:", value);
+
```
hhoegelo, line 133: Do we still need this info?

schlenkibus: nope, fixed

PR #2130, opened 2020-08-13
2122

### LINKED ISSUE #2122: WebUI: Wenn man "Highlight Changed Parameters" ausschaltet, sollten auch die gelben Rahmen im MC-Mode des Preset-Tabs unterdrückt werden.

schlenkibus:
modulation aspect buttons now take highlight changed setting into account when deciding to draw yellow border, closes #2122

### DIFF
--- a/projects/NonMaps/src/main/java/com/nonlinearlabs/client/presenters/ParameterPresenterProvider.java
+++ b/projects/NonMaps/src/main/java/com/nonlinearlabs/client/presenters/ParameterPresenterProvider.java
@@ -76,6 +76,16 @@ public class ParameterPresenterProvider extends Notifier<ParameterPresenter> {
 			updatePresenter(p);
 			return true;
 		});
+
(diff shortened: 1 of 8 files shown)


schlenkibus:
> In this PR, changing the Setup Setting will (like intended) update the display of yellow Frames around Modulation Icons for changed parameters (the frame only being showed when the setting is on).
> 
> But I noticed that changing the Setup Setting will not have an effect on already changed parameters (more precise: positions). Yellow Frames in the Main Area will still be displayed after disable, changed parameters will not be drawn with yellow frame after enable (only after loading a Preset and starting from scratch will then behave correctly).
> So, each time the setting changes all parameters and aspects should be evaluated.

I fixed mentioned issue with refresh of parameters. I also fixed a bug where after starting the C15 the recall values were not properly loaded.

hhoegelo (approved):

PR #2134, opened 2020-08-13
Issue 2047

### LINKED ISSUE #2047: WebUI: Compare Two Presets: bei zwei Single-Sounds die globalen Parameter (Master Level etc.) mit erfassen und die "Part I/II"-Menüs ausblenden.

### LINKED ISSUE #2137: [Compare] Dual Presets - include display of global parameters
see https://github.com/nonlinear-labs-dev/C15/pull/2134#pullrequestreview-468361995

- Split and Layer Sounds should also display **global** parameters (Master Volume, Tune) if they have changed (they should be independent from the Part selection menu and always be shown)
- additionally, Split Sounds treat the Split Point wrong (it will be treated as a Fade From when comparing two opposing Parts, but should rather be an additional global parameter)

schlenkibus:
closes #2047 
closes #2137
and added compare of different sized presets -> missing groups and parameters


projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/CompareDialog.java
```diff
@@ -481,6 +494,18 @@ public void addVoiceGroupMenus(int row, FlexTable table) {
...
 	}
 
+	private boolean hideParameter(int id, SoundType type) {
+		switch(type) {
+			case Single:
+				return id == 356 || id == 396 || id == 397;
+			case Split:
+				return id == 396 || id == 397;
+			case Layer:
+				return id == 356;
```
MatthiasSeeber, line 504: Question: can we somehow avoid pure written Parameter IDs? I'd much prefer an enum-like approach..

schlenkibus:
regarding your fist question: i think we came to the conclusion that comparing hw positions brings no real value 

regarding your second point: i will have to think about that a bit more, as we are not transparent about groups in layer / split sounds -> split point being global but appear as part of part in UI and mono / unison being available in both parts in split and only available in part I of layer sounds.

projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/CompareDialog.java
schlenkibus: Theoreticaly this should be possible, but ATM we do not have a Parameter Enum ready to use

MatthiasSeeber:
> regarding your fist question: i think we came to the conclusion that comparing hw positions brings no real value
> 

What about hw amounts? I think they are worth showing..

> regarding your second point: i will have to think about that a bit more, as we are not transparent about groups in layer / split sounds -> split point being global but appear as part of part in UI and mono / unison being available in both parts in split and only available in part I of layer sounds.

I get that this may be a bit tricky. I just suspect that the User might be able to compare a Layer Part II Voices / Mono, which shouldn't exist.

MatthiasSeeber:
At this stage, the basic fix can be considered successful.
I opened #2141, #2142, #2143 as separate tickets.
@schlenkibus feel free to proceed as you like.

projects/playground/src/presets/Preset.cpp
```diff
@@ -398,13 +398,32 @@ void Preset::writeDiff(Writer &writer, const Preset *other, VoiceGroup vgOfThis,
...
+
+  std::vector<std::string> writtenGroups;
+
   for(auto &g : m_parameterGroups[static_cast<size_t>(vgOfThis)])
+  {
     g.second->writeDiff(writer, g.first, other->findParameterGroup({ g.first.getName(), vgOfOther }));
+    writtenGroups.emplace_back(g.first.getName());
+  }
+
+  if(numGroups < numOtherGroups)
```
hhoegelo, line 418: I don't think this is sufficient. It might - at least theoretically - be possible, that A contains Groups 1, 2, 3 while B contains Groups 1, 3, 4. No entry for group 4 will be generated.

projects/playground/src/presets/PresetParameterGroup.cpp
```diff
@@ -66,16 +66,43 @@ void PresetParameterGroup::writeDiff(Writer &writer, const GroupId &groupId, con
...
+    std::vector<int> writtenParameters;
+
     for(auto &parameter : m_parameters)
     {
       auto otherParameter = other->findParameterByID({ parameter.first.getNumber(), other->getVoiceGroup() });
       parameter.second->writeDiff(writer, parameter.first, otherParameter);
+      writtenParameters.emplace_back(parameter.first.getNumber());
+    }
+
+    if(parameterCount < otherParameterCount)
```
hhoegelo, line 91: Count might not be sufficient - please see my previous comment.

projects/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ParameterLayout.cpp
```diff
@@ -40,6 +41,14 @@ ParameterLayout2::ParameterLayout2()
 
   Application::get().getPresetManager()->getEditBuffer()->onSoundTypeChanged(
       sigc::hide(sigc::mem_fun(this, &ParameterLayout2::onSoundTypeChanged)), false);
+
+  auto current = getCurrentParameter();
+  auto hwui = Application::get().getHWUI();
+  if(hwui->getCurrentVoiceGroup() == VoiceGroup::II && !current->getID().isGlobal()
```
hhoegelo, line 47: It does not look right to me, that creation of a layout is needed to fix a insane state in hwui. Lets try to figure out a less 'drive-by'-kind-of fix.

MatthiasSeeber (approved):

hhoegelo (approved):

PR #2144, opened 2020-08-19
Huge restructureing

hhoegelo:
This PR is probably far too large for a proper review. 
I am open to suggestions on how to deal with this. 

My suggestion is to just try all the make-targets (I tested them on my PC) and more or less implicitly test the resulting images during our normal work.

@KlausStrohhaecker: There is something I could not solve: during update, there is a call "/update/utilities/playcontroller reset 5" which fails:

LPC is down
LPC is down
LPC is down
LPC is down
LPC is down
>>> LPC reset failed after 10 retries

I have no idea why this broke by renaming most of the 'lpc' occurrences in the repo with 'playcontroller'. Do you have a clue?


KlausStrohhaecker:
> @KlausStrohhaecker: There is something I could not solve: during update, there is a call "/update/utilities/playcontroller reset 5" which fails:
LPC is down
LPC is down
LPC is down
LPC is down
LPC is down

This happens when either someone else steals LPC messages (like a running bridge, or lpc-read tool) or when the LPC actually doesn't respond after reset.

Renaming should have zero effect.
As the main user of the LPC tools I would vote against using such a bulky name like "playcontroller". This is internal use only, so I'd prefer to keep the historical name or find another short one (3 characters)

PR #2153, opened 2020-08-24
fixes #2112

### LINKED ISSUE #2112: [Soled] consider extending Ribbon's "last touched" indicator
A `<` indicates the last touched Ribbon, when the Soled is in Play or Edit mode.

In many cases (if a Ribbon is absolute), you will change the Ribbon just by touching it. This is inconvenient when only wanting to change the "last touched" state. We can consider integrating Parameter Selection into this, chaning the state each time, when Ribbon 1 or 2 are selected.

schlenkibus:
@Matthias does that behavior work as you intended it to be?

+ fix for BBBB emulation ribbons, they did not work since the HW Source -> EHC umbau


MatthiasSeeber (approved):
Yes, it does precisely that!
And nice to have the Ribbons back on the BBBB!

PR #2154, opened 2020-08-26
Ribbon Calibration tools self-contained on BBB

KlausStrohhaecker:
Environment for standalone ribbon calibration on-site, running natively on BBB with nothing else required other than a remote terminal. This very useful to update when updating older C15's in the field.

- [x] extend "lpc-read" tool capability to nicely display raw ribbon values for on-site calibration.
- [x] move the calibration compiler "make-cal-bin" from /realtime-system/supplements to /lpc-tools to be available on-site.
- [x] a helper script "readout-ribbon-adcs.sh" is now available and deployed on the target to simplify on-site calibration.
- [x] templates for the ribbon calibration source files are deployed at /usc/C15/scripts/calibration
- [x] fixed a bug in BBB espi driver for the ribbon LEDs, it got confused when a message contained an odd number of bytes whereas LED commands consist always of two bytes (LED_id and BitPattern).
- [x] added a reset feature to the BBB espi driver for the ribbon LEDs so that the step chain, which is was introduced in the fix above, can be reset to ensure commands following are interpreted correctly not matter what the previous state was.


hhoegelo (approved):
Personally, I'd prefer 'positive' flags in process-read-msgs.h, because for me statements like 

if(mask & NO_XYZ)  return;
or
if(!(mask & NO_XYZ)) doSomething();

are harder to understand than 

if(!(mask & YES_XYZ)) return;
if(mask & YES_XYZ) doSomething();

but that is only a matter of style, I suppose.

PR #2158, opened 2020-08-26
fixes #2027

### LINKED ISSUE #2027: WebUI [Modmatrix] - HW Amt Selection after setting Source as Return (to zero/center)
- ensure that in the WebUI Setup / GUI Settings / Edit Parameter == "always"
- take a Ribbon or Pedal Source with Return Behavior == "Non-Return"
--> the hw amounts will be switches, clicking on an amount will toggle the switch on/off

- now change behavior via right-click from Non-Return to Return (to zero or center)
- hw amounts will now be sliders, not switches
- now click on a particular amount
--> there is still the switch behavior, setting the slider to 100% (once - after editing or resetting the slider, this will not happen when selecting it - as it should)

schlenkibus:
fixes #2027


MatthiasSeeber (changes requested):
After setting HW Source from Non-Return to Zero or Center will produce bug in HW Amounts:
- they correctly become Sliders
- but they still behave like Switches
- the first click issue still persists

MatthiasSeeber (approved):

hhoegelo (approved):

PR #2161, opened 2020-08-27
hotfix for deleting the last bank

schlenkibus:
changed will not be true if no banks were present, now the manager itself will also play an role in that attribute


projects/playground/src/presets/PresetManager.cpp
```diff
@@ -713,8 +713,8 @@ void PresetManager::writeDocument(Writer &writer, UpdateDocumentContributor::tUp
                       m_editBuffer->writeDocument(writer, knownRevision);
 
                       bool anyBankChanged = false;
```
hhoegelo, line 715: How about initializing anyBufferChanged to didChangeSince() instead of first initializing and then changing?

MatthiasSeeber (approved):

hhoegelo (approved):

PR #2162, opened 2020-08-27
Issue 1840

schlenkibus:


projects/NonMaps/src/main/java/com/nonlinearlabs/client/WebSocketConnection.java
```diff
@@ -56,16 +56,17 @@ private void hideConnectionError() {
...
 
 		public void onServerUpdate(String text);
 	}
 
 	ServerListener listener;
-	static String clientId = Uuid.random();
+	String clientId;
 
 	public void connectToServer(ServerListener listener) {
+		clientId = Uuid.random();
```
hhoegelo, line 69: It is important, that a client instance uses the same clientId always, so the playground can close old web sockets tied to the same clientId. As it is now, each reconnect (e.g. in case of a connection error) will create a new web socket and keep the old one.

tl;dr: clientId shall be assigned at construction time of WebSocketConnection only.

projects/NonMaps/src/main/java/com/nonlinearlabs/client/integrationTests/OracleTests.java
```diff
@@ -0,0 +1,102 @@
...
+        deleteBank();
+
+        start();
+    }
+
+    private void selectPresetViaNewWebSocket() {
+        addStep(() -> {
+            webSocket.send("/presets/banks/select-preset?uuid=" + findBank().getPreset(0).getUUID().toString() + "&isOracle=true");
+        }, () -> {
+            boolean wasOracleOfServer = server.lastDocumentCouldOmitOracles();
```
hhoegelo, line 38: we have to wait for the server to receive the update here, don't we?
For example, we could store ServerProxys known updateId right before we send the command, then, in the 'done' callback we could require that the updateId actually changed AND the oracle flag is as expected.

projects/playground/src/http/ContentManager.cpp
```diff
@@ -225,9 +225,11 @@ bool ContentManager::feedWebSocket(const tWebsocketConnection &c)
 
     if(c->getLastSentUpdateId() != currentUpdateId)
     {
-      DebugLevel::info("Updating websocket", c->getConnection(), "with document for updateID", currentUpdateId);
+      auto canOmitOracles = c->canOmitOracles(currentUpdateId);
+      DebugLevel::info("Updating websocket", c->getConnection(), "with document for updateID", currentUpdateId,
```
hhoegelo, line 229: is the info still needed?

projects/playground/src/device-settings/DirectLoadSetting.cpp
```diff
@@ -8,11 +8,18 @@ DirectLoadSetting::DirectLoadSetting(Settings &settings)
 {
 }
 
-void DirectLoadSetting::load(const Glib::ustring &text)
+void DirectLoadSetting::load(const Glib::ustring &text, Initiator initiator)
 {
-  m_isLoading = true;
-  BooleanSetting::load(text);
-  m_isLoading = false;
+  if(initiator != Initiator::EXPLICIT_WEBUI)
```
hhoegelo, line 13: Why is HWUI different from WebUI regarding this? Maybe it should be (initiator == Initiator::Load) instead? If not, why did you add the Initiator::Load?

projects/NonMaps/src/main/java/com/nonlinearlabs/client/integrationTests/OracleTests.java
```diff
@@ -0,0 +1,128 @@
...
+        server.setSetting("DirectLoad", "off");
+    }
+
+    private void selectPresetViaNewWebSocket() {
+        addStep(() -> {
+            documentReceived = false;
+            webSocket.send("/presets/banks/select-preset?uuid=" + findBank().getPreset(0).getUUID().toString() + "&isOracle=true");
+        }, () -> {
+            boolean wasOracleOfServer = server.lastDocumentCouldOmitOracles();
+            return !wasOracleOfServer && documentReceived;
```
hhoegelo, line 50: Hmm, I am not sure: You assume that the server should have received an update document, because the local WebSocket did. I don't know if this is necessarily true. The logic as applied in 'selectPresetViaProxy' looks more solid to me (except the 'documentReceived'-part which should be dispensable.

hhoegelo (approved):

PR #2163, opened 2020-09-01
Issue 2108

### LINKED ISSUE #2108: [Boled] consider expanding Change Reset features
When a parameter was changed, pressing the Shift button will reveal this screen:

boled-8050686003847

When a modulation aspect (for example, amount) was changed, pressing the Shift button yields:

boled-8035698989268

Notice, that there is no way to compare or change between "old" and "new" values, like for the position. Instead, all possible changes to mod aspects (except upper/lower limits) can be reflected at once.

We should consider providing these features for the modulation aspects as well. The user has to focus on a mod aspect anyway. Pressing Shift could exclusively show the mod aspect (MC Pos, MC Sel, MC Amt) and offer an additional Soft Button in order to load both values. Further, the two remaining Soft Buttons could be used to navigate changed aspects. It could look something like this:

Concept

This screen could fit and reflect all position, mc pos, mc sel, mc amt (and maybe even upper/lower limit) changes, having one consistent behavior.

@nonlinear-labs what do you think?

schlenkibus:
closes #2108


hhoegelo (approved):

MatthiasSeeber (approved):

MatthiasSeeber (approved):
The min/max only triggers if position changes mechanism works.

I noticed that when in the Change Screen via the Shift Button, some Buttons (Setup, Preset, Sound) can be pressed and lead to the respective Screens. Returning to the Changed Screen by pressing such a button twice also works. However, some other Buttons (Default, Info) are bound to nothing when holding Shift. We may want to make the behavior consistent:
- either none of these buttons should change the screen while holding shift
- or all of these buttons should change the screen

hhoegelo (approved):

PR #2166, opened 2020-09-03
closes #1823

### LINKED ISSUE #1823: Toggle-Verhalten für den Default-Button und den Doppelklick auf einen Parameter in der WebUI.
- [x] Wenn man auf demselben Parameter die Default-Taste ein zweites Mal drückt, sollte der Wert vor dem Setzen auf Default wiederhergestellt werden. Beim dritten Mal dann wieder Default usw.
Damit kann man z.B. Signalwege ein- und ausschalten. Allerdings nur, solange kein anderer Parameter selektiert wird.

- [x] Das Gleiche kann in der WebUI für den Doppelklick auf einen Parameter realisiert werden. Solange die Parameter-Selektion unverändert ist, toggeln Doppelklicks zwischen ursprünglichem Wert und Defaultwert. (So etwas Ähnliches gibt es schon für den Zoom auf eine Gruppe und zurück.)

schlenkibus:
setDefault now takes argument whether it was called from recall or default, and if default it will allow toggling

closes #1823

### DIFF
--- a/projects/epc/playground/src/parameters/ModulateableParameter.cpp
+++ b/projects/epc/playground/src/parameters/ModulateableParameter.cpp
@@ -426,23 +426,17 @@ bool ModulateableParameter::isModAmountChanged() const
   if(getModulationSource() == MacroControls::NONE)
     return false;
 
-  if(auto original = getOriginalParameter())
-  {
-    const int denominator = static_cast<const int>(getValue().getFineDenominator());
-    const int roundedNow = static_cast<const int>(getModulationAmount() * denominator);
-    const int roundedOG = static_cast<const int>(original->getRecallModulationAmount() * denominator);
-    return roundedOG != roundedNow;
-  }
-  return false;
+  auto original = getOriginalParameter();
+  const int denominator = static_cast<const int>(getValue().getFineDenominator());
+  const int roundedNow = static_cast<const int>(getModulationAmount() * denominator);
+  const int roundedOG = static_cast<const int>(original->getRecallModulationAmount() * denominator);
+  return roundedOG != roundedNow;
 }
 
 bool ModulateableParameter::isModSourceChanged() const
 {
-  if(auto original = getOriginalParameter())
-  {
-    return original->getRecallModSource() != getModulationSource();
-  }
-  return false;
+  auto original = getOriginalParameter();
+  return original->getRecallModSource() != getModulationSource();
 }
 
 bool ModulateableParameter::isMacroControlAssignedAndChanged() const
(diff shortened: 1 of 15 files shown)


MatthiasSeeber (commented):
- is there a reason that the double click only works when clicking in the main area? (in the parameter tab it's the "classical" default)
- I see a potential conceptual issue with this mechanism: this only works as intended if the user didn't change the default value (store init, now in the setup menu > device settings)

projects/NonMaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/parameters/BeltParameterLayout.java
```diff
@@ -493,7 +494,8 @@ public void onMouseLost() {
 
 	@Override
 	public Control doubleClick(Position pos) {
-		startEdit().setToDefault();
+		final ParameterPresenter p = EditBufferPresenterProvider.getPresenter().selectedParameter;
```
hhoegelo, line 497: Please use the 'UseCase' indirection

MatthiasSeeber (approved):

projects/NonMaps/src/main/java/com/nonlinearlabs/client/useCases/EditBufferUseCases.java
```diff
@@ -205,17 +206,11 @@ public void convertToLayerSound() {
...
-		}
+		NonMaps.get().getServerProxy().defaultParameter(id);
 	}
 
 	public void setToDefault(int paramNumber) {
-		ParameterId id = toParamId(paramNumber);
-		BasicParameterModel p = EditBufferModel.get().getParameter(id);
-		double v = p.value.metaData.defaultValue.getValue();
-		setParameterValue(id, v, true);
+		NonMaps.get().getServerProxy().defaultParameter(paramNumber);
```
hhoegelo, line 213: Consider 

setToDefault(new ParameterId(paramNumber, EditBufferModel.get().voiceGroup.getValue())

here and removing the ServerProxy::defaultParameter(int paramNumber) instead.

hhoegelo: Id this method used at all?

projects/playground/src/presets/EditBufferActions.cpp
```diff
@@ -307,6 +307,14 @@ EditBufferActions::EditBufferActions(EditBuffer* editBuffer)
     auto scope = eb->getUndoScope().startTransaction("Select Part " + to_string(part));
     hwui->setCurrentVoiceGroupAndUpdateParameterSelection(scope->getTransaction(), part);
   });
+
+  addAction("default-param", [=](auto request) {
+    auto id = ParameterId(request->get("id"));
+    if(auto p = Application::get().getPresetManager()->getEditBuffer()->findParameterByID(id))
+    {
+      p->setDefaultFromHwui();
```
hhoegelo, line 315: This should actually be setDefaultFromWebUI. At least it should not say "fromHwui"

hhoegelo (approved):

hhoegelo (approved):

PR #2168, opened 2020-09-03
when adding a new preset, preset manager tried to scroll to it, even …

hhoegelo:
…though there is not even a position calculated


schlenkibus:
@hhoegelo was there a ticket for this PR?

PR #2171, opened 2020-09-08
cleanup upper/lower-mod-limit undo transactions

schlenkibus:
move limit change into mod-parameter, webui and hwui upper/lower limit changes are now more similar -> number of transactions / undo"

### DIFF
--- a/projects/playground/src/presets/EditBufferActions.cpp
+++ b/projects/playground/src/presets/EditBufferActions.cpp
@@ -194,6 +194,34 @@ EditBufferActions::EditBufferActions(EditBuffer* editBuffer)
     }
   });
 
+  addAction("set-lower-modulation", [=](std::shared_ptr<NetworkRequest> request) mutable {
+    auto id = ParameterId { request->get("id") };
+    auto newAmt = std::stod(request->get("mod-amt"));
+    auto newParamVal = std::stod(request->get("param-val"));
+    if(auto param = dynamic_cast<ModulateableParameter*>(editBuffer->findParameterByID(id)))
+    {
+      auto scope = param->getUndoScope().startContinuousTransaction(param, "Set Lower Modulation Limit for '%0'",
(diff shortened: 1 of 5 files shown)


MatthiasSeeber (approved):

projects/playground/src/presets/EditBufferActions.cpp
```diff
@@ -194,6 +194,34 @@ EditBufferActions::EditBufferActions(EditBuffer* editBuffer)
     }
   });
 
+  addAction("set-lower-modulation", [=](std::shared_ptr<NetworkRequest> request) mutable {
```
hhoegelo, line 197: If it is really only the undo string that differs for set-lower-modulation and set-upper-modulation, I'd suggest to rate lower code complexity higher than fancy undo texts and merge both in one function "Set modulation Limit for XYZ".
WebUI can simplified the same way.

projects/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/LowerModulationBoundControl.cpp
```diff
@@ -71,7 +71,7 @@ bool LowerModulationBoundControl::onRotary(int inc, ButtonModifiers modifiers)
         newModAmount /= 2;
 
       auto &undoScope = modulatedParam->getUndoScope();
-      auto scope = undoScope.startContinuousTransaction(modulatedParam, "Set '%0'",
+      auto scope = undoScope.startContinuousTransaction(modulatedParam, "Set Lower Modulation Limit for '%0'",
```
schlenkibus, line 74: change here aswell

MatthiasSeeber (approved):
The undo history entries are now consistently named "set modulation limit for..." on both WebUI and BOLED.

hhoegelo (approved):

Issue #2198, opened 2020-09-17
HWUI: MC assignment via quick buttons (button handling)


majjam:
Version information:
![grafik](https://user-images.githubusercontent.com/4758563/93503056-dcab4f80-f917-11ea-86be-ac5784554e0f.png)

When assigning a MC to parameter by pressing MC(A-F)-button and a parameter button at once,boled shows screen depending on last released button. Is this intended ?

Example 1
Press and hold Button MC-C
Press and hold SVF-Resonance button
Release SVF-Res
Release MC-C
Boled result is :
![grafik](https://user-images.githubusercontent.com/4758563/93503617-9e626000-f918-11ea-97d2-31b3db7583ca.png)

Example 2
Press and hold Button MC-C
Press and hold SVF-Resonance button
Release MC-C
Release SVF-Res
Boled result is :
![grafik](https://user-images.githubusercontent.com/4758563/93503715-c9e54a80-f918-11ea-9396-61d2011f52f5.png)

On older revisions I ended up always in the parameter screen independently from the button combo.
Personally i prefer the 'old' behaviour

PR #2199, opened 2020-09-22
Issue 2198

schlenkibus:
with the introduction of an delayed job to install the layout, subsequent checks for layout type after `setFocusAndMode`, `undoableSelectParameter` and similar have to be passed into `onLayoutInstalled`, as upon returning of `setFocusAndMode` the layout must not be installed yet

### DIFF
--- a/projects/epc/playground/src/parameters/Parameter.cpp
+++ b/projects/epc/playground/src/parameters/Parameter.cpp
@@ -170,12 +170,19 @@ bool Parameter::isDisabledForType(SoundType type) const
     case SoundType::Invalid:
       return true;
   }
+  return false;
 }
 
 void Parameter::loadDefault(UNDO::Transaction *transaction, Defaults mode)
 {
-  if(!isDisabled() && !isLocked())
-    loadFromPreset(transaction, mode == Defaults::UserDefault ? getDefaultValue() : getFactoryDefaultValue());
+  if(mode == Defaults::FactoryDefault)
+  {
+    loadFromPreset(transaction, getFactoryDefaultValue());
+  }
+  else if(!isLocked())
+  {
+    loadFromPreset(transaction, getDefaultValue());
+  }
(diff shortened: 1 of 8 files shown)


projects/epc/playground/src/parameters/Parameter.cpp
```diff
@@ -170,12 +170,19 @@ bool Parameter::isDisabledForType(SoundType type) const
...
-  if(!isDisabled() && !isLocked())
-    loadFromPreset(transaction, mode == Defaults::UserDefault ? getDefaultValue() : getFactoryDefaultValue());
+  if(mode == Defaults::FactoryDefault)
+  {
+    loadFromPreset(transaction, getFactoryDefaultValue());
+  }
+  else if(!isLocked())
+  {
+    loadFromPreset(transaction, getDefaultValue());
+  }
```
MatthiasSeeber, line 185: I don't quite get the relation to #2198 here...

And to me, the former condition included the !isLocked() also for the upper conditional block. Is it intentional to now not check for the locked flag when loading FactoryDefault?

MatthiasSeeber (approved):
Functionality-wise, the issue seems fixed: the final BOLED screen is the MC-Amt Soft Button, regardless of the order of releasing the Param / Macro Selection Buttons.

projects/epc/playground/src/parameters/Parameter.cpp
schlenkibus: Yes, this change was needed to pass the tests again, as loadDefault (with mode == FactoryDefault) is called inside the convert routines of the EditBuffer. When converting Single to Split or similar. The prior existing isDisabled - check (parameters that are not present in different sound types eg: SplitPoint in Single and Layer) was preventing some parameters to return to the defaults.

hhoegelo (approved):

PR #2200, opened 2020-09-22
Make Split Point Dual

schlenkibus:
started working on splitting split point into voice groups, making it dual, no more crashes for now

TODO:
- [x] convert Presets?
- [x] revisit compare
- [x] add overlap feature
- [x] fix tests
- [x] Load To Part should ignore Split Parameter -> never load split point
- [x] Convert Split To Layer should set Fade Parameters to Default
- [x] Sync Split Point Indicator in Split Sound Layout
- [x] convert layer to split should set split point I to mean of fade from and split II to mean of fade from + 1 step
- [x] sync split parameter toggle via webui context menu on split param
- [x] sync split parameter indicator to webui group header
- [x] split point aspects should always be mirrored to sibling parameter 
- [x] overlap split inidcator for webui
- [x] split sound init should initialize splitpoints without overlap
- [x] webui display value for split II is incorrect, always 1 step apart
- [x] increase bank version 
- [x] change split parameter range 0..60
- [x] prevent extremes for part I and II if sync active
- [x] parameter conversion to version 9
- [x] webui impl of edge cases
- [x] webui impl value changer split point in sound layout
- [x] cleanup undo transaction counts


projects/epc/playground/src/device-settings/SplitPointSyncParameters.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+Glib::ustring SplitPointSyncParameters::save() const
+{
+  return m_state ? "on" : "off";
+}
+
+Glib::ustring SplitPointSyncParameters::getDisplayString() const
+{
+  return m_state ? "Active" : "Disabled";
```
hhoegelo, line 22: Consider "Enabled/Disabled" or "Active/Inactive" or "On/Off"

projects/epc/playground/src/device-settings/SplitPointSyncParameters.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+bool SplitPointSyncParameters::persistent() const
+{
+  return false;
```
hhoegelo, line 27: The setting is reset to 'true' on each reboot by intention?

hhoegelo: Is it really a 'setting' then?

projects/epc/playground/src/presets/EditBuffer.cpp
```diff
@@ -1380,7 +1365,10 @@ void EditBuffer::calculateSplitPointFromFadeParams(UNDO::Transaction *transactio
   const auto f2 = findParameterByID({ C15::PID::Voice_Grp_Fade_From, VoiceGroup::II })->getControlPositionValue();
   const auto meanFadeFrom = (f1 + f2) / 2.0;
 
-  getSplitPoint()->setCPFromHwui(transaction, meanFadeFrom);
+  for(auto vg : { VoiceGroup::I, VoiceGroup::II })
+  {
+    findParameterByID({ C15::PID::Split_Split_Point, vg })->setCPFromHwui(transaction, meanFadeFrom);
```
hhoegelo, line 1370: Both have the same value? I would have expected the split(VG-II) = split(VG-I) + 1.

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -284,9 +293,12 @@ nltools::msg::SplitPresetMessage AudioEngineProxy::createSplitEditBufferMessage(
...
   fillDualMessage(msg, eb);
 
-  if(auto sp = eb.getSplitPoint())
+  for(auto i = 0; i < 2; i++)
   {
-    auto &t = msg.splitpoint;
+    auto sp = dynamic_cast<ModulateableParameter *>(
+        eb.findParameterByID({ C15::PID::Split_Split_Point, i == 0 ? VoiceGroup::I : VoiceGroup::II }));
+
+    auto &t = msg.splitpoint[1];
```
hhoegelo, line 301: did you mean msg.splitpoint[i] here? If not, 'i' is not used as int, so you could as well loop over {VoiceGroup.I, VoiceGroup.II}.

projects/epc/playground/src/proxies/hwui/descriptive-layouts/events/event-sources/edit-buffer/EditBufferUIEvents.h
```diff
@@ -151,9 +151,9 @@ namespace DescriptiveLayouts
     void onChange(const EditBuffer *eb) override
     {
       auto vg = Application::get().getHWUI()->getCurrentVoiceGroup();
-      if(auto splitPoint = eb->getSplitPoint())
```
hhoegelo, line 154: Do we still need SplitPointValueText at all? Isn't it a normal parameter now?

schlenkibus: It is a normal Parameter, but we still need the SplitPointValueText event source to be consumed by SplitPointLayout DynamicLayout.

projects/epc/playground/src/parameters/SplitPointParameter.cpp
```diff
@@ -7,35 +7,47 @@
...
+#include <device-settings/Settings.h>
+#include <device-settings/SplitPointSyncParameters.h>
+#include <parameter_declarations.h>
 
-SplitPointParameter::SplitPointParameter(ParameterGroup *group, const ParameterId &id)
+SplitPointParameter::SplitPointParameter(ParameterGroup* group, const ParameterId& id)
     : ModulateableParameterWithUnusualModUnit(group, id, ScaleConverter::get<SplitPointScaleConverter>(),
                                               ScaleConverter::get<LinearBipolar59StScaleConverter>(), 0.5, 59, 59)
 {
+  m_syncSetting = Application::get().getSettings()->getSetting<SplitPointSyncParameters>().get();
```
hhoegelo, line 20: I'd prefer to get the setting every time we need it instead of carrying a pointer with us.

projects/epc/playground/src/parameters/SplitPointParameter.cpp
```diff
@@ -7,35 +7,47 @@
...
+  if(m_syncSetting->getState() && initiator != Initiator::INDIRECT_SPLIT_SYNC)
+  {
+    getSibling()->setCpValue(transaction, Initiator::INDIRECT_SPLIT_SYNC, value, dosendToPlaycontroller);
+  }
 }
 
 Glib::ustring SplitPointParameter::getDisplayString() const
 {
-  return getDisplayString(Application::get().getHWUI()->getCurrentVoiceGroup());
+  return SplitPointDimension::stringizeSplitPointDisplay(getValue().getRawValue(), getVoiceGroup());
```
hhoegelo, line 50: Maybe I am wrong, but now that we have two distinct split point params, cannot we get rid of this awkward 'behind-the-scenes" routing of display stingizers and instead have 2 separate Stringizers, one for SplitParam in VG1 and one for SplitParam in VG2?

projects/epc/playground/src/parameters/SplitPointParameter.cpp
```diff
@@ -7,54 +7,100 @@
...
-Glib::ustring SplitPointParameter::getDisplayString(VoiceGroup vg) const
+void SplitPointParameter::setCpValue(UNDO::Transaction* transaction, Initiator initiator, tControlPositionValue value,
+                                     bool dosendToPlaycontroller)
 {
-  return SplitPointDimension::stringizeSplitPointDisplay(getValue().getRawValue(), vg);
+  Parameter::setCpValue(transaction, initiator, value, dosendToPlaycontroller);
+  auto other = getSibling();
+  tControlPositionValue siblingValue = other->getControlPositionValue();
+
+  if(Application::get().getSettings()->getSetting<SplitPointSyncParameters>().get()->get())
```
hhoegelo, line 43: .get()->get() can be replaced with ->get()

MatthiasSeeber (approved):
- [x] unlinked Split Point covers whole Key range (61 keys), linked Split Point covers range as previously established (60 keys)
- [x] bank format is now version 9 and seems to convert from v8 correctly
- [x] dual split point can be modulated (any assignment will always affect both parts), modulation tests showed the system working for linked and unlinked, overlapping split points
- [x] hwui and webui editing of split point seems smooth and correct (including Sound Belt / Screen)
- [x] split points are correctly recalled when loading presets
- [x] audio engine seems to handle new split mechanics as intended

schlenkibus:
@MatthiasSeeber can we merge this PR regardless of the new Symbols? I would need this PR to continue working on the fade editor (split). If we merge this, i would extract the missing graphics overhaul into a new ticket

hhoegelo (approved):

PR #2213, opened 2020-09-29
fix crash, closes #2212

### LINKED ISSUE #2212: [Playground] Crash when setting MC Sel to [-]
Issue:
If MC Sel is decreased from A to [-] Playground crashes and restarts.
(A Parameter which has no MC connected shows [-] without any problems)

Version:
grafik

Example setting :
grafik
Press Dec(-) button and Playgound crashes

schlenkibus:


MatthiasSeeber (approved):
The crash has vanished, the MC Selector can again be cycled in both directions.

PR #2217, opened 2020-09-30
added toFX and FB Layer arrows in webui, closes #2202

### LINKED ISSUE #2202: WebUI: Add To FX and FB Indicators to Sound Belt
Mockups:
image
image
image
image

TODO:
- [ ] Get SVGs

schlenkibus:


MatthiasSeeber (changes requested):
- [ ] the arrows only show up in a Layer sound, not in a Split Sound
- [ ] the Boled Sound Screen always shows a Single Sound
- [ ] in a Layer Sound, the Arrows for Cross FB and Cross Send (To FX) seem to be good, though

MatthiasSeeber (approved):
Arrows point in the right direction, both Split and Layer Sound Boled Screens and WebUI Belts are okay. 

Just a question: We curve the arrows up and down in the Split Screen / Belt, although we technically would no need to (as only cross FB, not self FB is inidcated) - meaning that the arrows could simply point horizontally. On the other hand, the appearance remains comparable to the Layer indication. So I'm just asking.

boled-22915938147793

MatthiasSeeber:
> Arrows point in the right direction, both Split and Layer Sound Boled Screens and WebUI Belts are okay.
> 
> Just a question: We curve the arrows up and down in the Split Screen / Belt, although we technically would no need to (as only cross FB, not self FB is inidcated) - meaning that the arrows could simply point horizontally. On the other hand, the appearance remains comparable to the Layer indication. So I'm just asking.
> 
> boled-22915938147793

Curved arrows are fine.

hhoegelo (approved):

PR #2222, opened 2020-10-01
EditBuffer is now also changed if part label was changed, closes #2219

### LINKED ISSUE #2219: [Split, Layer] changing Part Labels
- changed Part Labels should cause a * to be shown for the preset, indicating that this change has to be stored (which it certainly does)
- in the WebUI, there seems to be no way of changing the part label (which could be realized by a context menu on the labels, or maybe even within the Preset Info window?)

schlenkibus:


MatthiasSeeber (changes requested):
- [ ] The star does not appear to be reflected in the WebUI. The Boled is okay.
- [x] The Ticket also mentions the need for a Part Label Rename mechanism in the WebUI --> separate ticket?

projects/epc/playground/src/presets/EditBuffer.cpp
```diff
@@ -1643,3 +1646,15 @@ void EditBuffer::fakeParameterSelectionSignal(VoiceGroup oldGroup, VoiceGroup gr
...
+bool EditBuffer::isPartLabelChanged(VoiceGroup group) const
+{
+  if(auto preset = getOrigin())
+  {
+    if(preset->isDual())
+    {
+      return preset->getVoiceGroupName(group) != getVoiceGroupName(group);
+    }
+  }
+  return false;
```
hhoegelo, line 1659: What, if the original preset was a single sound, than the edit buffer was converted to a dual sound, than the part name was changed? Is this a relevant use case?

hhoegelo: I agree, that the problem I commented on has no effect for the current users of the method. 

Still it is a problem if we have a method where the name does not reflect the implementation. So either we should rename or at least comment the method to indicate to future users the lack of correctness.

schlenkibus: I removed the assumption about the source preset, and the labels are compared without checking if the source preset is dual. The current users of this function are already checking for sound types. IMO this function now does what it states it does.

MatthiasSeeber (approved):

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/dataModel/DataModelEntity.java
```diff
@@ -8,14 +8,20 @@ public DataModelEntity(T def) {
...
+		if (!v.equals(value) || forceNotify) {
 			value = v;
 			notifyChanges();
 			return true;
 		}
 		return false;
 	}
 
+	public void notifyOnEachSet(boolean n) {
+		forceNotify = n;
```
hhoegelo, line 20: This doesn't look necessary to me - let's talk about it later today.

hhoegelo (approved):

PR #2244, opened 2020-10-13
Issue 2238

### LINKED ISSUE #2238: WebUI: Fade Point/Fade Range Editor
Layer-Split-Editor-test-b
Fade-Editor-Btn-test-a

schlenkibus:
closes #2238


MatthiasSeeber (changes requested):
Since the Editor feature will be subject of a forthcoming PR, the following things may be included there. So far I found the following.

**Sound Check**
No issues found - seems to work properly, incl. Modulation of Split Point.

**Display Issues**
- [ ] the Icon for the Fade Editor moves slightly to the left when clicking on it, I think it should stay where it is?
- [ ] when I use a window width of around 960px, the only visible part of the editor is the piano, values and icons are not visible anymore (as the piano doesn't resize)
- [ ] [Layer] the Fade Range values round the decimal away (see/compare to corresponding Boled)
- [ ] [Layer] the colored areas seem wrong to me. If I set Fade From [I] to C1, C1 should be behind the blue area. If I set Fade From[II] to C6, C6 should be behind the orange area.The shown value of a Fade From indicates the last key on full level, before the fade begins, it has to be included.

**Other Issues**
- [ ] dragging on value fields is really odd. sometimes something happens, but either the "anfasser" focus changes (sometimes not focussing on the value I drag on), or the value simply jumps to minimum/maximum. changing a particular value item smoothly was not possible for me. if drag on values is on the to-implement checklist, maybe also include double click for default - if not, all mouse events on values should be disabled.
- [ ] [Layer] dragging the "anfasser" of a Fade Range can cause jumping behavior (maybe due to rounding of decimals?)
- [ ] consider having a Split or Layer sound and opening the Fade editor. then load a Single preset. I think, in this case the editor should disappear and show the sound tab instead. (For Split and Layer sounds, the editor can stay open i.m.o.)

MatthiasSeeber (approved):

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/useCases/EditBufferUseCases.java
```diff
@@ -35,7 +32,10 @@ public void setParameterValue(ParameterId id, double newValue, boolean oracle) {
 	private void setParameterValue(ParameterId id, double newValue, boolean oracle, boolean setAnimationTimeout) {
 		BasicParameterModel p = EditBufferModel.get().getParameter(id);
 		double oldQ = p.value.getQuantizedAndClipped(true);
-		p.value.value.setValue(newValue);
+
+		double clippedValue = p.value.getQuantizedAndClipped(newValue, true);
+		p.value.value.setValue(clippedValue);
```
hhoegelo, line 37: I am not 100% that this is correct. Aren't there use cases where it is temporarily desired to accept values outside the range and in between the quantization steps? For example when using modulation? This has to be tested very carefully.

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/belt/fadeeditor/KeyBed.java
```diff
@@ -0,0 +1,394 @@
...
+import com.nonlinearlabs.client.world.Position;
+import com.nonlinearlabs.client.world.RGB;
+import com.nonlinearlabs.client.world.RGBA;
+import com.nonlinearlabs.client.world.Rect;
+import com.nonlinearlabs.client.world.overlay.DragProxy;
+import com.nonlinearlabs.client.world.overlay.SVGImage;
+
+public class KeyBed extends SVGImage {
+
+    enum SelectedAnfasser {
```
hhoegelo, line 23: Consider renaming to 'SelectedHandle'

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/useCases/EditBufferUseCases.java
schlenkibus: I removed this lines again, as i could not recall the reason for adding them in the first place.

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/belt/fadeeditor/KeyBed.java
schlenkibus: Done, all occurrences of Anfasser are now Handle :/

hhoegelo (approved):

PR #2247, opened 2020-10-15
Issue 1962

MatthiasSeeber:
see #1962 
- [x] fixes "Pitch Bug" of the Envelopes A,B,C (missing the Breakpoint occasionally when using 0 times for Attack, Decay1)
- [ ] consider fast clock or variable (?) clock for updating Pitches (Osc A/B, Comb, SVF)


MatthiasSeeber:
I'm testing and comparing Pitch via slow agains fast clock rates, with a simple impulse-like envelope c on osc pitch, with bender-variable decay2 time - preset (3) of the attached bank.

- for really small decay times, the effect becomes almost unnoticeable, regardless of the rate
- the fast rate seems to use slightly more cpu (as expected), but only slightly improves a "really fast" env c on an osc pitch (as far as I can hear it)

[issue 1962.zip](https://github.com/nonlinear-labs-dev/C15/files/5385876/issue.1962.zip)

PR #2248, opened 2020-10-16
AudioEngine - Envelope fixes and potential improvements

### LINKED ISSUE #1721: AudioEngine - (Layer) Part Fades vs Mono (or long release times)
When in a Mono Layer Sound, the current implementation of applying Fade levels for a particular Voice (updated once on KeyDown - no smoothing) produces unwanted audible clicks.
_(This should also occur, when the Release Time of a Note is long enough to provoke this)_

**Possible solutions:**
- in Mono Mode, the Glide could cover this _(still offering no solution for polyphonic cases)_
- the application of Fade levels could be smoothed by a "Fade Time" parameter _(covering both mono and poly cases, but I am not entirely convinced that this could be benefitial?)_

@nonlinear-labs what do you think?

### LINKED ISSUE #1962: Reconsider Envelope Sequencers and Env C Rate
Weitere Untersuchungen und Beispiel-Presets folgen...

### LINKED ISSUE #2216: AudioEngine - Possible Optimizations
- [x] Output Mixer: simplify monophonic HP (by exploiting the static HP implementation of the Shaper Asym)

MatthiasSeeber:
replaces #2247 (now coming from and aiming into `staging`)

the bugfix and potential improvements can be tested with the following bank:
[PR 2248.zip](https://github.com/nonlinear-labs-dev/C15/files/5408440/PR.2248.zip)

fixes #1962 
- [x] primary bug fix: envelopes a, b, c with very short attack, decay1 times missed the Breakpoint occasionally (testable with Preset 1)
- [x] potential improvement: process pitch-related signals for osc a/b, comb and svf on fast clock (so at 9600 Hz instead of the slow 400 Hz) (testable with Presets 3 (disabled) and 4 (enabled))

fixes #2216
- [x] potential improvement: replace output mixer mono hp by simpler implementation (inspired from the neighboring shaper asym hp filter)

not related to a ticket, but in correspondence with @nonlinear-labs 
- [x] integrated an immediate env c effect on pitches: if the attack time is sufficiently close to zero (smaller than the first nonzero fine step, I thought), the envelope peak level will be used for processing the pitch during a key-event correctly (at this time, the envelope signal is not yet rendered and will be at peak level after rendering anyway - this just ensures that pitches reflect immediate env c effects on them accordingly)

fixes #1721
- [x] implement legato rule (no new voice fade level)
- [x] potential improvement: provide first iteration of lookup-based smoothing as a proof of concept - curve: sine, time: 1ms, affects currently: voice_level (layer sounds with fade ranges > 0) (testable with Preset 5)

(Potential Improvements work via defines in `ae_potential_improvements.h` - all mentioned potential improvements are `disabled` by default, but can be selectively enabled for test builds at any time. Once they are considered "permanent", a cleanup can be done)


schlenkibus (approved):

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.cpp
```diff
@@ -107,6 +120,26 @@ void PolySection::render_audio(const float _mute)
   m_svfilter.apply(m_signals, m_soundgenerator.m_out_A, m_soundgenerator.m_out_B, m_combfilter.m_out);
   m_outputmixer.combine(m_signals, m_voice_level, m_soundgenerator.m_out_A, m_soundgenerator.m_out_B,
                         m_combfilter.m_out, m_svfilter.m_out);
+#elif POTENTIAL_IMPROVEMENT_GAIN_CURVE == __POTENTIAL_IMPROVEMENT_ENABLED__
+  PolyValue tmp_voice_level;
+  for(uint32_t v = 0; v < m_voices; v++)
+  {
+    postProcess_poly_audio(v, _mute);
+    const uint32_t v_index = m_gain_curve_index[v];
```
hhoegelo, line 128: It should be possible to parallelize this loop:
If it is split into one loop calling  postProcess_poly_audio(v, _mute) per voice, and another loop processing the voice level. As far as I can see, all computations depend on 'v' only and loop cycles do not interfere. The Ln 131-134 should probably replaced with a unconditional "m_gain_curve_index[v] = std::min(m_gain_curve_size_m1, v_index + 1)" so the branching can be removed.

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.h
```diff
@@ -79,7 +79,7 @@ class PolySection
   NlToolbox::Curves::Shaper_2_BP m_svf_resCurve;
   ProtoSmoother<1> m_mono_glide;
   PolyValue m_comb_decay_times[2] = {}, m_voice_level = {};
-#if POTENTIAL_IMPROVEMENT_GAIN_CURVE == __POTENTIAL_IMPROVEMENT_DISABLED__
+#if POTENTIAL_IMPROVEMENT_GAIN_CURVE == __POTENTIAL_IMPROVEMENT_ENABLED__
```
schlenkibus, line 82: Usage of constexpr if could have prevented compile errors we encountered. Like this:

//inside ae_potentialimprovements.h:
#if POTENTIAL_IMPROVEMENT_GAIN_CURVE == POTENTIAL_IMPROVEMENT_ENABLED
constexpr bool POTENTIAL_IMPROVEMENT_GAIN_CURVE_BOOL = true;
#else
constexpr bool POTENTIAL_IMPROVEMENT_GAIN_CURVE_BOOL = false;
#endif

//usage
if constexpr(POTENTIAL_IMPROVEMENT_GAIN_CURVE_BOOL) {

} else if constexpr(!POTENTIAL_IMPROVEMENT_GAIN_CURVE_BOOL) {

}

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.cpp
MatthiasSeeber: Each Voice has/can have a different index, all read from one monophonic lookup table. In the Comb Filter, we could also not parallelize Buffer Reading (with polyphonic indices on polyphonic buffer) - the only thing not parallelizable. So, I don't really see how to achieve this.

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.h
MatthiasSeeber: Yes, constepxr..
Still, my QT highlights the syntax `if constexpr(...)` as an error. I can build still using that, but some files will "drown" in those error messages, which I find inconvenient.
I also would have thought to either use a define or a bool constexpr, not both.
The Defines are thought to disable whole sections if a feature is disabled, thus hiding any associated data. Here, the feature's content was wrongfully present when disabled instead of enabled, producing the error.

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.cpp
hhoegelo: ` 
for(uint32_t v = 0; v < m_voices; v++)
    postProcess_poly_audio(v, _mute);

for(uint32_t v = 0; v < m_voices; v++)
{
    const uint32_t v_index = m_gain_curve_index[v];
    const float v_fade = m_gain_curve_buffer[v_index];
    tmp_voice_level[v] = ((1.0f - v_fade) * m_voice_level_old[v]) + (v_fade * m_voice_level[v]);
    m_gain_curve_index[v] = std::min(m_gain_curve_size_m1, v_index + 1);
}
`

If written like this, I still cannot see why the compiler should not be able to SIMD the loop.

MatthiasSeeber: Okay - I thought I would need to implement something into ParallelData.h in order to achieve this.

projects/epc/audio-engine/src/synth/c15-audio-engine/ae_poly_section.h
hhoegelo: I don't think, the ```if constexpr ()``` makes a difference.

```
constexpr bool enableXYZ = true;
if (enableXYZ) {}
else {}
```

should yield the same assembly, as the compiler can still deduce the obvious impossibility of the second branch at compile time.

So I'd vote for this style as it still solves the problem and also works with @MatthiasSeeber IDE.

schlenkibus: I was just suggesting to use if constexpr as yesterday we had compile errors with changed defines. 
I agree that working with a lot of noise in the IDE is not nice, so no strong opinion here

hhoegelo (approved):

hhoegelo:
@schlenkibus Could you please review this PR so it can be merged?

schlenkibus (approved):

PR #2256, opened 2020-10-28
issue-2253

### LINKED ISSUE #2253: [Store] triggers PresetMessage(s) ?
Preset Messages are sent to the AudioEngine while storing:
- observed in `staging` via PR #2248 and `unstable` via https://github.com/nonlinear-labs-dev/C15/commit/73fa2d846183fefa8bda46959008ebdf2c85604b
- observed in Boled and WebUI when appending a Preset (doesn't seem to occur on overwrite)
- observed one Preset Message per trigger for Boled, two Preset Messages per trigger for WebUI

This can result in stopping a played Note, especially when the `Glitch Suppression` setting is enabled. There really is no need to send Preset Messages while storing anyway (the AudioEngine should be up-to-date at this point), so we may want to consider getting rid of this redundancy?

schlenkibus:
added check if preset message changed before sending same message again. closes #2253

- [x] todo test


MatthiasSeeber (changes requested):
> Preset Messages müssen versendet werden, wenn: der User ein Preset oder einen Part lädt (Recall) oder einen Sound konvertiert/randomisiert/initialisiert.
> Sie müssen/sollten nicht versendet werden, wenn ein Sound als Preset abgespeichert wird (Store append/overwrite/insert).

hhoegelo (approved):

schlenkibus:
I extracted the use cases for Store Actions from all over the codebase into one central location, now store actions called from the webui and the boled will run the same code. The original issue of sending preset messages when stored is now fixed. Every store will update the selection and loaded preset status to the just inserted, appended preset.

projects/epc/playground/src/presets/Bank.cpp
```diff
@@ -435,7 +436,7 @@ Preset *Bank::prependPreset(UNDO::Transaction *transaction, std::unique_ptr<Pres
 Preset *Bank::prependAndLoadPreset(UNDO::Transaction *transaction, std::unique_ptr<Preset> preset)
 {
   auto newPreset = prependPreset(transaction, std::move(preset));
-  getEditBuffer()->undoableLoad(transaction, newPreset);
+  getEditBuffer()->undoableLoad(transaction, newPreset, false);
```
MatthiasSeeber, line 439: I noticed that in [426, 453], the `getPresetManager()->onPresetStored(newPreset);` is invoked. Why isn't it called here?

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.h
```diff
@@ -31,24 +31,27 @@ class AudioEngineProxy
...
+    send(sendToEndPoint, msg);
   }
 
   void sendEditBuffer();
 
+  void setReceiver(nltools::msg::EndPoint endPoint);
+
   void freezeParameterMessages();
-  void thawParameterMessages();
+  void thawParameterMessages(bool send);
```
MatthiasSeeber, line 47: The name is rather confusing to me. What does `thaw` mean?

MatthiasSeeber (approved):
I observed that Presets are only sent to the AudioEngine only when needed - all Store process are "mute" now. Out of curiosity, I have two additional questions.

projects/epc/playground/src/presets/Bank.cpp
hhoegelo: Good question, I'd like to know that, too. @schlenkibus ?

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.h
hhoegelo: It is just the opposite of 'freeze' - maybe 'pause' and 'resume' would be more appropriate words. 
Pausing the hundreds of parameter messages makes sense, if we know that eventually there will be a full preset message anyway. 
The wording is borrowed from glib: https://developer.gnome.org/glibmm/stable/classGlib_1_1ObjectBase.html#a1bd8ea7bd8c4084ade6b3c27dddf06a4

projects/epc/playground/src/presets/PresetManager.cpp
```diff
@@ -999,3 +999,13 @@ void PresetManager::autoLoadPresetAccordingToLoadType()
     }
   }
 }
+
+void PresetManager::onPresetStored(const Preset *storedPreset)
+{
+  m_presetStoreHappened.deferedSend(storedPreset);
```
hhoegelo, line 1005: It is not actually a problem, because due to our undo logic a preset will 'never' be deleted, but I don't like the idea of deferred signaling of an object by pointer. Theoretically it is possible that the preset is already deleted at the time the signal notification eventually happens. Also, the - so far - one and only user of the signal doesn't care about the preset pointer. So I suggest to remove the pointer and just signal that "some preset store has happened".

schlenkibus: makes sense, i will implement your suggestion

hhoegelo (approved):

PR #2258, opened 2020-10-28
Issue 2246

hhoegelo:
The midi-over-ip binary on the bbb is started as soon as the systemd sound.target is reached - this is when a midi (or audio-) device is connected. It will open the first available midi device and send all its messages to the endpoint "ExternalMidiOverIPClient". This endpoint is provided by the audio-engine, which will interlace the received events in its midi data stream. Also, the audio engine will send a "ACK" package immediately, so the midi-over-ip tool can measure the round trip time. It will output min and max when it is shut down (i.e. when device is disconnected and systemd sound.target is not reached anymore).

@nonlinear-labs On my C15 I measured a round trip time of 1100-6000 us. During measuring, I reloaded the WebUI (with lots of preset banks) several times and so caused heavy traffic in the usb. I find this acceptable - how about you?


schlenkibus:
Looks good to me so far, I am building this PR and will test it on the device.

schlenkibus:
From my simple tests i can assure that hot plugging and unplugging of an midi keyboard works without hiccups. The only problem i could identify was pressed notes were not released after when unplugging the keyboard. +approve from me

schlenkibus (approved):

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1242,6 +1256,14 @@ void dsp_host_dual::reset()
   }
 }
 
+dsp_host_dual::HWSourceValues dsp_host_dual::getHWSourceValues() const
+{
+  dsp_host_dual::HWSourceValues ret;
+  std::transform(m_params.m_global.m_source, m_params.m_global.m_source + ret.size(), ret.begin(),
+                 [](const auto& a) { return a.m_position; });
+  return ret;
+}
```
MatthiasSeeber, line 1265: Just to clarify: the values of `m_params.m_global.m_source` are not affected by the `std::transform`, right?

MatthiasSeeber (approved):
Via IDE over PC, I can successfully launch the AudioEngine the usual way and I can redirect MIDI to the midi-over-ip. (Haven't yet tested functionality on C15) Apart from my question (which may be obsolete), I also can approve this.

PR #2262, opened 2020-10-29
Midi WIP

hhoegelo:
- decouple outgoing external midi from audio thread
- add lsb for hardware soucres and velocity
- use CC 20..25 for hardware sources MSB and 52..57 LSB


hhoegelo:
@MatthiasSeeber Changing the range of CC was requested by @nonlinear-labs

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1447,14 +1490,77 @@ float dsp_host_dual::scale(const Scale_Aspect _scl, float _value)
...
+
+    case 4:  // Pitch
+    {
+      uint8_t statusByte = static_cast<uint8_t>(0xE0);
+      uint16_t v = static_cast<uint16_t>(value * (1 << 13));
+      uint8_t valByte1 = static_cast<uint8_t>(v & 0x7F);
+      uint8_t valByte2 = static_cast<uint8_t>((v >> 7) & 0x7F);
+      out({ statusByte, valByte1, valByte2 });
+      break;
+    }
```
MatthiasSeeber, line 1544: The "Pitch" is not a Hardware Source. This `update_hw` function triggers the whole Modulation Process. It should not be triggered by a Note On/Off message. I'd suggest to process the pitch in a separate function.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -527,48 +543,75 @@ void dsp_host_dual::onMidiMessage(const uint32_t _status, const uint32_t _data0,
       onTcdMessage(0xED, 0, _data0);  // keyPos
       if(arg != 0)
       {
-        onTcdMessage(0xEE, arg >> 7, arg & 127);  // keyDown
+        onTcdMessage(0xEE, arg >> 7, std::exchange(m_velocityLSB, 0));  // keyDown
```
MatthiasSeeber, line 546: I have trouble understanding this. Why isn't this needed in `[538] onTcdMessage(0xEF, arg >> 7, arg & 127);  // keyUp`?

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -474,6 +485,11 @@ void dsp_host_dual::onTcdMessage(const uint32_t _status, const uint32_t _data0,
           auto vel = static_cast<float>(arg) * m_norm_vel;
           keyDown(vel);
 
+          uint8_t highResolutionVelocityStatusByte = static_cast<uint8_t>(0xB0);
+          uint16_t fullResolutionValue = vel * (1 << 13);
```
MatthiasSeeber, line 489: Will the compiler treat `(1 << 13)` as a constexpr? (I so far preferred initializing a constant variable for this, avoding such shifts within a function)

MatthiasSeeber (commented):
Let's talk about this, and try to find the reason for the shifted pitches..

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -474,6 +485,11 @@ void dsp_host_dual::onTcdMessage(const uint32_t _status, const uint32_t _data0,
...
+          uint8_t lsbValByte = static_cast<uint8_t>(fullResolutionValue & 0x7F);
+          out({ highResolutionVelocityStatusByte, 88, lsbValByte });
+
           uint8_t statusByte = static_cast<uint8_t>(0x90);
           uint8_t keyByte = static_cast<uint8_t>(m_key_pos) & 0x7F;
```
MatthiasSeeber, line 494: I think, here arises the problem of the shifted pitch. In `[472] m_key_pos = arg - C15::Config::key_from;` is shifted to transform a "Note Pitch" (36..96) into a "Key Position" (0..60). Futher using the `m_key_pos` like a "Note Pitch" requires adding `C15::Config::key_from` again.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
hhoegelo: Yes, the expression will be replaced with a const literal value by the compiler.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
hhoegelo: Okay thanks - will fix it.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
hhoegelo: You are right - I will add it for note off.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
hhoegelo: It is just a wrong comment - it should read "Bender".
Would this fix the problem or do I still miss something?

schlenkibus (approved):

MatthiasSeeber (approved):

PR #2265, opened 2020-10-29
playground will send program changes to audio-engine, which will

hhoegelo:
dispatch it to midi-over-ip and the other way around

Sorry, I can't test the feature as my Roland FP-30 does not send or receive program changes.


MatthiasSeeber (commented):
I installed https://github.com/nonlinear-labs-dev/C15/tree/midi-over-ip-0.5 on my C15 and can verify the following:
- [x] hot plugging/unplugging an external MIDI device worked without trouble
- [x] sending (from ext. device): Notes, the right Control Changes (as 7-Bit), Pitch Bend and Aftertouch works
- [x] even sending program changes from the ext. device works as expected (affecting the currently selected bank)

Unfortunately, observing the C15 MIDI Output is rather difficult for me as well.

MatthiasSeeber (approved):

PR #2285, opened 2020-11-04
midi-over-ip ammendments

MatthiasSeeber:
- [x] added optional logging of MIDI Out messages (disabled by default)
- [x] fixed MIDI Out Program Change message
- [x] code simplification: introduced `C15::PID` to `AudioEngineProxy.cpp`
- [ ] found a bug in `AudioEngineProxy.cpp` and fixed it - verification needed

### DIFF
--- a/projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
+++ b/projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
@@ -25,9 +25,6 @@
 #include <groups/SplitParameterGroups.h>
 #include <proxies/playcontroller/PlaycontrollerProxy.h>
 
-constexpr auto cUnisonVoicesParameterNumber = 249;
-constexpr auto cMonoEnableParameterNumber = 364;
-
 AudioEngineProxy::AudioEngineProxy()
 {
   using namespace nltools::msg;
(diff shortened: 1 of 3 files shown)


projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -252,14 +249,16 @@ void AudioEngineProxy::fillUnisonPart(nltools::msg::ParameterGroups::UnisonGroup
...
+         { C15::PID::Unison_Pan, from }))  // previously was 252: Unison Phase (seems like a bugfix..?)
   {
     auto &pan = unisonGroup.pan;
     pan.id = unisonPan->getID().getNumber();
     pan.controlPosition = unisonPan->getControlPositionValue();
   }
 
-  if(auto unisonPhase = g->getParameterByID({ 253, from }))
+  if(auto unisonPhase = g->getParameterByID(
+         { C15::PID::Unison_Phase, from }))  // previously was 253: Unison Pan (seems like a bugfix..?)
```
MatthiasSeeber, line 261: Seems like the PG took Unison Pan for Unison Phase and vice verca - right?

hhoegelo (approved):

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
MatthiasSeeber: The "bugfix" seems to only affect Playground Unit Tests - explaining why the fix made no difference to `unstable` (when observing Preset and ParameterChanged Message transmissions).

PR #2288, opened 2020-11-05
Feature midi cc

### LINKED ISSUE #2283: MIDI Implementierung: Änderungen der HW-Source-Positionen auch von der UI über MIDI-Out senden

### LINKED ISSUE #2286: MIDI-Implementierung: Bender- und Ribbon-Positionen sind auf den Bereich 0...8192 (2^13) statt auf 2^14 skaliert.
Daher das Pitchbend-Problem: unser Mittelpunkt hängt bei -50 % (4096) statt bei 8192 (PB-Null).

Bei den CCs ist unser Range bis jetzt von MSB=0, LSB=0 bis MSB=64, LSB=0. Das Maximum sollte aber MSB=127, LSB=127 sein.

**Zum Mittelpunkts/Symmetrie-Problem:**

Für den 14-Bit-Pitchbender nimmt Reaper (höchstwahrscheinlich "normgerecht") MSB=64, LSB=0 als Mitte. Damit ist der Bereich nach oben MSB=63, 127 (8191), nach unten MSB=64, LSB=0 (8192). Unten müsste also ein LSB weggeclippt werden. Dann wäre bei MSB=0, LSB=1 (1) das Minimum erreicht.

Übertragen auf 7-Bit-Quellen, bedeutet das, dass es um 64 herum einen Bereich von +/-63 gibt. Wenn wir aber immer 14 Bit erwarten, fehlen oben und unten 127 LSB-Steps, sodass z.B. ein bipolarer Parameter nur von -63/64 bis +63/64 ausgesteuert würde. Statt -100 % ... +100 % würden nur -98,44 % ... 98,44 % erreicht.

Vorschlag: Wir clippen schon bei MSB=1, LSB=0 und MSB=127, LSB=0 und bilden diesen Bereich (126 * 128 = 16128 LSB-Steps) auf unsere 0...100 % bzw -100...100 % ab.

hhoegelo:
fixes #2283  
fixes #2286

### DIFF
--- a/projects/epc/audio-engine/src/synth/C15Synth.cpp
+++ b/projects/epc/audio-engine/src/synth/C15Synth.cpp
@@ -50,7 +50,7 @@ C15Synth::C15Synth(const AudioEngineOptions* options)
 
   // receive program changes from playground and dispatch it to midi-over-ip
   receive<nltools::msg::Midi::ProgramChangeMessage>(EndPoint::AudioEngine, [this](const auto& pc) {
-    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage { 0xC0, pc.program, 0 });
+    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage { 0, 0xC0, pc.program, 0 });
     m_externalMidiOutThreadWaiter.notifyUnlocked();
   });
 
@@ -93,12 +93,8 @@ void C15Synth::doMidi(const MidiEvent& event)
 
 void C15Synth::doTcd(const MidiEvent& event)
 {
-  m_dsp->onTcdMessage(event.raw[0], event.raw[1], event.raw[2], [=](auto outgoingMidiMessage) {
-    nltools::msg::Midi::SimpleMessage msg;
-    std::copy(outgoingMidiMessage.begin(), outgoingMidiMessage.end(), msg.rawBytes);
-    m_externalMidiOutBuffer.push(msg);
-    m_externalMidiOutThreadWaiter.notifyUnlocked();
-  });
+  m_dsp->onTcdMessage(event.raw[0], event.raw[1], event.raw[2],
+                      [=](auto outgoingMidiMessage) { queueExternalMidiOut(outgoingMidiMessage); });
 }
 
 void C15Synth::simulateKeyDown(int key)
@@ -316,11 +312,22 @@ void C15Synth::onHWSourceMessage(const nltools::msg::HWSourceChangedMessage& msg
   if(element.m_param.m_type == C15::Descriptors::ParameterType::Hardware_Source)
   {
     m_dsp->globalParChg(element.m_param.m_index, msg);
+    m_dsp->hwSourceToMidi(element.m_param.m_index, msg.controlPosition,
+                          [this](const auto& msg) { queueExternalMidiOut(msg); });
+
     return;
   }
 #if LOG_FAIL
   nltools::Log::warning("invalid HW_Source ID:", msg.parameterId);
 #endif
+}
+
+void C15Synth::queueExternalMidiOut(const dsp_host_dual::SimpleRawMidiMessage& m)
+{
+  nltools::msg::Midi::SimpleMessage msg;
+  std::copy(m.begin(), m.end(), msg.rawBytes);
+  m_externalMidiOutBuffer.push(msg);
+  m_externalMidiOutThreadWaiter.notifyUnlocked();
 }
 
 void C15Synth::onSplitPresetMessage(const nltools::msg::SplitPresetMessage& msg)
--- a/projects/epc/audio-engine/src/synth/C15Synth.cpp
+++ b/projects/epc/audio-engine/src/synth/C15Synth.cpp
@@ -50,7 +50,7 @@ C15Synth::C15Synth(const AudioEngineOptions* options)
 
   // receive program changes from playground and dispatch it to midi-over-ip
   receive<nltools::msg::Midi::ProgramChangeMessage>(EndPoint::AudioEngine, [this](const auto& pc) {
-    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage { 0, 0xC0, pc.program, 0 });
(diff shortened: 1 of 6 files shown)


MatthiasSeeber (commented):
I observer MIDI Out for changed HW Sources from Playground as anticipated.
However, when using the Pitch Bend on my external MIDI Controller, it now appears as a changing Ribbon 1.

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -50,7 +50,7 @@ C15Synth::C15Synth(const AudioEngineOptions* options)
 
   // receive program changes from playground and dispatch it to midi-over-ip
   receive<nltools::msg::Midi::ProgramChangeMessage>(EndPoint::AudioEngine, [this](const auto& pc) {
-    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage { 0xC0, pc.program, 0 });
+    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage{ 0, 0xC0, pc.program, 0 });
```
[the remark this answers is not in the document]
hhoegelo: well spotted - will fix it

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -22,11 +45,82 @@ template <int EnumValue> inline constexpr uint32_t getMidiCC()
...
 {
   m_hwSourcesMidiLSB.fill(0);
   m_mainOut_L = m_mainOut_R = 0.0f;
   m_layer_mode = C15::Properties::LayerMode::Single;
+
+  // pitch bend
+  assert(encode14BitBipolarMidiValue<Midi::BenderRange>(-1.5) == 0x0000);
+  assert(encode14BitBipolarMidiValue<Midi::BenderRange>(-1) == 0x0000);
+  assert(encode14BitBipolarMidiValue<Midi::BenderRange>(0) == 0x2000);
+  assert(encode14BitBipolarMidiValue<Midi::BenderRange>(1) == 0x3FFF);
```
MatthiasSeeber, line 96: Is it necessary to have/keep these assertions in the constructor?
(this seems like a Unit Test to me)

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1499,63 +1625,71 @@ float dsp_host_dual::scale(const Scale_Aspect _scl, float _value)
...
-  float value = _raw * m_norm_hw;
+  if(m_params.get_hw_src(id)->m_behavior == C15::Properties::HW_Return_Behavior::Center)
+    doSendCCOut<msb, lsb>(out, encode14BitBipolarMidiValue<Midi::BipolarCCRange>(controlPosition));
+  else
+    doSendCCOut<msb, lsb>(out, encode14BitUnipolarMidiValue<Midi::UnipolarCCRange>(controlPosition));
+}
+
+void dsp_host_dual::hwSourceToMidi(const uint32_t id, const float controlPosition, const MidiOut& out)
+{
+  auto bipolar = m_params.get_hw_src(id)->m_behavior == C15::Properties::HW_Return_Behavior::Center;
```
MatthiasSeeber, line 1650: This is an unused variable (the `sendCCOut` function does this check on its own anyway).

MatthiasSeeber (commented):
This appears to work quite well so far - some details may need further attention:

- [x] LSB and MSB Control Changes seem to work as expected
- [x] User Edits from the UI on HW Sources seem to work
- [ ] MSB-only Control Changes are fine for a bipolar HW Sources (the zero in the center can be reached), a unipolar HW Source on the other hand cannot be set to 50.0% (here I'd like to propose and discuss some slight changes for decoding and encoding)

- [ ] returning HW Sources (and their automatic return movements) can be inconvenient:
- consider having set a Source via MIDI to a certain position and then selecting it --> in the WebUI, the auto-return is triggered on selection / via HWUI the auto-return is triggered only when having changed the source and selecting something else (it may be useful to only let a Source return automatically when the user changed it via the UI)
- via the WebUI, the auto return will fire lots of MIDI CC Out, via HWUI one single CC Out is sent (which should be sufficient, as the Macro Controls have their individual smoothing times)

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
hhoegelo: Indeed, it is a unit test. I was just too lazy to add a proper unit test framework to the audio engine. I'd say, it doesn't hurt, but of course a dedicated unit test would be much better.

MatthiasSeeber:
so far functional:

- [x] Pedals and Ribbons use clipped 14 bit format (compatible with MSB/LSB) and produce values within expected ranges (0 ... 16256)
- [x] Aftertouch uses full-range 7 bit format and produces values within expected ranges (0 ...127)
- [x] Bender uses full-range 14 bit format and produces values within expected ranges (0 ... 16383)

remaining flaws:

- [ ] returning HW Sources
(see https://github.com/nonlinear-labs-dev/C15/pull/2288#pullrequestreview-526148146)
It would be nice when HWUI and WebUI operate in similar fashion: Only once tell the AudioEngine and MIDI Out that a HW Source returned to zero (like in HWUI), indicate the smoothing only in the WebUI (supressing further updates to AudioEngine and MIDI)
- [ ] program changes (see #2292)
Indeed, incoming MIDI program changes will produce MIDI Out messages. I couldn't find any flaw within the AudioEngine and suspect Direct Load -related misbehavior in the Playground. Further, multiple identical program change messages will be sent per update (presumably also related to direct load?)
- [ ] aftertouch and program change 7-bit message encoding (see #2292 and #2286):
In the AudioEngine, I can observe that an Aftertouch message is only sent once (not followed by a zero AT message, like described in tickets). Could it be the case that the generic MIDI encoding of 7 Bit messages on 4-Byte USB packets or something related may produce the additional zero messages?

MatthiasSeeber (approved):

PR #2289, opened 2020-11-09
AudioEngine - MIDI CC Ammendments

MatthiasSeeber:
- migrated MIDI handling to separate file
- refactored Encoding and Decoding of 7-Bit (Aftertouch), true 14-Bit (Bender) and clipped 14-bit (remaining Sources)
- no cleanup yet
- first soundcheck worked as expected, further tests are needed


MatthiasSeeber:
tested further:
- MSB-only Ctrl Changes on unipolar HW Source can reach center (50.0%) now as well
- observed Ctrl Change Input und MIDI Output data: seems okay

hhoegelo (approved):

Issue #2250, opened 2020-10-21
AudioEngine - revisit improvements


MatthiasSeeber:
After evaluating smoothing improvement of PR #2248 and successful update `staging` to `unstable`:
- [ ] if lookup concept is benefitial, migrate settings to parameter-db
- [ ] resolve "click" issues: fb mix level (by key track), out mix panning (by key pan)
- [ ] simplify immediate env c: consider re-formulating envelopes (keydown already affects signal) and rendering them after everything else (avoiding override bool and maintaining sample-synchroicity)

majjam:
I am not quite sure if my findings are connected to this issues, but i think the momentary solution for the cenvelopes give artifacts that sound like bird tweeting or kind of portomentoish. (especially on high pitches.)

I am on version:
![grafik](https://user-images.githubusercontent.com/4758563/98445236-d155f280-2116-11eb-9969-e739a20ee7be.png)
 'In the zip files one can find an audio demo and a bank with a preset.
Warning : High pitches !
[tweeting_artifacts.zip](https://github.com/nonlinear-labs-dev/C15/files/5504990/tweeting_artifacts.zip)

MatthiasSeeber:
@majjam Thanks for the notice, this is strange indeed.

So far I found:

- even when forcing Envelope C to start from Zero (Retrig. Hardness = 100%) ...
- even when having no Env C influence on any pitch ...

... the effect occurs. Further, the effect only arises when there is a noticeable difference between last and new played Note in a particular voice. (Enable Mono, play C1, then C4 once: the effect is noticeable - play C4 twice: no effect)

All in all, I don't directly supsect the envelopes, but there's a bug in the Pitch KeyDown-Event processing (also recently touched) for sure (maybe, the Glide mechanism is compromised, maybe the Comb Filter is compromised - I don't know yet).

I hope I can provide a fix whithin the next week.

PR #2293, opened 2020-11-11
AudioEngine - Comb DelaySmoother Fixes

MatthiasSeeber:
Refers to: https://github.com/nonlinear-labs-dev/C15/issues/2250#issuecomment-723461077

Fixes testable with attached Bank:

- [x] fixed KeyDown processing order in Comb Filter: DelaySmoother is now again set after pitch processing (Preset 1 now sounds correct without pitch-sweep glitch)
- [x] Preset 2 ( = Preset 1 in Mono): the glitchy sweeping effect can arise when playing legato (because then there will be no delay smoothing - is this intended/a good idea?)
solved: glide now immediately effective (similar to envelopes)

[Comb Delay Smoother Tests.zip](https://github.com/nonlinear-labs-dev/C15/files/5523605/Comb.Delay.Smoother.Tests.zip)


majjam:
I had short hearing with my own presets yesterday. The issue seems to be fixed.

hhoegelo (approved):

PR #2302, opened 2020-11-12
- remove round trip time measuring for midi-over-ip (proven to be goo…

### LINKED ISSUE #2292: MIDI-Program Change: Beim Empfang eines Program Changes sollte nichts rausgesendet werden.
Zur Zeit werden (wenn der C15 durch den empfangenen PC auf ein neues Preset wechselt) fälschlicherweise sechs PC-Messages rausgeschickt: 
PC Nummer des Presets
PC mit Null
PC Nummer des Presets
PC mit Null
PC Nummer des Presets
PC mit Null

MIDI PC Out

hhoegelo:
…d enough)

SORRY! I had not time to split all the changes into smaller hunks, nor did I carefully test the changes. So, if someone has the time to test the branch, this would be terrific! If not - don't mind - I will test it next Wednesday.

- Midi::SimpleMessage supports 3byte and 2 byte messages
- users of Midi::SimpleMessage havbe to chose 2 byte ctor or 3 byte ctor now
- playground options can be statically configured to skip nltools::msg-initialization
  so tests can set up configurations as they need it
- remove comments in FileSystemSync
- fix race condition in C15Synth when object is destructed before background thread
  could even start
- SimpleMidiMessage used in dsp_dual_host was quite identical to Midi::SimpleMessage,
  so I chose to just use it
- program changes originating from external midi will not bounce back anymore (closes #2292 )


schlenkibus (approved):

projects/bbb/midi-over-ip/src/MidiOverIP.cpp
```diff
@@ -121,21 +112,13 @@ void configureMessaging(const Options &options, bool hasInput, bool hasOutput)
...
 }
 
-void measureRoundTripTime(const nltools::msg::Midi::MessageAcknowledge &msg)
-{
-  auto flightTime = std::chrono::high_resolution_clock::now() - sendTime[msg.id];
-  auto us = std::chrono::duration_cast<std::chrono::microseconds>(flightTime);
-  minRTT = std::min(minRTT, us);
-  maxRTT = std::max(maxRTT, us);
-}
-
```
MatthiasSeeber, line 138: I see that `nltools::msg::Midi::MessageAcknowledge` disappeared completely - hinting at Acknoledge Events only having been bound to round-trip-time measurements. Is this the case, though? (I thought that Acknoledge Events are part of the Heartbeat communication between LPC and AudioEngine)

MatthiasSeeber (approved):
- [x] Program Changes (as expected) do not bounce back anymore

Unfortunately, I cannot really verify if distinct 2- and 3- byte messages are sent correctly - the code looks good though. A test from the C15 itself would be the way to go.

However, I noticed that after starting the WebUI, only selecting a new Preset will not trigger anything. I have to select a Bank first, then the Program Changes will start coming.

(Btw, `#include <future>` is my statement of the day)

PR #2306, opened 2020-11-17
AudioEngine - cleanup

MatthiasSeeber:
- [x] made verified recent potential improvements permanent
- [x] removed unused (pre milestone 1.7) files from project
- [x] removed unused (pre milestone 1.7) defines
- [x] removed unused SimpleSynth from project (Synth, C15Synth and CPUBurningSynth should be fine and up-to-date)

**Note**: the "old" legacy code is still available in https://github.com/nonlinear-labs-dev/nlaudio or stale RELEASE branches and recent Tags (as it sometimes _is_ benefitial to have former implementations available as they are much closer to the Reaktor Prototype).


AntonSchmied:
@hhoegelo 
Here we are using namespace std, which includes our own definition of templates.
https://github.com/nonlinear-labs-dev/C15/blob/dcc07d7cbf813a11c23ae7ad3dce444f091c1132/projects/epc/audio-engine/src/synth/c15-audio-engine/ParallelData.h#L181

Is this a clean approach since there is already a std namespace in the C++ library?

MatthiasSeeber:
> @hhoegelo
> Here we are using namespace std, which includes our own definition of templates.
> 
> https://github.com/nonlinear-labs-dev/C15/blob/dcc07d7cbf813a11c23ae7ad3dce444f091c1132/projects/epc/audio-engine/src/synth/c15-audio-engine/ParallelData.h#L181
> 
> Is this a clean approach since there is already a std namespace in the C++ library?

@AntonSchmied 
I've discussed this with Henry. It in fact is common/good practice to add custom overloaded functions to `std` and we aggreed to leave this as is.
In Theory, an alternative approach would be to implement a namespace (for example `nl`) and provide all associated functions there (like `nl::abs()`). It would then be the responsibility of this namespace to delegate to either `std` or `ParallelData`, depending on the data type. We concluded that this would be a bit of overkill.

AntonSchmied (approved):

hhoegelo (approved):

PR #2335, opened 2020-11-26
Epc2 update intergration

AntonSchmied:
So, both ePC updates are deployed into the C15-update.tar
Then we check the kernel version and link to the corresponding ePC update within the run.sh.
tested both, seems to work fine for both ePCs

!!!***************************************!!!
https://github.com/nonlinear-labs-dev/C15/blob/epc2/build-tools/epc2/host-os/Dockerfile
is adjusted to personal needs, since my Docker Version is out of date

### DIFF
--- a/build-tools/create-c15-update/create-c15-update.sh
+++ b/build-tools/create-c15-update/create-c15-update.sh
@@ -19,22 +19,22 @@ UPDATE_PLAYCONTROLLER=0
 UPDATE_EPC=0
 UPDATE_EPC_2=0
 
-if [[ $ASPECTS = epc_5-7i3 ]]
+if [[ "$ASPECTS" == "epc_5-7i3" ]]
 then
     UPDATE_EPC=1
 fi
 
-if [[ $ASPECTS = epc_10i3 ]]
+if [[ "$ASPECTS" == "epc_10i3" ]]
 then
     UPDATE_EPC_2=1
 fi
 
-if [[ $ASPECTS = *playcontroller* ]]
+if [[ "$ASPECTS" == "playcontroller" ]]
 then
     UPDATE_PLAYCONTROLLER=1
 fi
 
-if [[ $ASPECTS = *bbb* ]]
+if [[ "$ASPECTS" == "bbb" ]]
 then
     UPDATE_BBB=1
 fi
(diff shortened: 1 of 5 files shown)


build-tools/create-c15-update/create-c15-update.sh
```diff
@@ -1,28 +1,34 @@
...
+ASPECTS=$8
 OUT_DIRECTORY=$BINARY_DIR/$OUTNAME
 OUT_TAR=$BINARY_DIR/$OUTNAME.tar
 
 UPDATE_BBB=0
 UPDATE_PLAYCONTROLLER=0
 UPDATE_EPC=0
+UPDATE_EPC_2=0
 
 if [[ $ASPECTS = *epc* ]]
```
hhoegelo, line 22: Not sure, but this will probably match epc2, too?

build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -151,18 +162,17 @@ epc_update() {
...
 
-    epc_fix
-    return_code=$?
-    if [ $return_code -ne 0 ]; then
-        /update/utilities/sshpass -p "sscl" scp -r sscl@192.168.10.10:/tmp/fix_error.log /dev/stdout | cat - >> /update/errors.log
-        pretty "" "$MSG_UPDATING_EPC" "$MSG_FAILED_WITH_ERROR_CODE $return_code" "$MSG_UPDATING_EPC" "$MSG_FAILED_WITH_ERROR_CODE $return_code"
-        sleep 2
-        return 1
+    if [ "$FIX_EPC" = "true" ]; then
+        epc_fix
```
hhoegelo, line 166: this could be simplified to 

```
if ! epc_fix; then
...
fi
```
, no?

build-tools/create-c15-update/create-c15-update.sh
AntonSchmied: I could leave out "epc2" altogether, or do we need to distinguish??

build-tools/create-c15-update/update_scripts/run.sh
AntonSchmied: well ... yes, but then we will not be able to evaluate the return_code we define ourselves in 
https://github.com/nonlinear-labs-dev/C15/blob/unstable/build-tools/create-c15-update/update_scripts/epc_fix.sh

hhoegelo: ah I see

build-tools/create-c15-update/create-c15-update.sh
hhoegelo: I think we should have epc and epc2: This is a developer option to speed up turn around times when debugging. If we remove it, then we have no option for the developer to focus on his ePC.

AntonSchmied:
!!!***************************************!!!
https://github.com/nonlinear-labs-dev/C15/blob/epc2/build-tools/epc2/host-os/Dockerfile
is adjusted to personal needs, since my Docker Version is out of date

hhoegelo (approved):

PR #2343, opened 2020-12-02
AudioEngine - integration of Overlapping Split Points

MatthiasSeeber:
When using overlapping Split Points, the VoiceAllocation had some issues associating different Mono und Unison settings in the different Parts ("Notenhänger"). Now, it should be possible to:
- play overlapping keys without issues, regardless of different Mono/Unison settings per part
- play overlapping keys, even if all voices in a part are exhausted (note steal) without issues
- play overlapping and non-overlapping keys simultaneously, even if all voices in a part are exhausted (note steal) without issue

A thorough Test of overlapping Split Points is necessary, of course.

### DIFF
--- a/projects/epc/audio-engine/src/synth/c15-audio-engine/voice_allocation.h
+++ b/projects/epc/audio-engine/src/synth/c15-audio-engine/voice_allocation.h
@@ -63,7 +63,7 @@ struct KeyAssignment
 
 struct VoiceAssignment
 {
-  uint32_t m_keyId = 0;
+  uint32_t m_keyNumber = 0;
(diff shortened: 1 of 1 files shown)


MatthiasSeeber:
When merging, there is no need to delete the branch. Another PR, implementing the first iteration of 128-Key-usage will hopefully come soon.

hhoegelo (approved):
I did not really work through und fully understand the PR but talked to @MatthiasSeeber. We decided to just merge the PR as he already tested carefully and voice allocation is about to be tested when stabilizing the December release anyway.

PR #2346, opened 2020-12-02
update packages are now installed with all their dependencies from

hhoegelo:
archive.archlinux.org


build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
```diff
@@ -19,30 +20,40 @@ download_package() {
     if find ./$1 > /dev/null; then
         return 0
     fi
-
+    
     for server in $UPDATE_PACKAGE_SERVERS; do
```
AntonSchmied, line 24: Do we need a loop here? I see only 1 server l. 4. Or might this change with time?

hhoegelo: Indeed the loop is a left-over from previous implementations. Still, I think we should target a NL-administrated package cache and only fall back to public arch-servers.  I will implement this right now.

AntonSchmied (approved):
Can't find anything else to complain about 👍

PR #2351, opened 2020-12-03
Remove existing link in order to not fail following "ln -s".

hhoegelo:
Respect operator precedence in error handling.


AntonSchmied (approved):
yep!

PR #2356, opened 2020-12-08
Issue 2284

schlenkibus:
feature is usable but not design complete
see #2363

### DIFF
--- a/projects/epc/playground/src/presets/PresetManager.cpp
+++ b/projects/epc/playground/src/presets/PresetManager.cpp
@@ -1019,9 +1019,18 @@ Uuid PresetManager::getMidiSelectedBank() const
 void PresetManager::selectMidiBank(UNDO::Transaction *trans, const Uuid &uuid)
 {
   auto swapData = UNDO::createSwapData(uuid);
-  trans->addSimpleCommand([=](auto) {
+
+  trans->addSimpleCommand([=, newUuid = uuid](auto) {
+    const auto oldMidiBankUuid = m_midiSelectedBank;
+
     swapData->swapWith(m_midiSelectedBank);
     m_sigMidiBankSelection.send(m_midiSelectedBank);
+
+    if(auto oldMidiBank = findBank(oldMidiBankUuid))
+      oldMidiBank->invalidate();
+
+    if(auto newMidiBank = findBank(newUuid))
+      newMidiBank->invalidate();
   });
 }
 
--- a/projects/epc/playground/src/presets/PresetManager.cpp
+++ b/projects/epc/playground/src/presets/PresetManager.cpp
@@ -762,7 +762,8 @@ void PresetManager::writeDocument(Writer &writer, UpdateDocumentContributor::tUp
                         forEachBank([&](Bank *bank) { anyBankChanged |= (bank->didChangeSince(knownRevision)); });
 
                       writer.writeTag("banks", Attribute("changed", anyBankChanged),
-                                      Attribute("selected-bank", getSelectedBankUuid().raw()), [&]() {
+                                      Attribute("selected-bank", getSelectedBankUuid().raw()),
+                                      Attribute("selected-midi-bank", getMidiSelectedBank().raw()), [&]() {
                                         if(anyBankChanged)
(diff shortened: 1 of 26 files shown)


projects/epc/playground/src/presets/PresetManager.cpp
```diff
@@ -1009,3 +1010,26 @@ sigc::connection PresetManager::onPresetStoreHappened(sigc::slot<void> cb)
...
 }
+
+Uuid PresetManager::getMidiSelectedBank() const
+{
+  return m_midiSelectedBank;
+}
+
+void PresetManager::selectMidiBank(UNDO::Transaction *trans, const Uuid &uuid)
+{
+  const auto oldUuid = m_midiSelectedBank;
```
hhoegelo, line 1021: I don't think, it's a good idea to put the notification into a separate transaction. For the observer the following assertion would trigger on undo, because, the event is sent first, then the change happens. Consider not using the addUndoSwap, but swap in the simple command.

Observer::onMidiBankSelectionChanged(String selectedBankUuid) {
 assert(selectedBankUuid == pm->findMidiSelectedBank().getUuid());
}

schlenkibus: Fixed!

projects/epc/playground/src/presets/PresetManager.cpp
```diff
@@ -1009,3 +1010,33 @@ sigc::connection PresetManager::onPresetStoreHappened(sigc::slot<void> cb)
...
+
+  trans->addSimpleCommand(
```
hhoegelo, line 1023: I usually try to avoid two lamdas and use a swap approach in such cases, like this:

`
auto swapData = UNDO::createSwapData(uuid);
transaction->addSimpleCommand([=](auto) mutable {
      swapData->swapWith(m_midiSelectedBank);
      m_sigMidiBankSelection.send(m_midiSelectedBank);
    });
`

But that is a matter of taste and your explicitly expressed solution is fine, too.

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -46,7 +46,7 @@ AudioEngineProxy::AudioEngineProxy()
 
   receive<Midi::ProgramChangeMessage>(EndPoint::Playground, [=](const auto &msg) {
     if(auto lock = m_programChangeRecursion.lock())
-      if(auto bank = pm->getSelectedBank())
+      if(auto bank = pm->findMidiSelectedBank())
         if(msg.program < bank->getNumPresets())
           bank->selectPreset(msg.program);
   });
```
hhoegelo, line 52: AudioEngineProxy::onBankChanged() has to be fixed, too. It has to send program changes only if the current bank is the midi bank.

projects/epc/playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/BankNameLabel.cpp
```diff
@@ -18,9 +20,16 @@ void BankNameLabel::updateLabel(Bank *newBank)
 {
   if(newBank)
   {
+    auto pm = Application::get().getPresetManager();
     auto bankName = newBank->getName(true);
-    auto pos = Application::get().getPresetManager()->getBankPosition(newBank->getUuid()) + 1;
+    auto pos = pm->getBankPosition(newBank->getUuid()) + 1;
     auto s = to_string(pos) + ": " + bankName;
+    lastBank = newBank;
```
hhoegelo, line 27: This smells.
I think you should not fiddle with the text and have all this logic to not add the sign twice (which will probably forget to add the sign if the bank is renamed).
Instead, consider decorating the text by overriding getText() like this:

```
StringAndSuffix BankNameLabel::getText()
{
   auto t = super::getText();
   if(isMidiBank())
      return addMidiSign(t);
   return t;
}
```

Then, onMidiSelectionHappened only needs to set the control dirty.

hhoegelo (approved):

PR #2357, opened 2020-12-08
Fixes: sound-tab opens on single preset load (if previous loaded preset was dual)

schlenkibus:
@MatthiasSeeber i fixed the issue you raised me: no ticket for it but i hope you remember which bug this was if not its this:
load dual preset and select preset / parameter tab, load single preset through world / boled -> webui belt tab got sound focus.


MatthiasSeeber (approved):
I do remember, yes, though I wasn't aware that this also happened when having the Parameter Tab open in the Sound Belt.
Anyway, the fix seems to be working fine and the code rule for focussing on the Sound Tab only when a "fadeview" is in focus (which is not present in Single sounds) makes total sense to me.

PR #2359, opened 2020-12-09
Refactor and rework Split/Fadepoint editors

### LINKED ISSUE #2277: [Layer] Fade Editor - clipping Ranges
Consider having a Fade Range of 12 Semitones (above, see dotted red shape). When the Fade From is shifted towards the right end, the current implementation uses a simple, but incorrect approach (below).

FadeEditor-FadeRangeClipping

It would be consistent if we showed the blue area as indicated by the dotted red line - additionally clipping the shape when the piano area is exceeded.

Further we may want to reconsider the clipping of drag gestures on the lower "Fade Range" icon / "anfasser". (Currently, the lower icon can not be shifted beyond the piano, although the value hasn't yet reached the full 60 semitones)

### LINKED ISSUE #2348: WebUI: Split-Point-Editor und Fade-Range-Editor: Rahmen um Anfasser nur während des Click+Drag zeigen

hhoegelo:
This will also fix #2348 (highlight when mouse down)
Also closes #2277


MatthiasSeeber (changes requested):
- the "Anfassers" not exactly following the mouse cursor position is an inconvenience I could live with, but it certainly is not optimal and I can imagine that some users may voice complaints about that (whether or not this should be fixed immediately is another question)
- [x] in Split sounds, there still is an issue:
Screenshot from 2020-12-09 14-01-19
These are the initial (synced) Split Point settings. The indication of Part I is wrong, it should cover the black F#3 key as well

schlenkibus:
> * the "Anfassers" not exactly following the mouse cursor position is an inconvenience I could live with, but it certainly is not optimal and I can imagine that some users may voice complaints about that (whether or not this should be fixed immediately is another question)
> * [x]  in Split sounds, there still is an issue:
>   Screenshot from 2020-12-09 14-01-19
>   These are the initial (synced) Split Point settings. The indication of Part I is wrong, it should cover the black F#3 key as well

The issue with default Splitpoint CPs seems to be fixed.
But the mouseposition and "Anfasser" discrepancy is very noticable on my UHD (1440p) Monitor. I would like to see that fixed, if possible.

schlenkibus (changes requested):
See my comment above.

KlausStrohhaecker:
Somewhat relatetd, see also https://github.com/nonlinear-labs-dev/C15/issues/2365 for an alternate way of setting a splitpoint (or whatever) by actually pressing a keybed key.

hhoegelo:
I found some issues and now everything seems to work (on my machine ;-) )

projects/epc/playground/src/parameters/SplitPointParameter.h
```diff
@@ -5,6 +5,33 @@
...
+ * 60 is the highest key on the C15 keyboard, C6
+ * 
+ * Note: If SplitPointSync-Setting is enabled, then 
+ *       SplitPointVGI has to be <= 59
+ *       SplitPointVGII has to be >= 1
+ * 
+ * For example: 
+ * 
+ * SplitPoint Parameter |   VG I   |   VG II   
+ *    Display Value     |          |
```
MatthiasSeeber, line 27: nomencalture: these are rather raw values or key numbers than display values.

MatthiasSeeber (approved):

schlenkibus (approved):

PR #2364, opened 2020-12-09
WebUI: Value Edit Bug Fix

schlenkibus:
changed doubleClick handler to be ignored, previously continously clicking on the change arrow lead to random jumps over values, as click was called from double click, this is now fixed and you can spam-click the changed arrows as much as you like without missing value steps


MatthiasSeeber (approved):
Not sure if I can really spot any difference when spam-clicking on the arrows (in comparison to #2359), but certainly, the arrows do not trigger a reset of the position and behave convincingly. Reset of the position by double-click on the value also seems okay to me.

hhoegelo (approved):

PR #2369, opened 2020-12-10
closes #2367

### LINKED ISSUE #2367: BOLED/SOLED-Screensaver: Kosmetische Verbesserungen (s.u.)
- [ ] Statt "Nonlinear Labs - C15" nur "C15" als bewegten Schriftzug verwenden.
- [ ] Im SOLED die Bewegungs-Geschwindigkeit halbieren.

schlenkibus:
soled screensaver slower movement, and both screensaver-strings are now 'C15'

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/base-unit/soled/SOLEDScreenSaver.cpp
+++ b/projects/epc/playground/src/proxies/hwui/base-unit/soled/SOLEDScreenSaver.cpp
@@ -11,7 +11,7 @@ SOLEDScreenSaver::SOLEDScreenSaver(OLEDProxy& oled)
 {
 
   m_animationConnection = Application::get().getMainContext()->signal_timeout().connect(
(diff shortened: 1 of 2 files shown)


MatthiasSeeber (commented):
I noticed an odd behavior: the timeout setting on my system is one minute, and after running the BBBB and navigating to the Setup Menu, the Screensaver started despite me clicking and navigating the Setup. I couldn't reproduce that, though.

One further question: only in the Soled, the C15 moves slower. Is this intended?

schlenkibus:
Yes the movement of the Soled was supposed to be half as fast as the Boleds movement. I will take a look at what could have caused that issue with the screensaver starting.

MatthiasSeeber (approved):
Seems good to me. The Labels can "wander" in any direction up to the last pixel on the corresp. corner. The Screensaver timeout seems to work even when starting and doing nothing. Setup Menu oddities could not be reproduced.

I think we're good to go.

PR #2378, opened 2020-12-15
BOLED: Cleanup: move all startTransactions into UseCases

schlenkibus:


projects/epc/playground/src/groups/ParameterGroup.h
```diff
@@ -30,6 +30,13 @@ class ParameterGroup : public UpdateDocumentContributor, public IntrusiveListIte
 
   tParameterPtr findParameterByID(const ParameterId &id) const;
 
+  template <typename tParameterPtrT> tParameterPtrT *findAndCastParameterByID(const ParameterId &id) const
+  {
+    if(auto p = getParameterByID(id))
```
hhoegelo, line 35: Not 100% sure, but I'd say you do not need the branch, as dynamic casting a nullptr is valid.

hhoegelo (approved):
I did not fully work through the PR. But I can approve that the refactoring is the right direction and your work definitely improves code quality and thus should be merged to 'unstable' even though there might be new errors introduced.

PR #2419, opened 2021-01-06
- Fix build problems by bumping to ubuntu 20.04 (LTS)

hhoegelo:
- switch to new nonlinux version which has nlimagemaker for rescue usb sticks

@All: I had to remove the build-tools/bbb folder in my build directory after applying that fix.
@schlenkibus I will add another PR for staging, without the nonlinux update


AntonSchmied (approved):
seems legit and I was able to build successfully!

schlenkibus (approved):

PR #2420, opened 2021-01-06
fix build problems by using ubuntu 20.04 LTS as docker os

hhoegelo:


build-tools/bbb/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM ubuntu:16.04
```
schlenkibus, line 1: Can we also use 20.04 here? As 16.04 is scheduled to reach EOL in April 2021.

hhoegelo: No, the buildroot version we use only works with ubuntu 16.04

PR #2423, opened 2021-01-07
Fix direct load

### LINKED ISSUE #2415: UNDO Direct Load
Selecting and the consequential loading of a Preset is done in two UNDO Transactions. Therefore 'undoing' an DL-Load action leads to inconsistent states (loaded presets is not selected)

schlenkibus:
closes #2415


projects/epc/playground/src/device-settings/DirectLoadSetting.cpp
```diff
@@ -29,7 +29,7 @@ bool DirectLoadSetting::set(BooleanSettings m)
   if(!m_isLoading)
     if(auto pm = Application::get().getPresetManager())
       if(m == BooleanSettings::BOOLEAN_SETTING_TRUE)
-        pm->doAutoLoadSelectedPreset();
+        pm->doAutoLoadSelectedPreset(nullptr);
```
schlenkibus, line 32: Use UseCase -> call autoloadselected

PR #2429, opened 2021-01-19
Midi Settings Part I

schlenkibus:
- [x] Create new Settings
- [x] create BOLED setup layouts
- [x] propagate model to WebUI: xml -> model -> presenter  
- [x] build html midi page
- [x] implement midi functionality
- [ ] refactor AE to use generic Input/Output events instead of multiple in channels (TCD/Midi/possible KeyDown/KeyUp message from Bridge)
- [ ] fix local notes setting -> if disabled no midi out will be send
- [ ] implement split channel functionality


projects/epc/playground/src/device-settings/BooleanSetting.cpp
```diff
@@ -30,3 +30,16 @@ const std::vector<Glib::ustring> &BooleanSetting::enumToDisplayString() const
   static std::vector<Glib::ustring> s_modeNames = { "On", "Off" };
   return s_modeNames;
 }
+
+void BooleanSetting::incDec(int inc, bool wrap)
```
hhoegelo, line 34: I don't get it - lets whereby about it.

projects/epc/playground/src/proxies/hwui/panel-unit/boled/setup/SettingEditors.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
+
+#include "MenuEditor.h"
+#include "device-settings/Settings.h"
+#include "device-settings/Setting.h"
+#include <Application.h>
+
+template <typename tSetting> class BooleanSettingsEditor2 : public MenuEditor
```
hhoegelo, line 8: I wonder why the generic EnumSettingEditor doesn't work for boolean values?

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/dataModel/setup/SetupModel.java
```diff
@@ -69,6 +69,23 @@ static public SetupModel get() {
 		off, percent_10, percent_25, percent_50
 	}
 
+	public enum MidiReceiveChannel {
```
hhoegelo, line 72: Not sure, but maybe we can just have a "SendMidiChannel" enumeration and a "ReceiveMidiChannel" enumeration and share these between the according settings?

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/presenters/MidiSettingsProvider.java
```diff
@@ -0,0 +1,129 @@
...
+
+	private LinkedList<Function<MidiSettings, Boolean>> clients = new LinkedList<Function<MidiSettings, Boolean>>();
+	private MidiSettings settings = new MidiSettings();
+
+	private MidiSettingsProvider() {
+		SystemSettings s = SetupModel.get().systemSettings;
+
+        //Local
+        s.localControllers.onChange(t -> {
+            settings.localControllers.value = t == BooleanValues.on;
```
hhoegelo, line 25: I know it's a annoying task, but we should only "notifyClients()" if the according value did actually change. I did not manage to come up with a generic solution for this so far, so I am afraid we have to do it the 

if(a != b) { a = b; notifyClients();} 

... style. Maybe you have a better idea, Java 11 is an option, too.

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/dataModel/setup/SetupModel.java
schlenkibus: i would rather prevent missuses of the complete enums at every usecase. For example: prevent "follow" for non split channels. So if you would be fine with 4 enums in total I would prefer to keep the code as is.

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -52,24 +52,31 @@ C15Synth::C15Synth(AudioEngineOptions* options)
 
   // receive program changes from playground and dispatch it to midi-over-ip
   receive<nltools::msg::Midi::ProgramChangeMessage>(EndPoint::AudioEngine, [this](const auto& pc) {
-    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage{ 0xC0, pc.program });
+    const uint8_t channel0Pattern = 0b11000000;
+    const uint8_t newStatus = channel0Pattern | m_midiOptions.getSendChannel();
+    nltools::Log::error("received ProgramChange Message at AE with preset:", (int) pc.program,
```
hhoegelo, line 57: Can the error message be removed?

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -52,24 +52,31 @@ C15Synth::C15Synth(AudioEngineOptions* options)
 
   // receive program changes from playground and dispatch it to midi-over-ip
   receive<nltools::msg::Midi::ProgramChangeMessage>(EndPoint::AudioEngine, [this](const auto& pc) {
-    m_externalMidiOutBuffer.push(nltools::msg::Midi::SimpleMessage{ 0xC0, pc.program });
+    const uint8_t channel0Pattern = 0b11000000;
```
hhoegelo, line 55: name channel0Pattern is misleading - consider "programChangeStatus" or similar.

hhoegelo: (you already have the MIDI_PROGRAMCHANGE_PATTERN const below, that could be reused here)

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1739,8 +1741,9 @@ void dsp_host_dual::hwModChain(HW_Src_Param* _src, const uint32_t _id, const flo
           macro->m_unclipped += (_inc * amount->m_position);  // fixing #2023:
           // only rely on unclipped (always up-to-date)
           // NOTE: hopefully, this won't introduce accumulating floating point rounding errors !!!
-          const float clipped
-              = macro->m_unclipped < 0.0f ? 0.0f : macro->m_unclipped > 1.0f ? 1.0f : macro->m_unclipped;
+          const float clipped = macro->m_unclipped < 0.0f ? 0.0f
```
hhoegelo, line 1744: consider using std::clamp from <algorithm>

PR #2445, opened 2021-01-26
closes #2431

### LINKED ISSUE #2431: SOLED: Preset Mode: An den Presetlabels fehlen die Split/Layer-Symbole

schlenkibus:


projects/epc/playground/src/presets/Preset.cpp
```diff
@@ -552,3 +552,13 @@ std::vector<std::pair<GroupId, const PresetParameterGroup *>> Preset::getGroups(
   }
   return ret;
 }
+
+Glib::ustring Preset::getTypeUnicode(const Preset *preset)
```
hhoegelo, line 556: why did you choose to make this a static function?

schlenkibus: i made this an proper member now

projects/epc/playground/src/proxies/hwui/panel-unit/boled/preset-screens/controls/PresetTypeLabel.cpp
```diff
@@ -120,9 +120,9 @@ std::string typeToString(const SoundType &type)
     case SoundType::Invalid:
```
hhoegelo, line 120: As far as I can see, this method can be removed. Instead, we can call getTypeUnicode on the preset at hand.

hhoegelo (approved):

PR #2523, opened 2021-03-18
Enable Wifi on ePC2

AntonSchmied:
This deploys a config file for the Networkmanager to read from and enables the latter. For now the SSID is fixed and cannot be changed. It is enough to start testing the functionality though.

### DIFF
--- a/build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
+++ b/build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
@@ -24,42 +24,42 @@ install_packages() {
 }
 
 setup_wifi() {
-    (
-    cat <<-'ENDOFHERE'
-[connection]
-id=C15
-uuid=61679179-6804-4197-b476-eacad1d492e4
-type=wifi
-interface-name=wlp0s20f3
-permissions=
 
-[wifi]
-band=bg
-channel=7
-mac-address-blacklist=
-mode=ap
-ssid=NL-C15-Unit-EPC-00001
+    cat <<- ENDOFHERE > /overlay-fs/etc/NetworkManager/system-connections/C15.nmconnection
+    [connection]
+    id=C15
+    uuid=61679179-6804-4197-b476-eacad1d492e4
+    type=wifi
+    interface-name=wlp0s20f3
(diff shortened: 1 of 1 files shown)


build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
```diff
@@ -21,7 +21,47 @@ install_packages() {
     
     echo "en_US.UTF-8 UTF-8" > /overlay-fs/etc/locale.gen
     /bin/arch-chroot /overlay-fs locale-gen
+}
+
+setup_wifi() {
+    (
```
hhoegelo, line 27: [did you find | can you explain] the reason for the braces?

AntonSchmied: no ... so i kicked them out and got rid of the quotes, which were also unnecessary in out case.
tested, works fine.

build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
```diff
@@ -21,7 +21,46 @@ install_packages() {
     
     echo "en_US.UTF-8 UTF-8" > /overlay-fs/etc/locale.gen
     /bin/arch-chroot /overlay-fs locale-gen
+}
+
+setup_wifi() {
+
+    cat <<- ENDOFHERE > /overlay-fs/etc/NetworkManager/system-connections/C15.nmconnection
```
hhoegelo, line 28: Looks good - but I have still problems understanding the bash heredoc syntax: piping into the file is done between declaring the delimiter and the actual here-doc? WTF? How does the interpreter decide where the doc starts? After a new line? Suspicious, in my opinion.
Anyway, glad you found a way of having readable AND working code.

hhoegelo (approved):

build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
AntonSchmied: Best description I found was here https://linuxize.com/post/bash-heredoc/

PR #2541, opened 2021-03-24
Issue 2528

AntonSchmied:
closes #2528

PR #2542, opened 2021-03-24
MIDI/TCD IOEventStage

schlenkibus:


AntonSchmied:
Since Stephan is testing on a NUC 10i3, the latest changes on "unstable" are important, especially #2541, which ensures that amixer is set to 100% once the audio-engine services starts. The approach we used so far unfortunately only works for older NUCs.

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
```diff
@@ -704,8 +693,37 @@ void InputEventStage::onMIDIHWChanged(MIDIDecoder *decoder)
       }
       else
       {
-        //we land here in the special mapping cases
-        nltools::Log::warning("This mapping is not implemented yet!");
+        if(hwID == 4 && hwRes.cases == MIDIDecoder::MidiHWChangeSpecialCases::ChannelPitchbend)
+        {
+          onHWChanged(4, decoder->getValue(), DSPInterface::HWChangeSource::MIDI);
+        }
```
MatthiasSeeber, line 699: Unfortunately, a `MIDI::PitchBend` Message will still affect the `HW::Bender` regardless of the Mapping (when the Bender is set to a `MSB/LSB CC`, the CC will correctly change the Bender, but the `PitchBend` will still do so as well).
In order to prevent this, I would suggest to encapsulate `onHWChanged` by another check to be sure that the PitchBend Message is allowed to affect the Bender:

`if(m_options->getBenderMSBCC().first == false) // only if Bender setting matches Pitchbend`
`{`
`onHWChanged(4, decoder->getValue(), DSPInterface::HWChangeSource::MIDI);`
`}`

MatthiasSeeber (changes requested):

projects/epc/audio-engine/src/synth/input/MIDIDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  if(controlChangeID == 88 && m_options->enableHighVelCC())
+  {
+    m_MidiLSB = _data1;
+    return;
+  }
+
+  const auto isMSB = controlChangeID > 0 && controlChangeID < 32;
+  const auto isLSB = controlChangeID > 33 && controlChangeID < 64;
+  const auto isSwitching = controlChangeID > 65 && controlChangeID < 70;
```
MatthiasSeeber, line 135: - [x] `isMSB` is correct: Ctrl ID is within [1 .. 31] Range
- [ ] `isLSB` is incorrect: Ctrl ID is within [33 .. 63] Range, so `controlChangeID > 32 && controlChangeID < 64` applies
- [ ] `isSwitching` is incorrect: Ctrl ID is within [64 .. 69] Range, so `controlChangeID > 63 && controlChangeID < 70` applies

projects/epc/audio-engine/src/synth/input/MIDIDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  if(isMSB || isSwitching)
+  {
+    hwResult.cases = MidiHWChangeSpecialCases::CC;
+    hwResult.receivedCC = controlChangeID;
+
+    if(isSwitching)
+      hwResult.undecodedValueBytes = { static_cast<uint8_t>(_data1 > 64 ? 127 : 0), 0 };
```
MatthiasSeeber, line 143: Our "Switch Condition" is: **0** (for x < 0.5) or **1** otherwise. The **64** means exactly 0.5, so the current implementation does **0** (for x <= 0.5) or **1** otherwise. I'd prefer identical Switch Conditions for Receivev and Send an would like to change this test to:
`...(_data1 > 63 ? 127 : 0)...`

MatthiasSeeber (approved):

MatthiasSeeber (approved):

Issue #2575, opened 2021-04-20
Fix Acceptance-Test VoiceAllocation -> Should the Level really be Above 0.5?


schlenkibus:
How is this Feature supposed to function, do parameters/types of the current and next loaded preset play a role if the notes will re-trigger or stay?

Scenario:
Voices remain if old Type == new Type && old unisonVoices == new unisonVoices && old monoEnable == new monoEnable
+ Envelope + Comb can also have impacts on this

maybe use gate signal as Matthias suggested

MatthiasSeeber:
Impacts can also be: Sustain and Level Parameters, Part Mute and other extreme conditions.

In #2576 I provided a new Acceptance Tester for the AudioEngine, capable of obtaining the "active" Voices of the C15. This should suffice to perform `SCENARIO("Voices remain on preset change")`, given the following circumstances.

A Reset happens when:
- specific changed Parameters **Unison Voices**, **Mono Enable** (in Preset or Parameter Message)
- received Preset Message: next to Unison and Mono change, a Reset occurs when the Sound Type differs (Preset Glitch Suppression DOES NOT affect this)
- received MIDI Setting Message from Playground

PR #2576, opened 2021-04-20
Ae acceptance tester

MatthiasSeeber:
Provides `DspHostDualTester` class for the Audio Engine, usable for AE-specific acceptance tests. _(This covers #2575, but doesn't provide any actual Glitch Suppression tests.)_

- [x] can obtain active and assignable voices
- [x] can mock mono enable and unison voices parameter messages
- [x] can load malformed, initial single/split/layer presets with definable mono/unison
- [x] can perform TCD key and MIDI Note events (independant from InputEventStage)

So far established VoiceAllocation acceptance-tests:

- **Initial State** _(no active voices, full polyphony, silence)_
- **Key Association** _(repeated Key Events ignored, separation of TCD and MIDI Events, three C3 in a Split Sound: Internal, Primary, Secondary)_
- **128 Keys** _(Internal, External: full 7 Bit MIDI Note Range)_
- **Reset Behavior** _(active Voices vs Reset Events: Synth Reset, MIDI Setting, Parameter Changed, Preset)_

I think this should do for now. Further Tests (Preset Glitch Suppression, Split Point Behavior, ...) would require InputEventStage and Application, as well as prepared Preset XML Files. I'd suggest to structure this in forthcoming tickets.


projects/epc/acceptance-tests/src/VoiceAllocation.cpp
```diff
@@ -7,6 +7,85 @@
 
 namespace Tests
 {
+  TEST_CASE("Initial State")
+  {
+    using namespace std::chrono_literals;
+    auto options = createEmptyAudioEngineOptions();
+    auto synth = std::make_unique<C15Synth>(options.get());
+    DspHostDualTester tester{ synth->getDsp() };
+    GIVEN("No Active Voices")
```
schlenkibus, line 16: I would use THEN here as it describes the CHECK better. I generally would use GIVEN only if the state/data, that is modified in that scope, is used for further nested CHECKS and operations. 

For example:
TEST_CASE "Voice Allocation Simple Tests"
    WHEN("synth is in initial state")
         THEN("no voices are active")
         CHECK(tester.getActiveVoices() == 0);
         THEN("full polyphony is available")
         CHECK(tester.getAssignableVoices() == 24);
         WHEN("TCD KeyDown")
             THEN("1 Voice active")
             CHECK(...)
             WHEN("Same TCD")
             
But in general it reads much better than the first version :+1:

MatthiasSeeber: Would the following do better?

TEST_CASE("Initial State")
  {
    //  ...
    THEN("No Active Voices")
    {
      CHECK(tester.getActiveVoices(VoiceGroup::Global) == 0);
    }
    THEN("Full Polyphony")
    {
      CHECK(tester.getAssignableVoices() == C15::Config::total_polyphony);
    }
// ...

PR #2587, opened 2021-04-28
do wave file export instead of flac export

hhoegelo:


projects/epc/audio-engine/src/recorder/NetworkServer.cpp
```diff
@@ -3,6 +3,44 @@
...
+      fileSize = dataLength + sizeof(WaveHeader) - 8;
+    }
+
+    char riff[4];
+    uint32_t fileSize;
+    char wave[4];
+    char fmt[4];
+    uint32_t fmtLength = 16;
+    uint16_t formatTag = 3;  // floats
+    uint16_t channels = c_numChannels;
```
KlausStrohhaecker, line 34: Why float32 format? It just uses more bytes for the same precision (24bit mantissa) than packed 24bit, a 33% file size increase.

hhoegelo: Yes you are right. I took f32 because that's the FlacDecoders output and I could save another conversion. But after having a second look, I see that the library used by *MY" flac decoder outputs i32, which I convert into f32. So I should add an interface for grabbing the i32 before they are converted to f32 and thus *really* save some cycles.
Tl;dr: I will fix it according to your proposal.

PR #2591, opened 2021-04-28
issue 2574

schlenkibus:
- [x] added midi bank editor for boled
- [x] midi bank editor for webui ?


projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
```diff
@@ -301,6 +311,29 @@ public void connectUpdate() {
...
+	}
+
+	private Map<Integer, Bank> midiBankIndexToBankMap = new HashMap<Integer, Bank>();
+
+	private void applyBanks(Map<String, Bank> map) {
+		pcBanks.clear();
+		midiBankIndexToBankMap.clear();
+
+		int index = 0;
+		for(String uuid: map.keySet()) {
```
hhoegelo, line 328: I expect a wrong order. The index will be according to the map order - so alphabetically regarding the uuid, but should be ordered by the order number. 
I think you have to resort the banks according to their order number first.
Instead of preparing the midiBankIndexToBankMap, I would prefer searching for the right bank in pcBanks.addChangeHandler, because of possible data races that would finally lead to exceptions in midiBankIndexToBankMap.get().

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
```diff
@@ -305,6 +324,42 @@ public void connectUpdate() {
...
+			}
+		});
+		return bb;
+	}
+
+	private void applyBanks(Map<String, Bank> map) {
+		pcBanks.clear();
+
+		List<com.nonlinearlabs.client.world.maps.presets.bank.Bank> banks = getBanksSortedByNumber();
+		Tracer.log("new Map.Size(): " + map.size());
```
hhoegelo, line 352: are the logs still needed?

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
```diff
@@ -279,6 +291,13 @@ public void connectEventHandlers() {
 		benderMapping.addChangeHandler(e -> settings.setPitchbendMapping(BenderCCMapping.values()[benderMapping.getSelectedIndex()]));
 		highVeloCCOn.addValueChangeHandler(e -> settings.setHighVelocityCC(BooleanValues.on));
 		highVeloCCOff.addValueChangeHandler(e -> settings.setHighVelocityCC(BooleanValues.off));
+
+		pcBanks.addChangeHandler(e -> {
+			List<com.nonlinearlabs.client.world.maps.presets.bank.Bank> sortedBanks = getBanksSortedByNumber();
```
hhoegelo, line 296: I am a bit worried here, because I am not 100% sure that the array referred to by this method is identical to the one used when setting up the control in applyBanks(t). If, for whatever reason the  data model changed in between, but the applyBanks has not yet been called, we will have an exception here.

I suggest using addItem(java.lang.String item, HasDirection.Direction dir, java.lang.String value) in applyBank and put the UUID as value into the item. Then you can just use getSelectedValue() here and there is no need for creating another sorted array.

PR #2599, opened 2021-04-29
Fix combining preset loads into one undo transaction

hhoegelo:
Subsequent preset selections should be combined into a ContinuousTransaction.
With direct load enabled, those transactions should loose their 'continueing'
state and create new undo entries each.

### DIFF
--- a/projects/epc/playground/src/libundo/undo/ContinuousTransaction.cpp
+++ b/projects/epc/playground/src/libundo/undo/ContinuousTransaction.cpp
@@ -44,7 +44,7 @@ namespace UNDO
     m_continuing = false;
   }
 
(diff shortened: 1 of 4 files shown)


projects/epc/playground/src/libundo/undo/ContinuousTransaction.cpp
```diff
@@ -39,6 +39,16 @@ namespace UNDO
     return m_closingCommand.get();
   }
 
+  void ContinuousTransaction::stopContinuation()
+  {
+    m_continueing = false;
```
schlenkibus, line 44: minor: correct spelling to 'continuing'

schlenkibus (approved):

PR #2601, opened 2021-04-29
Epc2 wifi

AntonSchmied:
working on passing on the wifi host credentials to the networkmanager on epc2
TODOS:
- [x] allow changing SSID/Passphrase
- [x] disable/ enable Wifi
- [x] check on which ePC we're working on -> unnecessary


projects/epc/playground/src/device-settings/EpcWifi.cpp
```diff
@@ -0,0 +1,92 @@
...
+    m_newPassphrase = _newPassphrase;
+}
+
+bool EpcWifi::syncCredentials()
+{
+    if (checkConnectionStatus())
+    {
+        if (m_currentPassphrase != m_newPassphrase)
+        {
+                m_currentPassphrase = m_newPassphrase;
```
hhoegelo, line 30: This should happen only on successfully setting the passphrase, i.e. in Ln 61.

projects/epc/playground/src/device-settings/EpcWifi.cpp
```diff
@@ -0,0 +1,92 @@
...
+                updatePassphrase();
+                reloadConnection();
+        }
+
+        if (m_currentSSID != m_newSSID)
+        {
+            m_currentSSID = m_newSSID;
```
hhoegelo, line 37: This should happen only on successfully setting the ssid, i.e. in Ln 69

projects/epc/playground/src/device-settings/EpcWifi.cpp
```diff
@@ -0,0 +1,92 @@
...
+            updateSSID();
+            reloadConnection();
+        }
+    }
+    return true;
+}
+
+
+bool EpcWifi::checkConnectionStatus()
```
hhoegelo, line 46: checkConnectionStatus is misleading:
It always triggers a new check, but returns the result of a previous check, as the command is executed asynchronously. In between, connection might already be down. For sure, the first invocation will return false, even though the connection might already exist.

So, to be correct, one would need a daisy chain:

I suggest to 
- have a member counter indicating the number of calls in flight
- timer triggers syncCredentials() as already is
- if callsInFlight is != 0 (indicating change already in progress) or all members are as expected, do nothing
- else start the daisy chain, where, for each command line spawned, you would increment the callsInFlight, and for each returned callback, successful or failing, you would decrement the counter. This will avoid multiple update chains simultaneously.  
- the 'chain' is formed by doing one step after another, triggered only in the success-callback of the previous command, so checkConnectionStatus' will updateSsid and/or updatePassphrase on success. The success-callbacks of the two update-commands will check the callsInFlight: If the according call was the last one, reloadConnection can be called from within the success callback.

AntonSchmied:
Daisy-Chain implemented, I still have to add an On/Off Switch for ePC local Wifi.
However, maybe we can decide first, whether EpcWifi should follow the rules of other Setting or remain a shared pointer. As far as I understand, this would effect the On/Off switch implementation ...

AntonSchmied:
I have tested it and couldn't find any faulty behavior. 
ToDo:
- [ ] leave it as a shared pointer?
- [ ] Find a way to mask the IP Address (currently set to 192.168.100.101) with "C15webUI" or similar, for more user friendlier access

schlenkibus (approved):

build-tools/epc2/create-update/createEPC2UpdateDockerCommands.sh
```diff
@@ -61,6 +61,74 @@ ENDOFHERE
 
     /bin/arch-chroot /overlay-fs chmod 600 /etc/NetworkManager/system-connections/C15.nmconnection
     /bin/arch-chroot /overlay-fs systemctl enable NetworkManager
+
```
hhoegelo, line 64: Please remove all this dead code

projects/epc/playground/src/device-settings/EpcWifi.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+void EpcWifi::updatePassphrase()
+{
+    SpawnAsyncCommandLine::spawn(std::vector<std::string> { "nmcli", "con", "modify", "C15", "802-11-wireless-security.psk", m_newPassphrase},
+                                [this, p=m_newPassphrase](auto) { m_currentPassphrase = p;
+                                            updateCredentials(true); },
+                                     [this](const std::string& e) { nltools::Log::warning(__FILE__, __LINE__, __PRETTY_FUNCTION__, e);
+                                                                m_busy = false; });
+}
```
hhoegelo, line 75: all this method implementations are quite repetitive, so I'd like to suggest a method like:

```
void EpcWifi::spawn(const std::vector<std::string> &command, std::function<void ()> onSuccess)
{
    SpawnAsyncCommandLine::spawn(command,  onSuccess, [this](const std::string& e) { 
        nltools::Log::warning(__FILE__, __LINE__, __PRETTY_FUNCTION__, e);
        m_busy = false; 
    });
}
```

so that no one can forget resetting m_busy, and invocations are better readable, like

```
void EpcWifi::updateSSID()
{
    spawn({ "nmcli", "con", "modify", "C15", "wifi.ssid", m_newSSID},
        [this, p=m_newSSID](auto) { m_currentSSID = p; updateCredentials(true); });
}
```

projects/epc/playground/src/device-settings/Passphrase.cpp
```diff
@@ -3,8 +3,9 @@
 #include "Passphrase.h"
 #include "Settings.h"
 
-Passphrase::Passphrase(Settings& parent)
+Passphrase::Passphrase(Settings& parent, const std::shared_ptr<EpcWifi>& shrd_ptr)
```
hhoegelo, line 6: shrd_ptr is not a proper name for a variable - how about 'wifi'?

projects/epc/playground/src/device-settings/SSID.h
```diff
@@ -10,14 +11,15 @@ class SSID : public Setting
  public:
   Glib::ustring getDisplayString() const override;
 
-  explicit SSID(Settings& parent);
+  explicit SSID(Settings& parent, const std::shared_ptr<EpcWifi>& shrd_ptr);
```
hhoegelo, line 14: pls rename shrd_ptr so that readers of the interface know what the pointer is useful for

projects/epc/playground/src/device-settings/WifiSetting.cpp
```diff
@@ -36,8 +36,9 @@ std::vector<std::string> getArgs(WifiSettings s)
              "accesspoint;" };
 };
 
-WifiSetting::WifiSetting(UpdateDocumentContributor& settings)
+WifiSetting::WifiSetting(UpdateDocumentContributor& settings, const std::shared_ptr<EpcWifi>& shrd_ptr)
```
hhoegelo, line 39: see above comments regarding name

projects/epc/playground/src/device-settings/WifiSetting.cpp
```diff
@@ -46,6 +47,11 @@ bool WifiSetting::set(tEnum m)
 {
   auto ret = super::set(m);
 
+  if (m == WifiSettings::Enabled)
```
hhoegelo, line 50: consider using boolean logic, e.g.
``` 
m_wifi->setNewWifiState(m == WifiSettings::Enabled);
```

projects/epc/playground/src/device-settings/WifiSetting.h
```diff
@@ -10,7 +11,7 @@ class WifiSetting : public NLEnumSetting<WifiSettings>, public sigc::trackable
 {
  public:
   typedef NLEnumSetting<WifiSettings> super;
-  explicit WifiSetting(UpdateDocumentContributor& settings);
+  explicit WifiSetting(UpdateDocumentContributor& settings, const std::shared_ptr<EpcWifi>& shrd_ptr);
```
hhoegelo, line 14: see above comments regarding name

projects/epc/playground/src/device-settings/WifiSetting.cpp
AntonSchmied: WifiSettings has more than 2 states unfortunately ...
https://github.com/nonlinear-labs-dev/C15/blob/b5ffa01102af0fdcc357812c78e8b5a15d58574e/projects/epc/playground/src/device-settings/WifiSetting.h#L5

hhoegelo (approved):

PR #2607, opened 2021-05-04
added auto start recorder setting

### LINKED ISSUE #2592: Setup: Device Settings: neuer Eintrag: "Auto-Start Recorder" (Enabled/Disabled)
Enabled: der Recorder startet gleich nach dem Booten
Disabled: der Recorder bleibt nach dem Booten inaktiv, kann aber im Recorder-Fenster mit dem "Record"-Button eingeschaltet werden.

### LINKED ISSUE #2593: WebUI: Neuer Eintrag im "View"-Menü: "Open Recorder" (Ctrl-R)
Öffnet den Recorder in einem neuen Tab

schlenkibus:
and new FLAC/Recorder setting page and entry in setup-menu, Enabling this Setting will also start the recorder if disabled, but disabling the setting will not stop the recording as i would see that behavior as not intuitive

closes #2592 
closes #2593

### DIFF
--- a/projects/epc/audio-engine/src/recorder/Recorder.cpp
+++ b/projects/epc/audio-engine/src/recorder/Recorder.cpp
@@ -25,6 +25,14 @@ Recorder::Recorder(int sr)
                                                              [this](const auto &msg) { return api(msg); }))
     , m_http(std::make_unique<NetworkServer>(RECORDER_HTTPSERVER_PORT, m_storage.get()))
 {
+  m_in->setPaused(true);
+  m_settingConnection
+      = nltools::msg::receive<nltools::msg::Setting::StartStopRecorderMessage>(nltools::msg::EndPoint::AudioEngine,
+                                                                               [this](const auto &msg)
+                                                                               {
+                                                                                 if(msg.m_paused == false)
+                                                                                   m_in->setPaused(msg.m_paused);
+                                                                               });
 }
 
 Recorder::~Recorder() = default;
@@ -99,7 +107,8 @@ nlohmann::json Recorder::queryFrames(FrameId begin, FrameId end) const
   std::chrono::system_clock::time_point recordTimeOfLastFrame = invalidTime;
   uint8_t maxOfLastFrame = 0;
 
-  auto cb = [&](const auto &f, auto isLast) {
+  auto cb = [&](const auto &f, auto isLast)
+  {
     bool skipFrame = false;
 
     if(!isLast && recordTimeOfLastFrame != invalidTime)
--- a/projects/epc/audio-engine/src/recorder/Recorder.cpp
+++ b/projects/epc/audio-engine/src/recorder/Recorder.cpp
@@ -27,12 +27,12 @@ Recorder::Recorder(int sr)
 {
   m_in->setPaused(true);
   m_settingConnection
-      = nltools::msg::receive<nltools::msg::Setting::StartStopRecorderMessage>(nltools::msg::EndPoint::AudioEngine,
-                                                                               [this](const auto &msg)
-                                                                               {
-                                                                                 if(msg.m_paused == false)
-                                                                                   m_in->setPaused(msg.m_paused);
-                                                                               });
+      = nltools::msg::receive<nltools::msg::Setting::FlacRecorderAutoStart>(nltools::msg::EndPoint::AudioEngine,
+                                                                            [this](const auto &msg)
+                                                                            {
+                                                                              if(msg.enabled)
+                                                                                m_in->setPaused(false);
+                                                                            });
 }
 
 Recorder::~Recorder() = default;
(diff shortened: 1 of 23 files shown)


projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/presenters/MidiSettingsProvider.java
```diff
@@ -238,6 +241,15 @@ private MidiSettingsProvider() {
             }
             return true;
         });
+
+        s.autoStartRecorder.onChange(t -> {
+            boolean val = t.equals(BooleanValues.on);
+            if(settings.autoStartRecorder.value != val) {
+                settings.enable14BitCC.value = val;
```
hhoegelo, line 248: this doesn't look right

hhoegelo (commented):
Could you please check the format settings? We should agree on the same clang-format file and version so we have less PR changes to review.

projects/epc/audio-engine/src/recorder/Recorder.cpp
```diff
@@ -25,6 +25,14 @@ Recorder::Recorder(int sr)
                                                              [this](const auto &msg) { return api(msg); }))
     , m_http(std::make_unique<NetworkServer>(RECORDER_HTTPSERVER_PORT, m_storage.get()))
 {
+  m_in->setPaused(true);
+  m_settingConnection
+      = nltools::msg::receive<nltools::msg::Setting::StartStopRecorderMessage>(nltools::msg::EndPoint::AudioEngine,
+                                                                               [this](const auto &msg)
+                                                                               {
+                                                                                 if(msg.m_paused == false)
```
hhoegelo, line 33: I think, naming of the feature/message is not what it implements. If StartStopRecorder only does something if !m_paused, it doesn't need the argument at it will actually never "Stop" the recorder. 
So either the StartStop logic here should be complete and the AudioEngineProxy uses it only one-way. Or the Message should be named something like "FlacRecorderAutoStart{ bool enable; }" and the logic here can be as is.

projects/epc/audio-engine/src/recorder/RecorderInput.cpp
```diff
@@ -36,6 +36,18 @@ void RecorderInput::togglePause()
   }
 }
 
+void RecorderInput::setPaused(bool p)
```
hhoegelo, line 39: this is quite duplicating togglePause - I suggest " if (!m_paused) togglePause()" to avoid this.

hhoegelo: Alternatively togglePause could be setPaused(!m_paused);

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -52,41 +53,49 @@ AudioEngineProxy::AudioEngineProxy()
   onConnectionEstablished(EndPoint::AudioEngine,
```
hhoegelo, line 53: If AE crashes or starts after playground has loaded its settings, the FlacAutoStart setting would not be sent to AE. You should send the setting "onConnectionEstablished()", too.

schlenkibus: onConnectionEstablished ruft bereits connectMidiSettingsToAudioEngineMessage() auf, welche sich dann mit dem 'AutoStartRecorderSetting' verbindet. Dort wird ConnectAndInit aufgerufen, denkst du nicht das das reicht?

schlenkibus:
> Could you please check the format settings? We should agree on the same clang-format file and version so we have less PR changes to review.

I Checked my IDE settings, and clang-format is using the project defined .clang-format found at 'C15/.clang-format'. I am not sure what is going on with the formatting

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
hhoegelo: OK, das habe ich nicht gesehen. Sollte funktionieren. Auch wenn ich "AutoStartRecorderSetting" nicht in connect**MidiSettings**ToAudioEngineMessage erwartet haette. Vielleicht faellt uns da ja noch ein treffender Name ein, der ausdrueckt, dass man ganz generell die AE auf den erwarteten Zustand hievt.

projects/shared/nltools/include/nltools/messaging/Message.h
```diff
@@ -193,6 +193,16 @@ namespace nltools
         int m_shift;
       };
 
+      struct FlacRecorderAutoStart
+      {
+        constexpr static MessageType getType()
+        {
+          return MessageType::StartStopRecorder;
```
hhoegelo, line 200: MessageType name and message name are inconsistent

hhoegelo (approved):

PR #2611, opened 2021-05-06
Issue 2596

schlenkibus:
TODO: Specify how the upload should be communicated with the user -> loading screen / progress bar?


projects/epc/playground/src/http/HTTPRequest.cpp
```diff
@@ -24,8 +24,13 @@ HTTPRequest::HTTPRequest(SoupMessage *msg)
     {
       auto header = soup_message_headers_get_one(msg->request_headers, "Content-Type");
       bool containsBlob = header && (g_strcmp0(header, "application/binary") == 0);
+      bool containsTar = header && (g_strcmp0(header, "application/tar.gz") == 0);
 
-      if(!containsBlob)
+      if(containsTar)
```
hhoegelo, line 29: This should probably not be here?

projects/epc/playground/src/http/UpdateContentManager.cpp
```diff
@@ -0,0 +1,39 @@
+#include "UpdateContentManager.h"
```
hhoegelo, line 1: does this class serve any purpose?

hhoegelo (approved):

PR #2635, opened 2021-05-12
added reset to default options for 'classic' and 'high-res' midi

### LINKED ISSUE #2573: Neue Aktion (BOLED-Zeile, GUI-Button) in MIDI Settings für "Set to Defaults" (s.u.)
Vielleicht bieten wir auch zwei Aktionen (BOLED-Zeilen, GUI-Buttons) an:
- "Set to High-Res Defaults" (Pedals, Ribbons und AT auf 14-Bit CCs 20...26) 
- "Set to Classic MIDI Defaults" (7-Bit-CCs 01, CC 07, etc., switching CC64, AT auf Channel Pressure)

schlenkibus:
closes #2573

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/panel-unit/boled/setup/SetupLayout.cpp
+++ b/projects/epc/playground/src/proxies/hwui/panel-unit/boled/setup/SetupLayout.cpp
@@ -94,6 +94,7 @@
 #include <device-settings/flac/AutoStartRecorderSetting.h>
 
 #include <presets/Bank.h>
+#include <use-cases/SettingsUseCases.h>
 
 namespace NavTree
 {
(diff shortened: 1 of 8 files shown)


projects/epc/playground/src/proxies/hwui/panel-unit/boled/setup/SetupLayout.cpp
```diff
@@ -967,6 +968,45 @@ namespace NavTree
     }
   };
 
+  enum class ResetSettingType
+  {
+    HighRes,
+    Classic
+  };
+
+  template <ResetSettingType type> struct ResetMidiSettingsTo : public OneShotEntry
```
hhoegelo, line 977: I don't see a benefit of using a template class here when all methods have to implement switching over the template argument.
I think, being simple and explicit here produces less code and is easier to read:

```
struct ResetMidiSettingsToHighRes : public OneShotEntry
  {

    explicit ResetMidiSettingsToHighRes(InnerNode *parent)
        : OneShotEntry(parent, getName(), []() {
          SettingsUseCases useCases(Application::get().getSettings());
          useCases.setMappingsToHighRes();
        })
    {
    }

    constexpr const char *getName()
    {
      return "Set to High-Res. Defaults";
    }
  };

  struct ResetMidiSettingsToClassic : public OneShotEntry
  {

    explicit ResetMidiSettingsToClassic(InnerNode *parent)
        : OneShotEntry(parent, getName(), []() {
          SettingsUseCases useCases(Application::get().getSettings());
          useCases.setMappingsToClassicMidi();
        })
    {
    }

    constexpr const char *getName()
    {
      return "Set to Classic MIDI Defaults";
    }
  };
```

hhoegelo (approved):

PR #2636, opened 2021-05-12
fix patch position for numSamples

hhoegelo:


MatthiasSeeber (approved):
As far as I can tell, the resulting Flac file now seems valid and will open and play with the shown tools (top: Rhythmbox, Audacity, bottom: VLC, Videos) on Ubuntu 18.04. Although the position is still weird in VLC, the duration seems to work now.

Linux

PR #2652, opened 2021-05-25
Issue 2651

### LINKED ISSUE #2651: Screensaver - Konflikt mit Preset Naming
Matt Petersen (der Apple-Quality-Mitarbeiter) hat folgendes beschrieben

> I have found a rather impactful bug that's reproducible in 21-16.
> 
> Steps:
> 1. Press Store
> 2. Insert mode (I didn't test others)
> 3. Press Enter. Naming screen should appear
> 4. Delete all characters (I only tested this scenario since that's how I 
> found it)
> 5. Wait for screen saver to start (I haven't changed whatever the 
> default time is)
> 6. Press Store to wake (I didn't test others but it was the most 
> intuitive choice when I was unexpectedly faced with the screen saver)
> 
> Result:
> The naming screen is gone. Now, the Preset and Store buttons both flip 
> between Preset and Store, and Inserting a new preset doesn't bring up 
> the naming screen.
> 
> The only way to fix this situation is a reboot.

schlenkibus:
closes #2651


projects/epc/playground/src/proxies/hwui/UsageMode.cpp
```diff
@@ -28,7 +28,7 @@ bool UsageMode::onButtonPressed(Buttons buttonID, ButtonModifiers modifiers, boo
 
 void UsageMode::setupButtonConnection(Buttons buttonID, tAction action)
 {
-  if(buttonID < Buttons::NUM_BUTTONS)
+  if(buttonID < Buttons::NUM_BUTTONS && buttonID != Buttons::FIRST_PARAMETER_BUTTON)
```
hhoegelo, line 31: I don't understand this change - could you please explain why the first button should not be set up?

projects/epc/playground/src/proxies/hwui/panel-unit/PanelUnit.cpp
```diff
@@ -116,14 +117,24 @@ void PanelUnit::onScreenSaverStateChanged(bool state)
 {
   if(state)
   {
-    setUsageMode(new ScreenSaverUsageMode());
+    setOverlayUsageMode(new ScreenSaverUsageMode());
```
hhoegelo, line 120: The BaseUnit does not have this extra feature. I wonder if this was an oversight or by purpose?

projects/epc/playground/src/proxies/hwui/panel-unit/PanelUnit.cpp
```diff
@@ -205,7 +216,11 @@ const EditPanel &PanelUnit::getEditPanel() const
 
 bool PanelUnit::onButtonPressed(Buttons buttonID, ButtonModifiers modifiers, bool state)
 {
-  if(super::onButtonPressed(buttonID, modifiers, state))
+  if(m_overlayUsageMode)
```
hhoegelo, line 219: I am a bit worried about how m_overlayUsageMode makes the HardwareUserInterfaceUnit behaving inconsistent:
onButtonPressed redirects to m_overlayUsageMode, while all other users of 'the usage mode' still refer to the underlying m_usageMode (for example all users of getUsageMode()). 
I cannot layout exactly how this is a problem, but I have the strong feeling, that there should be *one* UsageMode presented by a HWUIUnit at a time.

How about making getUsageMode(s) virtual and return a 'static' OverlayUsageMode whenever the screensaver is active? Or even handle this in the base class? Like:

```
std::shared_ptr<UsageMode> HardwareUserInterfaceUnit::getUsageMode()
{
   static auto overlayUM = std::make_shared<>;
   if(isScreensaverSetting)
      return overlayUM;
   return m_usageMode;
}
```

hhoegelo: while looking at this, I think it is time to remove the silly

```
std::shared_ptr<UsageMode> HardwareUserInterfaceUnit::getUsageMode()
std::shared_ptr<const UsageMode> HardwareUserInterfaceUnit::getUsageMode() const
```

and replace this with

```
[[nodiscard]] std::shared_ptr<UsageMode> getUsageMode() const;
```

projects/epc/playground/src/proxies/hwui/panel-unit/PanelUnit.cpp
schlenkibus: This is on purpose. The BaseUnit does not have have an Overlay-Feature like the PanelUnit has.

projects/epc/playground/src/proxies/hwui/UsageMode.cpp
schlenkibus: agreed, i was confused when i was debugging an crash and misidentified this line to be wrong. But that was obviously not the case instead an missing "init" to the screensaver was the fault of my crash. I also was confused as the FIRST_PARAMETER_BUTTON and BUTTON_0 have the same index value of 0 so for me it made sense to not setup an Button Connection for that 'invalid' button just to be replaced the next iteration with buttonID = 0. If that makes sense.

hhoegelo (approved):

PR #2670, opened 2021-06-01
closes #2669

### LINKED ISSUE #2669: MIDI Send: Wenn der Split Channel nicht auf "None" oder "Common" steht, werden von Single- oder Layer-Sounds die Hardware Sources auf Primary und Secondary Channel gesendet.
Nur im *Split*-Betrieb wollen wir etwas auf dem Secondary (*Split*) Channel senden.

schlenkibus:


MatthiasSeeber (commented):
Well, the behavior is as was intended (now, only in Split Sounds, HW Sources can be sent on separate Secondary) - however, I'm afraid we have to clarify the issue a bit, see https://github.com/nonlinear-labs-dev/C15/issues/2669#issuecomment-852295402

PR #2693, opened 2021-06-10
Issue 2487

hhoegelo:


MatthiasSeeber (approved):
- increments and decrements for hw sources seem correct in both UIs
- auto return is triggered correctly in both UIs (WebUI: MouseUp/KeyUp, HWUI: Focus change)
- boled change detection seems to be working again (no matter how small the changes)

_(As expected, WebUI auto return still produces lots of ParameterChanged messages.)_

schlenkibus (approved):

PR #2699, opened 2021-06-13
issue 2698

### LINKED ISSUE #2698: noten hängen bei split point modulation [bug]
Ein letzter vor meinem Urlaub (2 Wochen Ruhe vor mir ) 

1 Split preset wählen
2 Part 1 spielen und halten
3 Split point so modulieren/per Menu legen , dass Wert unter gehaltenen Noten liegt.
4 Tasten loslassen => Sound läuft weiter

MatthiasSeeber:
- [x] **VoiceAlloc**: refactor `getSplitPartForKey(int key)` into `getSplitPartForKeyDown(int key)` and `getSplitPartForKeyUp(int key, int sourceId)`
- [x] **DSPInterface** (and descendants `dsp_host_dual, MockDSPHost, ...`): refactor `getSplitPartForKey(int key)` into `getSplitPartForKey(int key, InputEventSource from, bool upOrDown)` (or alternatively separate keyDown and keyUp versions)
- [x] **InputEventStage** (`onMIDIEvent, onTCDEvent`): refactor `calculateSplitPartForEvent` in similar fashion (only on KeyDown, the Split Point parameters should ever be used; for KeyUp there now is a separate event flow available)

- runs acceptance-tests and audio-engine-tests without fail
- sound check revealed bug is fixed

closes #2698

### DIFF
--- a/projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
+++ b/projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
@@ -2359,25 +2359,14 @@ SoundType dsp_host_dual::getType()
   return SoundType::Invalid;  // should never be reached
 }
 
-// TODO: refactor
-VoiceGroup dsp_host_dual::getSplitPartForKey(int key)
+VoiceGroup dsp_host_dual::getSplitPartForKeyDown(int key)
 {
-  // also a little inconvenient but should work
-  switch(m_alloc.getSplitPartForKeyDown(key))
-  {
-    case AllocatorId::Local_I:
-      return VoiceGroup::I;
-      break;
-    case AllocatorId::Local_II:
-      return VoiceGroup::II;
-      break;
-    case AllocatorId::Local_Both:
-      return VoiceGroup::Global;
-      break;
-  }
-  // NOTE: this should never be reached and represents an invalid state!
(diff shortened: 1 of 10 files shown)


projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -2359,10 +2359,11 @@ SoundType dsp_host_dual::getType()
...
 
+// TODO: refactor
 VoiceGroup dsp_host_dual::getSplitPartForKey(int key)
 {
   // also a little inconvenient but should work
-  switch(m_alloc.getSplitPartForKey(key))
+  switch(m_alloc.getSplitPartForKeyDown(key))
   {
     case AllocatorId::Local_I:
       return VoiceGroup::I;
```
MatthiasSeeber, line 2369: Here's the goal: this method still calculates the associated Split Part by using the Split Point Parameters. But it shall only apply for a KeyDown Event. For KeyUp Events, a similar method shall provide the Split Part by using instead 

`m_alloc.getSplitPartForKeyUp(key, inputSourceId)`

(see `dsp_host_dual::onKeyDownSplit` on how to determine `uint32_t inputSourceId`)

schlenkibus (approved):

PR #2711, opened 2021-06-23
closes #2704

### LINKED ISSUE #2704: Der Screensaver wird nicht mehr durch Spielen auf der Tastatur beendet.
Wir hatten es schon einmal so implementiert, dass auch das Anschlagen von Noten den Screensaver abschaltet. Das sollte wieder so werden.

### LINKED ISSUE #2734: Note-On/Off has no effect on exiting Screensaver
Apparently it worked before

schlenkibus:
- [x] test on device
- [x] ignore non matching events

also closes #2734


AntonSchmied:
Checked and NoteOn exits the screensaver. Cannot judge the code though ...

My other observation, is that while in Screensaver mode pushing any button does not lead to the parameter of the button. If i touch any MC (a ribbon f.e.) however, the value is recalled straight away. Wouldn't it be more consistent if the buttons behaved the same way?

schlenkibus:
As Henry already saw this code and submitted feedback to me, makes me comfortable enough to merge this PR.
I will think about what you mentioned, if you deem it important enough, open an issue please.

hhoegelo (approved):

PR #2715, opened 2021-06-24
WebUI: changed HW-Source-Enable Radiobuttons to Checkboxes, minor Layout improvements

schlenkibus:
not the whole ticket yet: #2714


MatthiasSeeber (approved):
(The new Checkboxes look a little misaligned as they are not centered within their table cell.)

PR #2725, opened 2021-07-06
Issue 2722

### LINKED ISSUE #2722: "unstable" is unoperational wrt hardware sources, atm
As of 2021-06-29, unstable branch creates code that seems to ignore _hardware sources_ alltogether.
Playcontroller code is working, though (checked with lpc-read) but HW source data does make it neither into AE nor into PG. Keybed is working.

AE service is nagging:
 `Jun 29 13:38:46 archlinux audio-engine[694]: Debug:  http://192.168.10.11:40108  ->  The server did not accept the WebSocket handshake.`
BBB bridge is working.

schlenkibus:
This issue said that no hardware source is working. I suspect that this is an configuration and communication issue rather than an software issue. 
Previously I added more granular MIDI/Local Settings for HW-Sources. They can be currently only found inside the WebUI -> Settings -> MIDI-Tab -> Very Bottom.
I unfortunately made the mistake of defaulting all enable-switches to 'off'. This prevents the changes to propagate thorough the system and it appears as if the Sources were not working. 

This PR fixes the defaults to be true. Also it changes the notifcation system in the InputEventStage to be able to notify either the Playground Sync or the Midi Sync. As I and @MatthiasSeeber  talked about 2 Weeks ago.

preview build can be found here: https://we.tl/t-XPsHETPTRE (this build is one commit behind, but the commit only removed dead code and one missed log-statement)

closes #2722

### DIFF
--- a/projects/epc/audio-engine/src/synth/C15Synth.cpp
+++ b/projects/epc/audio-engine/src/synth/C15Synth.cpp
@@ -15,7 +15,6 @@ C15Synth::C15Synth(AudioEngineOptions* options)
     , m_syncExternalsTask(std::async(std::launch::async, [this] { syncExternalsLoop(); }))
     , m_syncPlaygroundTask(std::async(std::launch::async, [this] {syncPlaygroundLoop(); }))
     , m_inputEventStage { m_dsp.get(), &m_midiOptions, [this] {
-                            nltools::Log::error("syncPG notified");
                             m_syncPlaygroundWaiter.notify_all(); },
                           [this](auto msg) { queueExternalMidiOut(msg); } }
 {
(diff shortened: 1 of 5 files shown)


MatthiasSeeber (changes requested):
After installation, HW Sources were not operable.

The Boolean Matrix initialized to:
Screenshot from 2021-07-06 14-36-56
(only Primary Send was initialized to true)

Further:
- Boolean for PitchBend Receive (Prim and Split) enabled Aftertouch, not Pitchbend
- and vice versa: Boolean for Aftertouch Receive (Prim and Split) enabled Pitchbend, not Aftertouch

schlenkibus:
image
@MatthiasSeeber reicht das erstmal als beweis? Wenn nicht kann ich schauen das ich das hier noch auf einem C15 durchspiele und verifiziere.

PR #2726, opened 2021-07-06
Test hw issues

schlenkibus:


MatthiasSeeber (commented):
UI-wise, the re-introduction of updating HW Return Behavior in the Synth seems solve the primary issue (no moving Mod Amounts on MC Clipping, no Offset when controlling HW Source via MIDI).

However, a Sound Check revealed an issue here, when setting the ReturnBehavior of a Source:

When the UI sends a `HWChangedMessage`, the Synth needs to distinguish between a mere **Position Update** (which should start a Modulation) and a **ReturnBehavior update** (which should only update HW values, not starting a Modulation). Currently, a Modulation is started anytime, causing MC and modulated Param values to jump in Position after setting a connected HW Source to ReturnToCenter.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -2511,3 +2511,13 @@ void dsp_host_dual::onMidiSettingsReceived()
...
+
+void dsp_host_dual::updateBehaviour(const nltools::msg::HWSourceChangedMessage& msg)
+{
+  auto element = getParameter(msg.parameterId);
+  if(element.m_param.m_type == C15::Descriptors::ParameterType::Hardware_Source)
+  {
+    auto param = m_params.get_hw_src(element.m_param.m_index);
+    param->update_behavior(getBehavior(msg.returnMode));
+  }
+}
```
MatthiasSeeber, line 2523: `param->update_behavior` returns a Boolean indicating if the Behavior has changed. This (previously was and) can be used to determine if a Modulation needs to start.

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -345,7 +345,9 @@ void C15Synth::onHWAmountMessage(const nltools::msg::HWAmountChangedMessage& msg
 
 void C15Synth::onHWSourceMessage(const nltools::msg::HWSourceChangedMessage& msg)
 {
-  m_playgroundHwSourceKnownValues[InputEventStage::parameterIDToHWID(msg.parameterId)] = msg.controlPosition;
+  m_dsp->updateBehaviour(msg);
+  auto latchIndex = InputEventStage::parameterIDToHWID(msg.parameterId);
+  m_playgroundHwSourceKnownValues[latchIndex] = static_cast<float>(msg.controlPosition);
   m_inputEventStage.onUIHWSourceMessage(msg);
 }
```
MatthiasSeeber, line 352: If `m_dsp->updateBehavior` would return the mentioned Boolean, we could pass it to `m_inputEventStage.onUIHWSourceMessage` as an additional argument.

MatthiasSeeber: The Boolean would need to be passed down to
``` c++
void dsp_host_dual::onHWChanged(const uint32_t id, float value, const bool returnBehaviorChanged = false)
{
  auto source = m_params.get_hw_src(id);
  const float inc = value - source->m_position;
  source->m_position = value;
  if(returnBehaviorChanged) hwModChain(source, id, inc);
}
```

schlenkibus:
https://we.tl/t-bOwb3vGEu7

MatthiasSeeber (changes requested):
The primary issue seems to be fixed. A Sound Check revealed no further issues.

Unfortunately, a new bug seems to have appeared: now, when a Macro Control enters clipping (from TCD, MIDI, WebUI Edit), the HWUI now shows moving Mod Amount Ranges on modulated Parameters. The WebUI shows the Ranges correctly.

schlenkibus:
last preview build: https://we.tl/t-boPIQVkZY0

MatthiasSeeber (approved):

projects/epc/playground/src/use-cases/MacroControlParameterUseCases.cpp
```diff
@@ -6,7 +6,8 @@
 #include <libundo/undo/Scope.h>
 
 MacroControlParameterUseCases::MacroControlParameterUseCases(MacroControlParameter* parameter)
-    : m_mc { parameter }
+    : ParameterUseCases{parameter}
+    , m_mc { parameter }
```
hhoegelo, line 10: if base class stores parameter, do we need to store it in derived class, too?

projects/epc/playground/src/use-cases/PedalParameterUseCases.cpp
```diff
@@ -4,7 +4,8 @@
 #include <libundo/undo/Scope.h>
 
 PedalParameterUseCases::PedalParameterUseCases(PedalParameter* p)
-    : m_param { p }
+    : PhysicalControlParameterUseCases{p}
+    , m_param { p }
```
hhoegelo, line 8: duplicate m_param

projects/epc/playground/src/use-cases/PhysicalControlParameterUseCases.cpp
```diff
@@ -0,0 +1,12 @@
+#include "PhysicalControlParameterUseCases.h"
+
+PhysicalControlParameterUseCases::PhysicalControlParameterUseCases(PhysicalControlParameter *p)
+    : ParameterUseCases(p)
+    , m_physicalParam { p }
```
hhoegelo, line 5: dup

projects/epc/playground/src/use-cases/RibbonParameterUseCases.cpp
```diff
@@ -4,23 +4,30 @@
 #include <libundo/undo/Scope.h>
 
 RibbonParameterUseCases::RibbonParameterUseCases(RibbonParameter* p)
-    : m_param { p }
+    : PhysicalControlParameterUseCases{p}
+    , m_param { p }
```
hhoegelo, line 8: mimimi

PR #2736, opened 2021-07-13
2730

schlenkibus:
changed css rules to only match with midi-table tagged tables in webUI. Preset and Bank info look just like before now


schlenkibus:
this PR also changes a minor issue mentioned here: https://github.com/nonlinear-labs-dev/C15/issues/2714#issuecomment-876449910

hhoegelo (approved):

PR #2759, opened 2021-07-28
changed z-index of disconnected dialog to appear over all open popups + add modal overlay for ssid-changed dialog

### LINKED ISSUE #2758: Web UI: Setup Window is blocking further actions, when SSID is changed
When I change the SSID, the Setup Window remains in the foreground and blocks further actions. 
Everything that is click-able seems to be dead and the key "F5" has no effect. 
The only way to turn off the Setup Window is by deactivating it within the menu in the top right corner of the Web UI. 
BUT only once the Web UI has successfully reloaded itself (which might take a while). 

Only then does Window "C15 SSID change detected ..." show up and I can use the Web UI as usual.

Tested with an Opera browser

schlenkibus:
closes #2758

### DIFF
--- a/projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/WebSocketConnection.java
+++ b/projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/WebSocketConnection.java
@@ -32,6 +32,7 @@ public class WebSocketConnection {
 
 		public ConnectionErrorPopup() {
 			super(false);
+			getStyleElement().addClassName("dc-dialog");
 			Label l = null;
 			setWidget(l = new Label("The connection to the Nonlinear Instrument seems to be broken :-("));
 			l.getElement().addClassName("dc-text");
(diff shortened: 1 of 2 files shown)


AntonSchmied:
What happens if the Passphrase is changed? shouldn't the behavior be somewhat similar to SSID change?

AntonSchmied:
Tested, works!

AntonSchmied:
Is this dead code, we can get rid of?

https://github.com/nonlinear-labs-dev/C15/blob/df70320d61b1e25c84236446d48c02784624abe1/projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/WebSocketConnection.java#L93

schlenkibus:
@AntonSchmied not quite dead code rather inline `native` javascript: 


> JSNI methods are declared native and contain JavaScript code in a specially formatted comment block between the end of the parameter list and the trailing semicolon. A JSNI comment block begins with the exact token /*-{ and ends with the exact token }-*/. JSNI methods are called just like any normal Java method. They can be static or instance methods.


http://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html#writing

hhoegelo (approved):

PR #2763, opened 2021-07-29
Ipad problems

hhoegelo:


projects/epc/web-ui/recorder/index.html
```diff
@@ -99,6 +101,8 @@
         </div>
     </div>
 
+    <!--<div id="log"></div> -->
```
schlenkibus, line 104: Will that commented-out div serve any purpose in the future?

PR #2766, opened 2021-07-30
editable passphrase

schlenkibus:
- [x] BOLED: added more options in passphrase setup item: edit and explicit default
- [x] WebUI: make passphrase editable

### DIFF
--- a/projects/epc/playground/src/device-settings/Passphrase.cpp
+++ b/projects/epc/playground/src/device-settings/Passphrase.cpp
@@ -13,7 +13,10 @@ Passphrase::~Passphrase() = default;
 
 void Passphrase::load(const Glib::ustring& password, Initiator initiator)
 {
-  updatePassword(password);
+  if(!password.empty())
+  {
+    updatePassword(password);
+  }
 }
 
 Glib::ustring Passphrase::save() const
--- a/projects/epc/playground/src/device-settings/Passphrase.cpp
+++ b/projects/epc/playground/src/device-settings/Passphrase.cpp
@@ -24,6 +24,15 @@ Glib::ustring Passphrase::save() const
   return m_password;
 }
 
+bool Passphrase::isValidPassword(const Glib::ustring& newPassword)
+{
+  if(newPassword.length() >= 8 && newPassword.length() <= 16)
+  {
+    return newPassword.is_ascii();
+  }
+  return false;
+}
+
 void Passphrase::dice()
 {
   static const std::string dict = "abcdefghkmnpqrstuvwxyzABCDEFGHKMNPQRSTUVWXYZ123456789";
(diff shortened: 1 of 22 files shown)


projects/epc/playground/src/device-settings/Passphrase.cpp
```diff
@@ -40,15 +52,21 @@ void Passphrase::resetToDefault()
...
+  if(isValidPassword(password))
+  {
+    if(m_password != password)
+    {
+      m_password = password;
 
-  auto shortened = m_password.size() <= 8 ? m_password : m_password.substr(0, 8);
-  auto passwordMsg = nltools::msg::WiFi::SetWiFiPasswordMessage(shortened);
-  nltools::msg::send(nltools::msg::EndPoint::BeagleBone, passwordMsg);
+      auto shortened = m_password.size() <= 8 ? m_password : m_password.substr(0, 8);
```
hhoegelo, line 61: if we accept 16 chars in 'isValidPassword', we should probably accept 16 here, too?

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
```diff
@@ -307,7 +309,43 @@ public void connectEventHandlers() {
 		pedal4Range.addValueChangeHandler(e -> EditBufferUseCases.get()
 				.setParameterValue(new ParameterId(269, VoiceGroup.Global), e.getValue().doubleValue(), true));
 
+		passphrase.addChangeHandler(new ChangeHandler() {
+			@Override
+			public void onChange(ChangeEvent e) {
+				String text = passphrase.getValue();
+				scheduledCheckPassphrase = true;
+				NonMaps.theMaps.getServerProxy().isPassphraseValid(text, v -> {
```
hhoegelo, line 317: the three copies could probably be refactored into a method

projects/epc/web-ui/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
```diff
@@ -307,7 +309,43 @@ public void connectEventHandlers() {
...
+			scheduledCheckPassphrase = true;
+			NonMaps.theMaps.getServerProxy().isPassphraseValid(passphrase.getValue(), v -> {
+				scheduledCheckPassphrase = false;
+				savePassphrase.setEnabled(v);
+			});
+		});
+
 		saveDeviceName.addClickHandler(e -> settings.setDeviceName(deviceName.getValue()));
+		savePassphrase.addClickHandler(e -> {
+			if(scheduledCheckPassphrase == false) {
```
hhoegelo, line 342: this async race has some smell on it
might eat button clicks - maybe the validity check is not worth the async operation and should be checked in java code?
At least, the Save button should be disabled when there is an async check in progress, so it can n ot happen that user presses a button and just nothing happens.

PR #2769, opened 2021-08-05
Just removed all the z-index stuff and dit not found any flaws 

hhoegelo:
@schlenkibus The preset rename dialog did not work anymore - it was below the glass. So I just removed all the z-indexes and did not find issues but probably I am missing something? Please test what you had in mind when doing the initial change.


schlenkibus:
z-index-not-working
I can see that the bug I was trying to fix is no longer fixed. The Setup-Dialog is shown above the modal-glass-pane, which we do not want. We only want the SSID Changed Dialog to appear above the blackout. 
While debugging I found that the Setup Dialog had an z-index of 1004, eventhough the SetupDialog did not call pushToFront or has an explicit z-index in the css.

schlenkibus (changes requested):
see my top level comment

PR #2777, opened 2021-08-19
remove maximum upload for posted data, so banks with >10 MB can be im…

hhoegelo:
…ported


hhoegelo:
Note: This fix will solve the (import) problem @majam found, but not the original (export) problem @nonlinear-labs had.

schlenkibus:
closes #2779

PR #2778, opened 2021-08-19
Test websync

schlenkibus:


projects/epc/playground/src/parameters/Parameter.cpp
```diff
@@ -44,6 +45,10 @@ Parameter::Parameter(ParameterGroup *group, ParameterId id, const ScaleConverter
 Parameter::~Parameter()
 {
 }
+nlohmann::json Parameter::serialize() const
+{
+  return { { "id", getID().toString() }, { "name", getLongName().c_str() }, { "value", getDisplayString().c_str() } };
```
hhoegelo, line 50: c_str() is not needed

hhoegelo: (Maybe you have to include adl serializers)

projects/epc/playground/src/presets/EditBuffer.cpp
```diff
@@ -300,6 +302,16 @@ bool EditBuffer::isParameterFocusLocked() const
 {
   return m_lockParameterFocusChanges;
 }
+namespace nlohmann
+{
+  template <> struct adl_serializer<ParameterGroup *>
+  {
+    static void to_json(json &j, const ParameterGroup *v)
+    {
+      j = v->getID().toString();
```
hhoegelo, line 311: if you include the adl serializers, you do not need to call toString() anymore

hhoegelo (approved):

PR #2781, opened 2021-08-25
Splitpoint-Preset Conversion

schlenkibus:
Split-Parameters are now loaded into Part I and II upon reading the XML. This will also bump the Software-Version to 10.


projects/epc/playground/src/serialization/PresetParameterGroupSerializer.cpp
```diff
@@ -21,28 +28,56 @@ Glib::ustring PresetParameterGroupSerializer::getTagName()
...
 
-      auto param = std::make_unique<PresetParameter>(id);
-      auto ret = new PresetParameterSerializer(param.get(), m_type);
-      m_paramGroup->m_parameters.push_back(std::move(param));
-      return ret;
-    });
+                   std::vector<PresetParameter *> params;
+                   for(auto &g : m_paramGroups)
+                     for(auto &p : g->m_parameters)
+                       params.emplace_back(p.get());
```
hhoegelo, line 64: this looks wrong to me: I think we should add only parameters with matching id to the vector. Vector should have max 2 entries. Also I think, the code can be simplified as the algorithm can be implemented to work on a vector of one or two entries.

projects/epc/playground/src/xml/Reader.cpp
```diff
@@ -139,3 +139,17 @@ sigc::connection Reader::onFileVersionRead(sigc::slot<FileVersionCheckResult, in
 {
   return m_sigFileVersionRead.connect(slot);
 }
+
```
hhoegelo, line 142: Do we still need this? I don't think it is used.

projects/epc/playground/src/serialization/PresetParameterGroupSerializer.cpp
```diff
@@ -2,17 +2,24 @@
 #include "PresetParameterSerializer.h"
 #include <presets/PresetParameterGroup.h>
 #include <presets/PresetParameter.h>
+#include <parameter_declarations.h>
+#include <nltools/logging/Log.h>
+#include <utility>
+#include <nltools/Assert.h>
 
-PresetParameterGroupSerializer::PresetParameterGroupSerializer(PresetParameterGroup *paramGroup, SoundType type)
+// change this signature and following serializers to accept a list of groups/parameters
```
hhoegelo, line 10: do we still need the comment here?

hhoegelo (approved):
Its only the useless comment that should be removed, otherwise accepted.

PR #2793, opened 2021-09-01
BBB without SD-Card solution

### LINKED ISSUE #1712: BBB: SD dependency
We should focus on deploying "rootfs" on the internal flash of the BBB and cut the SD dependency.
We should also discuss whether, we should maintain the extra partition, which allows to boot from USB-Stick and how it can be part of the new BBB image. 

### current state of affairs (2020-10-01 KSTR)
- booting from internal flash (eMMC) is working fine and the setup to do so (clone the system from SD-card to flash and make it bootable) is actually quite simple.
- a problem pops up when we want to boot from eMMC when an SD-card is still inserted. It turns out that the low-level bootloaders cannot handle an inserted SD-card which does *not* have a proper linux boot config, that is no /boot/uImage file in a (linux) file system on any of its partitions. The problem only shows up with out proprietary bootloader, the stock bootloader simply ignores 'invalid' SD-card content. To fix this we'd have to start here: https://github.com/nonlinear-labs-dev/u-boot/commit/3fde2563258ae1f488ee737168f69f2c7cf51e61
- I also found a lot of confusion, many files both on the SD-card and the eMMC that look to responsible for booting and loading our device tree actually are not, etc. Looks like the only relevant information is *hardcoded* in the second stage bootloader (u-boot.img) which resides in the partition section of the eMMC, see above.

AntonSchmied:
closes #1712 

So this PR has two main Goals:

Create an SD Card Installer for the BBB with the new Bootloader
(IMPORTANT: This currently only works with Nonlinux add_mke2fs Branch and Uboot emmc-solo-boot-test Branch. Once everything is clean all 3 PRs have to be merged at the same time, with little adjustments such as Repo pointers.)

Add functionality to update the C15s with the new bootloader and sync the rootfs from SD Card to the eMMc.

TODOs:
- [x] MLO chekcsums don't seem to be the same after install 
- [x] Getting eMMC and Partition Info should only happen once
- [x] exclude /usr/C15/scripts/calibration from update overwrite


build-tools/create-c15-update/create-c15-update.sh
```diff
@@ -171,15 +181,15 @@ create_update_tar () {
...
-calc_checksum() {
-    echo "Calc checksum..."
-    if (cd $OUT_DIRECTORY/ && touch $(sha256sum ./nonlinear-c15-update.tar | grep -o "^[^ ]*").sign); then
-        echo "Calc checksum done."
-        return 0
-    fi
-    echo "Calc checksum failed."
-    return 1
-}
+#calc_checksum() {
```
hhoegelo, line 184: can this dead code be removed?

build-tools/create-c15-update/create-c15-update.sh
```diff
@@ -3,14 +3,16 @@
...
 
-EPC_UPDATE=$1
-EPC_2_UPDATE=$2
-BBB_UPDATE=$3
-PLAYCONTROLLER_UPDATE=$4
-BINARY_DIR=$5
-SOURCE_DIR=$6/build-tools/create-c15-update
-OUTNAME=$7
-ASPECTS=$8
+EPC_UPDATE=$1 && shift
```
hhoegelo, line 6: This will probably break other callers (see epc-update, playcontroller-update etc)

build-tools/create-c15-update/update_scripts/bbb_update.sh
```diff
@@ -92,13 +96,63 @@ update(){
...
+    printf "eMMC partition for rootfs is $EMMC_DEVICE_P1\n"
+    return 0
+}
+
+check_bootloader() {
+    get_emmc_device
+
+    dd if=${EMMC_DEVICE} of=/tmp/u-boot.img.dd bs=512 skip=768 count=1024 conv=notrunc
+    truncate -s $(wc -c < /update/BBB/u-boot.img) /tmp/u-boot.img.dd
+    [ $(md5sum /tmp/u-boot.img.dd | cut -d' ' -f1) != $(cat /update/BBB/UBOOT_sum) ] && return 1
```
hhoegelo, line 113: nitpick: [ a == b ] || return 1
easier to read (at least for me) as [ a != b ] && return 1

hhoegelo (approved):

PR #2810, opened 2021-09-16
Simplify bbb build tool

hhoegelo:
have privilege problems anymore
- do not anymore build in a ext4-container, but use the hosts
  file system instead to reduce host dependencies


build-tools/bbb/runInDocker.sh
```diff
@@ -1,35 +1,39 @@
 #!/bin/sh
 
+set -e
```
schlenkibus, line 3: do you want to keep the verbosity?

schlenkibus (approved):

build-tools/bbb/runInDocker.sh
hhoegelo: That's not regarding verbosity, but to stop on first error. 'd like to keep it.

schlenkibus:
- [x] build okay
- [x] component test okay

PR #2813, opened 2021-09-21
closes #2798

### LINKED ISSUE #2798: Voice-Allocation: Held note + setting Send Notes to off, leads to hung note in externals
we should send all notes/all sound off message whenever the note send settings have changed.

schlenkibus:
Stuck external notes should be fixed by sending CC 123 'All Notes Off' on both channels (if split is applicable) if the sending of MIDI notes is turned off

### DIFF
--- a/projects/epc/acceptance-tests/src/MIDI/Issue2798.cpp
+++ b/projects/epc/acceptance-tests/src/MIDI/Issue2798.cpp
@@ -17,7 +17,8 @@ TEST_CASE("When Notes Local is turned off All Notes off gets send", "[MIDI][TCD]
   msg.sendChannel = MidiSendChannel::CH_1;
   msg.sendSplitChannel = MidiSendChannelSplit::CH_2;
   options.update(msg);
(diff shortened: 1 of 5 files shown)


schlenkibus:
@davidm2894 Preview Build is coming soon!

schlenkibus:
@davidm2894 here we go: https://we.tl/t-4IjP33fa3a

davidm2894:
no more stuck notes in split mode and whenever changing the routings. however, we still have stuck notes, when changing primary or split channel and holding down a key simultaneously.

schlenkibus:
I didn't consider that. Thanks for pointing it out!

schlenkibus:
new preview: https://we.tl/t-LZGNOrMF5J

hhoegelo (approved):

davidm2894:
I still have stuck notes when changing send channel in both single and split mode. In split mode it doesn't matter if I change the channel wether of the primary or the split channel and holding down a key in the same channel, it will stuck either way.

Besides, I noticed that the "Global Local Enable" has lost its functionality.

PR #2823, opened 2021-09-23
closes issue #2817

schlenkibus:


schlenkibus:
preview build: https://we.tl/t-RDE6saF6UA

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -465,23 +465,29 @@ void C15Synth::onTuneReferenceMessage(const nltools::msg::Setting::TuneReference
...
-  const auto didSplitChange = oldSplitChannel != newSplitChannel;
-
+  InputEventStage::OldSettingSnapshot snap(m_midiOptions);
   m_midiOptions.update(msg);
-
-  m_inputEventStage.handlePressedNotesOnMidiSettingsChanged(msg, oldPrimSendState, oldSecSendState, didPrimChange,
-                                                            didSplitChange, oldPrimChannel, oldSplitChannel);
+  m_inputEventStage.handlePressedNotesOnMidiSettingsChanged(msg, snap);
   m_dsp->onMidiSettingsReceived();
 }
```
MatthiasSeeber, line 472: It seems to me that `handlePressedNotesOnMidiSettingsChanged` may trigger an AllNotesOff Event (which will reset the synth completely - voice alloc, envelopes, buffers). However, in any case `onMidiSettingsReceived` will trigger (possibly another) Reset Event (voice alloc, envelopes).

MatthiasSeeber (changes requested):
- [x] Global Local Enable has been renamed (see comment below for a suggestion)
- [x] the Local Enable option still operates with the first iteration of the event flow (the UI-parameters are detached from changes when Local Enable is OFF, though MIDI is sent) - a topic for later
- [x] Panic also sends AllNotesOff (after Panic, a previously held key will still send a NoteOff message when released - I'm not sure if this is a problem or not?)
- [x] AllNotesOff messages seem to be only sent when necessary (Panic, Local Enable = Off, Change in Send Primary Channel, Change in Send Split Channel when in Split Sound, Note Send Routing = Off, Note Local Routing = Off)
- [ ] Reset behavior seems a little odd:
- I cannot observe Reset Messages for PitchBend / Aftertouch when turned to OFF
- likely, every Setting message will always reset the voice allocation and envelopes (as we implemented a while ago - but not matching the ticket requirements)
- some messages (Local Enable, Receive Channels, Note Receive/Local Routings) also reset everything (like Panic does) - which is not necessary and seems to produce glitchy behavior when keys are held (I strongly suspect, that two fade/reset tasks can be fired within one setting message, see second comment)

schlenkibus:
New Preview: https://we.tl/t-mzxIjaoQj5

davidm2894:
bei geladenen Split Sound und Local Enable OFF, wird nur Note On, aber nicht Note Off geschickt

projects/epc/audio-engine/src/synth/c15-audio-engine/key_event.h
```diff
@@ -106,19 +109,22 @@ template <uint32_t From, uint32_t To> class ShifteableKeys
   // note: this could be the right place for key remapping as well (if we decide to implement it)
 
  public:
+  //63
   int32_t keyDown(const uint32_t _keyPos)
   {
     if((_keyPos >= From) && (_keyPos <= To))
     {
       const uint32_t keyIdx = _keyPos - From;
+      nltools_assertAlways(keyIdx < Keys);
```
hhoegelo, line 118: pls remove

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
```diff
@@ -331,6 +343,9 @@ void InputEventStage::sendKeyUpAsMidi(TCDDecoder *pDecoder, const VoiceGroup &de
   constexpr const uint8_t keyType = 0x80;
   constexpr const uint8_t ccType = 0xB0;
 
+  nltools::Log::error(__PRETTY_FUNCTION__, "determinedPart:", toString(determinedPart), "primChannel:", (int)mainC,
```
hhoegelo, line 346: pls clean up all the traces

PR #2862, opened 2021-10-13
experiment: add rendered samples display in boled

schlenkibus:


projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -77,6 +77,12 @@ AudioEngineProxy::AudioEngineProxy(PresetManager &pm, Settings &settings, Playco
                                                           useCases.setGlobalLocal(msg.m_state);
                                                         });
 
+  receive<LastRenderedFrames>(EndPoint::Playground,
+                              [=](const auto &msg) {
+                                m_lastReceivedFrames = msg.frames;
```
schlenkibus, line 82: maybe i should be syncronizing the access to m_lastReceivedFrames as getData() can be called from anywhere anytime

schlenkibus:
@nonlinear-labs is that a feature we want to have? If not, lets close this PR.
 I remember that the only feedback you gave was: that we wouldn't need to show the left and right samples. But I don't recall what the preferred alternative was. Or if you wanted to introduce this kind of feature at all. Some feedback would be great.
 If you want a preview-build again to know what this was about, let me know.

nonlinear-labs:
@schlenkibus - I can't remember very much about this potential feature. It would be the basis for an oscilloscope for the output signal, right? At the moment I would not give it a priority. More important would be a level-display or a overload indicator. This would require some DSP in the audio engine and a low-bandwidth connection to the UI.

PR #2897, opened 2021-11-04
fix issue #2892

schlenkibus:


projects/epc/playground/src/use-cases/EditBufferUseCases.h
```diff
@@ -70,6 +70,8 @@ class EditBufferUseCases
   void randomize(double amount);
   void randomizePart(VoiceGroup part, double amount);
 
+  void setName(const Glib::ustring& string);
```
hhoegelo, line 73: Is this method actually used somewhere?

schlenkibus: I removed the unused code path

hhoegelo (approved):

PR #2907, opened 2021-11-09
closes #2831

### LINKED ISSUE #2831: MIDI: Single -> Convert to Split -> kein All Notes Off
Haben auch auf diesem Weg noch einen Notenhänger, bei gedrückter Note und gleichzeitigem Convert Aktion. Es kommt gar kein Note Off, also weder auf Part 1 noch auf 2.

schlenkibus:


projects/shared/nltools/include/nltools/messaging/Message.h
```diff
@@ -238,6 +238,14 @@ namespace nltools
       int32_t framesPerPeriod = 0;
     };
 
+    struct ConvertedMessage
```
hhoegelo, line 241: Let's try to find a better name for this message. 'ConvertedMessage' reads like there is a message that has been converted. How about something with 'Notify' or "Notification" - maybe 'SoundConvertedNotification' or "SoundConversionNotification' or "NotifySoundConverted" or something? (Of course, all the methods dealing with this renamed message have to be renamed, too).

MatthiasSeeber: After having discussed this with @schlenkibus I've provided another approach in #2909 - no Notifications will be needed as the Reset Detection is implemented within the synth, using already existing logic.

MatthiasSeeber (approved):
Installer Test was successful.

PR #2927, opened 2021-11-16
Issue 2925

schlenkibus:


projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -2634,3 +2634,34 @@ bool dsp_host_dual::resetIsNecessary()
...
+      break;
+    case VoiceGroup::Invalid:
+      break;
+  }
+
+  // having summed up all voices, we obtain the number of "active" Voices (0 ... 24)
+  // the obtained value should be sample-precise and is not dependant from Sustain or Level Parameters
+
+  return static_cast<uint8_t>(gateSignal);  // 0 : no voice active / 1, 2, ... 23, 24 : n voices active
+}
```
MatthiasSeeber, line 2667: I would like to advise against that approach.
1. Although this might work in most cases and appears to theoretically produce the desired behavior, we would introduce a new condition (to set the gate release time in stone and never change it). Further, we would settle for tolerating a small statistical chance for this to fail, as the gate release time is almost, but not exactly zero.
2. The operation seems to originate from Acceptance Tests, in which we don't really care about CPU consumption for test helpers. But i don't consider this production-ready code. Summing up all voices is SIMD, so the CPU cost is probably not too much. But still, I don't see a particular benefit here.
3. If an unepxected `VoiceGroup::Invalid` is received, an unitiialized float will be used for the return value.
4. The previous implementations of Reset functionality for Midi Setting changes already established an event-based mechanism for detecting active voices within the InputEventStage. I would prefer using that approach as a base for new preset-and parameter-change- based reset enhancements.

MatthiasSeeber (changes requested):

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
schlenkibus: I understand that using the gate signal is not advised. I changed the implementation to use the prior available `isResetNecessary()`.

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.cpp
```diff
@@ -1677,7 +1677,7 @@ DSPInterface::OutputResetEventSource dsp_host_dual::recallSingle(const nltools::
   // update unison and mono groups
   auto polyChanged = evalPolyChg(C15::Properties::LayerId::I, msg->unison.unisonVoices, msg->mono.monoEnable);
   // reset detection
-  const bool resetDetected = layerChanged || polyChanged;
+  const bool resetDetected = (layerChanged || polyChanged) && resetIsNecessary();
```
hhoegelo, line 1680: I am absolutely not involved in the topic, but just from the wording I suspect a smell here:
If something is necessary, I would expect it to cause an action, so either there should be an "||" instead of the "&&"' or it should read "resetIsPossible()"?
Please ignore this comment if it doesn't make sense after consideration.

PR #2932, opened 2021-11-17
part aware reset behavior - addendum

MatthiasSeeber:
I spotted a flaw in the reset behavior when recalling Split Sounds.
As it turns out, the logic behind the reset detection gets rather complicated.

Here is an attempt to describe the current behavior:

1. LayerChanged: Has the Sound Type changed? (if yes: case non-split -> split, otherwise: case split -> split) (the voice allocation will reset if LayerChanfed)
2. Are keys pressed (depending on the old sound type)?
3. PolyChanged: While loading the parts: did Unison or Mono change in a part? (if yes, the voice allocation will reset for that part)
4. After loading the parameters: did the Sound Type change? (if yes: return outputResetEvent according to 2., otherwise: combine pressed keys with PolyChanged reflecting the parts)


MatthiasSeeber:
`All tests passed (33163452 assertions in 52 test cases)`

hhoegelo (approved):

PR #2953, opened 2021-11-30
Fix webui update installation failure

schlenkibus:
WebUI Updates from 33 to 47 stopped in the middle of the installation. 
USB updates and webui updates where the final script that is called was started by hand where fine.

Failed Updates resulted in broken rootfs on bbb and the devices are unable to start the install-update service on later boots, as well as no UI, only the splashscreen is shown.

The Problem seems to have been the install-update-from-epc-oneshot.service which did exactly as named: the service was a oneshot service which means systemd waits for the return of the script before declaring the service as active. Per default the Timeouts for oneshot services are set to 30 seconds. Now that the service is declared as simple, systemd will return after forking and our update can go through.

In order to prevent updates to be installed on vulnerable versions our run.sh now includes an early exit if following checks are true:
the install-from-epc service is declared as oneshot
the epc:/tmp contains an nonlinear-update.tar -> means the current update was started via the webui


schlenkibus:
I did not yet test the latest changes on the device.

build-tools/create-c15-update/Installer_Error_Codes.txt
```diff
@@ -37,6 +37,7 @@ List of current Error Messages
 - 86 :: "E86: ePC update missing" "Please retry download!"
 - 87 :: "E87: BBB update missing" "Please retry download!"
 - 88 :: "E88: playcontroller update missing" "Please retry download!"
+- 89 :: "E89: WebUI update forbidden" "Use USB Update!"
```
AntonSchmied, line 41: Just for consistency
 - 89 :: "Update aborted! E89" "Please use USB update!"

build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -239,7 +241,44 @@ stop_services() {
...
+is_any_install_from_epc_service_oneshot() {
+  serviceFile='/usr/lib/systemd/system/install-update-from-epc.service'
+  serviceFileOneshot='/usr/lib/systemd/system/install-update-from-epc-oneshot.service'
+
+  if [[ -f $serviceFile ]]; then
+    cat $serviceFile | grep oneshot
+    ret=$?
+    if ret -eq 0; then
+      return ret
+    fi
```
AntonSchmied, line 258: I assume `return $?` would be enough here too like a couple of lines below?

AntonSchmied (commented):
I will build an update with the fix today and see how well it works.

AntonSchmied:
- The check catches an initialized update from WebUI well and the C15 prompts the User to use the USB as expected!

However, the PR is based off "Main 21-33" ... 
If i am not mistaken, we agreed to branch it of 21-47?

PR #2954, opened 2021-11-30
issue 2951

MatthiasSeeber:
The FB Mix Shaper Asym possesses a 30 Hz HP filter that is only present in the "asym-wet" part of the signal. With high FB Mix Settings (see provided bank in #2951) the Synth can quickly run into NaN.

Todo:
- [x] Proposal 1: 30 Hz HP after Asym dry/wet crossfade: compromises sound, doesn't fix NaNs
- [ ] explore alternative solutions (like a separate hp for the "asym-dry" branch, clipping/saturation?)

In addition, we may want to explore ways to avoid/clip NaN occurences within the synth similar to the denormal handling (`_mm_setcsr(_mm_getcsr() | (1 << 15) | (1 << 6));`) - if possible. This way we may be able to run the Synth exception-free?


schlenkibus:
What is the current status on this PR? @MatthiasSeeber
The ticket in question was: #2951
@hhoegelo  High CPU load sounds rather critical. And the reported UI freezing sounds also very alarming.

MatthiasSeeber:
> What is the current status on this PR? @MatthiasSeeber The ticket in question was: #2951 @hhoegelo High CPU load sounds rather critical. And the reported UI freezing sounds also very alarming.

Unfortunately, the status hasn't changed much.. there still is a need for exploring the issue/possible solutions further. However, the situation may sound more critical than it actually is: The required Presets (see Ticket) are edge cases (with extreme FB Mixer settings), there only are a few known experimental sounds to produce the NaN behavior. The Factory Library doesn't include these. If I remember correctly, the CPU load and possibly UI freezing was caused by (possibly loads of) Exceptions.

As there is no solution present at the moment, and the PR being stale, it may be better to close it?

PR #2955, opened 2021-12-01
closes #2926

### LINKED ISSUE #2926: Presets: Nach Speichern mit Overwrite, springt der Selektor zum Preset, das bei Beginn des Store-Mechanismus ausgewählt war
https://user-images.githubusercontent.com/84857645/141982965-9e3a0ec9-bfd0-4fb4-bfd4-aeb60515d3fd.mov

Es sollte stattdessen das gespeicherte Preset selektiert sein.

schlenkibus:


projects/epc/playground/src/use-cases/PresetUseCases.cpp
```diff
@@ -16,7 +16,16 @@ PresetUseCases::PresetUseCases(Preset* p, Settings& settings)
...
     , m_settings{ settings }
 {
+  if(auto bank = dynamic_cast<Bank*>(m_preset->getParent()))
+  {
+    m_bank = bank;
+    m_presetManager = bank->getPresetManager();
+    m_editBuffer = m_presetManager->getEditBuffer();
+  }
   nltools_assertAlways(m_preset != nullptr);
+  nltools_assertAlways(m_bank != nullptr);
```
hhoegelo, line 26: The if() above mismatches with the assertion. Either you expect m_bank etc to be set (then you don't need the if()) or you can handle no bank, then the assertions are wrong.

projects/epc/playground/src/use-cases/PresetUseCases.h
```diff
@@ -24,5 +25,8 @@ class PresetUseCases
 
  private:
   Preset* m_preset;
```
hhoegelo, line 27: preset should be reference (if nullptr is not supported)
ptrs should be initialized

PR #2975, opened 2021-12-23
Update bbb_update.sh

AntonSchmied:
Allow rsync tell more about its actions


hhoegelo (commented):
So far it is not clear that we need this: Currently I am trying to reduce update package to normal size and fix the x-runs without kernel update. The update time might be too short to get a snack then.

AntonSchmied:
> So far it is not clear that we need this: Currently I am trying to reduce update package to normal size and fix the x-runs without kernel update. The update time might be too short to get a snack then.

This has nothing to do with the ePC part of the update. This only alerts the user when we sync the eMMC with the SD card on the BBB and udpate the bootloader to boot from the eMMC.

hhoegelo (approved):

PR #2997, opened 2022-01-11
Issue 2993

schlenkibus:


projects/web/static/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/parameters/SendCCDisplay.java
```diff
@@ -0,0 +1,131 @@
...
+
+public class SendCCDisplay extends Label {
+    public SendCCDisplay(OverlayLayout parent) {
+        super(parent);
+        EditBufferModel.get().selectedParameter.onChange(sel -> {
+            onParameterSelectionChanged(sel);
+            return true;
+        });
+
+        SetupModel.get().systemSettings.benderMapping.onChange(e -> {
```
hhoegelo, line 21: Consider moving the logic of where to subscribe and how to combine model data into a displayString into presenter providers. I don't really understand the feature so I can not tell whether parameter presenter or device settings presenter would be appropriate.

hhoegelo (approved):

PR #3031, opened 2022-02-08
closes #2873. First check if Apple Audio Adapter is present. Select t…

AntonSchmied:
…he apropriate audio-card and set Volume to max


KlausStrohhaecker:
FWIW, checking the existence of /proc/asound/PCH/id does not actually test for an existing on-board _analog_ output. PCH(PlatformControlHub)-connected audio devices also exist with HDMI-only digital output version.
Not that this would be areal issue for us, that is.

PR #3040, opened 2022-02-10
BOLED: removed HW-Source arrow, moved CC Label, truncate display String for Channel Pressure to Ch. Press.

### LINKED ISSUE #2973: Kosmetik: Getrennte HW-Sources, CC-Nummern + Undo-Pfeile
Problem:
boled-dummy3

Mein Vorschlag wäre, die Schriftgröße der CC-Nummern an die der Parameter-Werte anzupassen (7 Pixel Höhe, Emphase-8) und 5 Pixel unterhalb des Pfeils anordnen, dann schließt der CC Text an den Parameter-Wert ab (1 Pixel Überlappung). Das sähe auch ohne Pfeil in Ordnung aus:
Slide 16_9 - 3
Slide 16_9 - 6

schlenkibus:
…
closes #2973

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/controls/SelectedParameterValue.cpp
+++ b/projects/epc/playground/src/proxies/hwui/controls/SelectedParameterValue.cpp
@@ -185,7 +185,11 @@ void PhysicalControlValueLabel::onHWChanged(const Parameter *p)
 }
 
 HardwareSourceCCLabel::HardwareSourceCCLabel(const Rect &e)
-    : Label(e)
+    : LabelStyleable(e,
+                     { .size = FontSize::Size8,
+                       .decoration = FontDecoration::Regular,
+                       .justification = Font::Justification::Left,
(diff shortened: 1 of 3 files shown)


MatthiasSeeber (approved):
Boled-Fri-Feb-11-15:02:54-2022
Boled-Fri-Feb-11-15:04:39-2022
Labels and Arrows are looking good, but I don't get the meaning of "removed HW-Source arrow" (also mentioned in the ticket).
Boled-Fri-Feb-11-15:12:51-2022
When I select between "Send"/"Receive", this will be an undoable event. Regardless of which item is currently selected, changing the slider by Inc/Dec can be undone and shows an arrow (which is correct i.m.o.). Undoing the selection will also indicate an undo arrow (seems also correct to me). Am I missing something?

schlenkibus:
Regarding the "HW-Source Arrow", we used to have an dotted line on the left half of the Sliders if the Send-Parameter was seleteced. This arrow has been removed.

PR #3044, opened 2022-02-11
Removed HW Source Recall in AudioEngine

MatthiasSeeber:
Recalled Presets will not have any effect on the State of a Hardware Source, which is only determined by Realtime MIDI/LPC or UI ParameterChanged Events.
Further, the Hardware Sources can be initialized - before the initial Preset is transmitted.


MatthiasSeeber:
A discussion about this revealed that there is no real need to change the current implementation. In a Preset Message, it is guaranteet that all values (even HW Sources) reflect the latest state.

Issue #3030, opened 2022-02-07
ePC: add S24_LE support for 24bit resolution


AntonSchmied:
If we want to use external usb audio, especially the Apple USB-C Adapter, we need to offer 24bit resolution.

AntonSchmied:
true! However, i wanted to specifically point out what has to be done, since this is only mentioned in a comment.

PR #3049, opened 2022-02-17
issue #3030: add 24 bit support

hhoegelo:
support 24 bit audio as needed for external usb audio cards
and prefer 24 bit over 16 bit audio


MatthiasSeeber (commented):
I am not quite familiar with the code, but the changes make sense to me and I couldn't find anything questionable. The only thing I noticed is that there doesn't seem to be any Dithering for 16 or 24 Bit modes - does ALSA handle that or is this another subject?.

hhoegelo:
> I am not quite familiar with the code, but the changes make sense to me and I couldn't find anything questionable. The only thing I noticed is that there doesn't seem to be any Dithering for 16 or 24 Bit modes - ~does ALSA handle that or~ is this another subject?

To be honest, I don't know if dithering should be done when converting floats (24 bit mantissa) to 24 bit int - so far I never did this in any of the projects I maintained and I can not say it the conversion from log resolution to linear resolution would require it.

@nonlinear-labs and @KlausStrohhaecker, what do you think?

AntonSchmied:
I will only be able to test it with 10i3+USB Audio next week. I don't mind it being merged into milestone-sp-a3 though

schlenkibus:
I only approved the possible implementation based on its code, not with the question in mind if dithering is needed or not.

KlausStrohhaecker:
IME dithering to just below the analog noise level always is useful, or at least not harmful at all.
The Apple adapter has a "no signal" muting function that can only be avoided when dithering to ~20bits.

And theoretically all conversion from float/double to int must be dithered to the given resolution when using computed/noise-free signals.
Simple dither will do, rectangular and no noise shaping. Basically adding a few bits of white noise (with clean zero RMS value, and saturating sample value range of course), which is computationally very benign. A per-sample random value will be available in the system anyway (I would assume so at least), and using the same dither value for L/R channels is OK.

----------

My proposal would be:
- dither the actual DA converter stream to 20bits for given hardware resolutions greater than 16bit, and to 16bits with 16bit hardware in case it is ever used.
- dither the stream sent to digital recorder etc to 24bits (to comply to digital audio hygiene rules ;-). 

In general, the whole issue is only critical for 16bit hardware which is never what we want to install in the actual product so we might skip the whole idea, I won't be offended.
With the Apple dongle dithering would remove the muting... of course the muting may actually help zero-signal noise (measured it does but I found it effectively audible as it happens at or below our total system noise floor which is not stellar anyway).

OTOH, for the recorder digital signal, full featured dithering (with noise-shaping) can easily add another 4bits of apparent resolution with negligible noise increase in the critical audio passband, notably 3kHz region. This would help later editing, say if someone found the recording level was way too low and adds 20..30dB of boost for snippets they want to re-use, dto with heavy downward compression and spot EQ.

KlausStrohhaecker:
@hhoegelo,
Code looks very indirect and unclear/bloated to me.

Simple conversion without dithering is always the same for int16, int24 and int32. You convert always to int32 and use only the upper 16 bit of that int32, or 24bit respectively.

Spitting int24 values into a struct with 3 uint8's (where we don't even know/specify that it should be packed struct/array) looks awkward.

The core question in the end is which memory respresentation ALSA is actually using for buffering. I would bet it uses 32bits words for 24 bit values just the same as it would for 32bit values. Anything else doesn't make any sense as there are no native 24bit machines. I would have assumed the only point where the 24bit spec is ever used is the moment the driver packs up USB packets where the 24bit format is finally used to save some USB bandwidth.

But maybe I'm all wrong and ALSA actually requires one to use that stupid scheme, I don't know.
I do note, though, that SND_PCM_FORMAT_S24_LE format is defined as "Signed 24 bit samples for each channel (Little Endian byte order in **_4_** bytes)", which suggests all that byte mangling is not needed at all (we are not using PCM_FORMAT_S24_3LE which really is packed 3-byte values). This means that the buffer is 32bits and 24bit can be treated as 32 bits (maybe with 8bit arithmetic right shift).

AntonSchmied:
This is not part of milestone-sp-a3 update anymore. 
And we do not have to add usb-audio support just yet. So I will change this PR to draft.

PR #3055, opened 2022-02-25
Poll hw sources #3051

### LINKED ISSUE #3028: Unused "Parameter" (HW Source) type Messages from Playcontroller-->BBB->PG
Currently, the Playcontroller sends changing HW Source parameters to the Playground (via an SPI "BB_Message" which then is relayed to PG via the BB Bridge), additionally to the TCD-encoded direct stream of events to the audio engine via USB.

It looks like those messages are simply thrown away in https://github.com/nonlinear-labs-dev/C15/blob/unstable/projects/epc/playground/src/proxies/playcontroller/PlaycontrollerProxy.cpp,
`void PlaycontrollerProxy::onMessageReceived(const MessageParser::NLMessage &msg)`, lines 86pp
as there is no action for `PLAYCONTROLLER_BB_MSG_TYPE_PARAMETER`.

Whilt this is useful for debugging on the LPC/BBB side of things, in normal operation it just adds useless additional data transfer as it seems the PG extracts the HW source parameter values _directly_ from the TCD messages as parsed in the AE's event stage.

I would suggest to make that output optional (to allow debugging) by adding a control BB message `PLAYCONTROLLER_SETTING_ID_ENABLE_PARAMETER_MSGS`

### LINKED ISSUE #3051: Poll Hardware Sources
Implement a BB_MSG to allow polling of hardware sources.
Only HW sources whose value can be determined correctly will be send.

/projects/shared/playcontroller/playcontroller-defs.h :
New `PLAYCONTROLLER_REQUEST_ID_POLLHWS = 0x0009`
New `PLAYCONTROLLER_NOTIFICATION_ID_POLLHWS  = 0x0009`

See also: https://github.com/nonlinear-labs-dev/C15/issues/3028

EDIT:
The MIDI Data sent to AE should be framed with a START and STOP marker so that the AE can determine that the polling has finished and all data has been sent.

KlausStrohhaecker:
closes #3051
closes #3028

# 3051:
BB Messaging was added to initiate polling of Hardware Sources (with valid values):
- send request : `PLAYCONTROLLER_REQUEST_ID_POLLHWS`
- returns a notification : `PLAYCONTROLLER_NOTIFICATION_ID_POLLHWS`
(/projects/shared/playcontroller/playcontroller-defs.h)
- updated playcontroller tools for test/debug of this feature

Within about 30ms current HW-Source values will be returned:
- nothing for ribbons (those will self-update properly when touched)
- current values for PitchBender and Aftertouch (always available)
- current values for all EHCs (External Hardware Controllers) when there is a valid value at the moment.

A value from an EHC is considered valid when:
- the EHC is initialized
- and the cable is pugged in
- and any auto-ranging has finished
- and internal calculations produced a valid number
-------------
The request was added as there is no other way to obtain "initial" HW-source values from the playcontroller at any time when there is no value change that creates a normal HW-source event.

-------------

# 3028:
BB Messaging was added to enable/disable sending hardware source values to the Playground as BB_MSG (because they get thrown away anyway and just cause traffic over BBB bridge):
- send setting: `PLAYCONTROLLER_SETTING_ID_ENABLE_UI_PARAMETER_MSGS` with a flag (!=0 : enabled)
Now disabled by default.
(/projects/shared/playcontroller/playcontroller-defs.h)
- updated playcontroller tools for test/debug of this feature


KlausStrohhaecker:
Test Status:
- [x] Tested with the (updated) playcontroller tools `lpc` and `lpc-read` (BBB bridge off)
- [ ] Tested with request initiated from Playground

schlenkibus:
I merged this  into my working branch for #3035, the plan is to merge this into mainline as soon as this feature is ready, and milestone a2 is live.

Issue #3054, opened 2022-02-25
PanelUnit: Macro buttons have no effect [MS-SP-A3]


AntonSchmied:
instead journalctl outputs when pushing buttons:
```
playground[450]: 
                                           unhandled exception (type Glib::Error) in signal handler:
                                           domain: g_convert_error
                                           code  : 1
                                           what  : Invalid byte sequence in conversion input

``` 
I have tried merging milestone-a2, since the bug does not occur there. but it didn't help..
Am I missing something? Open for suggestions! 

***EDIT***
This seems to only effect 10i3 and 11i3 NUCs

AntonSchmied:
while #3057 fixes the problem we should investigate the origin of it, to prevent additional/ further harm.

PR #3057, opened 2022-03-02
issue #3054: Add utf-8 capable locale to epc2 os

hhoegelo:
The macro controls names use utf-8 encoded characters for
(A)... (E). The epc2 os so far only supported ascii encoding,
thus any string function processing macro control names
failed.
This patch will not only add proper utf-8 support to the OS, but
also make the playground crash immediately if en_US.UTF-8 support
is not found on the system.


AntonSchmied (commented):
Just tested it on a C15, the Macros work.

However, why not make it part of install.iso? This works on older systems, so we do not need to update, or?
Which brings me to the next question: why has this become an issue, if it hasn't been before?

schlenkibus:
Just a reminder for @hhoegelo and myself to talk about the ordering of our custom symbols, I remember we had to do something with the locale on the EPC 1 to support custom ordering. Not sure if the default UTF8-Locale already includes that.

PR #3064, opened 2022-03-09
Issue 3054 2

### LINKED ISSUE #3054: PanelUnit: Macro buttons have no effect [MS-SP-A3]

hhoegelo:
fixes #3054


AntonSchmied (approved):
I will leave the code review to @schlenkibus.
but i have installed this PR to a C15 running a 11i3 and the macro buttons work just fine.

PR #3066, opened 2022-03-09
issue-3063: rootfs should come with host key

hhoegelo:
If host keys are missing, they are generated on sshd startup.
Our boot process removes all files created during last run,
so rebooting will first remove the keys, then sshd will re-create
new ones. Clients will complain about changed keys - so better
ship keys with the rootfs.


AntonSchmied (approved):
The date+time of the keys corresponds with the boot time of the ePC itself, when checking under /etc/ssh/ 
However, the keys do not seem to change. So I declare the issue as fixed.

PR #3075, opened 2022-03-23
issue-3047: Fix race condition between sshd on bbb and ssh on epc

hhoegelo:
Sometimes, in particular after an update, the sshd on beagle bone takes
some time to start (maybe because of dicing a new host key?).
Any attempts to connect to the server are denied during that phase.
When epc tries to ssh-mount the USB stick connected to BBB, it only
tries once. If this try failes because sshd is not yet ready,
Epc has no access to the USB stick.
This patch adds a dependency to playground service in order to poll
the ssh service at BBB for at most 30s and then brute-force mounts
the usb stick. If mounting fails (because previous attempt already
suceeded), the error is just ignored (note the '-' sign).

### DIFF
--- a/projects/epc/playground/systemd/wait-for-bbb-ssh.service
+++ b/projects/epc/playground/systemd/wait-for-bbb-ssh.service
@@ -1,0 +1,11 @@
+[Unit]
+Description=Wait up to 30s for BeagleBone sshd
+After=network.target systemd-modules-load.service
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c "for i in {1..30}; do ssh -o StrictHostKeyChecking=no -o ConnectTimeout=1 root@192.168.10.11 exit && exit || sleep 1; done"
+ExecStart=-mount /mnt/usb-stick
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


projects/epc/playground/systemd/wait-for-bbb-ssh.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Wait up to 30s for BeagleBone sshd
+After=network.target systemd-modules-load.service
+
+[Service]
+Type=oneshot
+ExecStart=/bin/sh -c "for i in {1..30}; do ssh -o StrictHostKeyChecking=no -o ConnectTimeout=1 root@192.168.10.11 exit && exit || sleep 1; done"
+ExecStart=-mount /mnt/usb-stick
+
```
AntonSchmied, line 9: Does the second ExecStart wait for the first ExecStart to finish? Or how does this work in comparison to "ExecStartPre"?

hhoegelo: Yes, if type is "oneshot", multiple ExecStart commands can be set, which are executed one after another.

schlenkibus (approved):

PR #3108, opened 2022-04-14
Implement Boot-Polling

### LINKED ISSUE #3090: implement HW Poll Start/Stop
When polled, the PlayController frames sent HW positions with Start/Stop markers. The Stop marker can be used to complete the polling process before forming the initial preset after booting, before it is sent to the synth.

Here is a work in progress Chart of the intended HW Polling process (using [mermaid](https://github.com/mermaid-js/mermaid)) :

[](https://mermaid.live/edit#pako:eNqNVMFq4zAQ_ZXBUEggYe8-BDZpactmW0MOvRgW1RrHYu1RVho1G0r_fUeJHdLGcdYnWbw3evP0NO9JYTUmaeLxT0Aq8NaotVNNTlmtdvfOBtIwnc1gtSOuUhh9v4ONs393UFjy7ELB1o3Bo8Ayhx4ZSmcbuNOG56Es0eX0ZBnBmXXFYMuuUGmc546yNVyBrwJruyV4U3VAPwGHSu_OhMTfhSV2tq7RiaJltmgleVaOv3kOzWuNY8gEAQ8vsLLBFehzurn5TN7Xm8uXHrCryI-wuHc87HD2KURcWPwYR2Bva2TbviZxuUJmQ2svGmWP-NDQFxEt81TFfmtAxFdfT1GvuDa0x8rJ8FNuOKfa2g1kz8vl49N9TiDfOf-Tr9sKaW-fK8B4UG_K1Coae2APmPnw8it7Xo2MnsDG-iO-19Ve7GWHeuH9VvVCL3vWuUWWpg45OIp_GWpVe5hOZ5JxfiRtHBZ8rZblSmS3uevIty1Vnsr_RNFuriVR7rMN4tVExWKDgYqAy_3sX_dJmmCkIquvv_FQnWOGYq7kUgwbmSKgHMJvkqc_xA10gV0oksBHBcAWDMm2qocqtZCTGdXVkAakMdQSgm13R-czsHvRl2bk6TAcza3l6BRpcSaZJA26RhktE_c9pihPJCoN5kkqS42lCjXnSU4fAg0brRijSuuStJQc4iRRga2cUySpDF_sQO3UblEf_wB0BwJ2)

- Polling should be fail-safe: if the PlayController doesn't respond (f.e. during IDE tests), the Shutdown State will be transmitted to the AudioEngine anyway
- PlayGround during Polling: (after PollStart from PlayController)
  - the actual Boot Sound (with known HW ADC Values) will be created granularly (with/for each responding HW Source)
  - non-returning Pedals should be set indirectly (not modulating any Macro)
- PlayGround PollStop: the actual Boot Sound can be transmitted to the AudioEngine
- expectations regarding Changed-Indicators:
  - if the Shutdown State already was changed, the Boot State should be, too
  - if a non-returning Pedal differs from its Shutdown State, the Boot State shouldn't be affected
  - if any returning Source differs from its Return Position, the Boot State should indicate Changed until the Source returns

schlenkibus:
This is an Hotfix for issue: #3090
- [x] prevented localOnOff rules to run while we initialize the settings
- [x] implement transmission of polled values to PG via one message that is distinct from HWChangeMessage
- [x] polled values of non returning pedals will no longer change MC position 

closes #3090

### DIFF
--- a/projects/epc/audio-engine/src/synth/C15Synth.cpp
+++ b/projects/epc/audio-engine/src/synth/C15Synth.cpp
@@ -182,6 +182,16 @@ void C15Synth::doChannelModeMessageFunctions()
       case AllNotesOff:
         m_dsp->fadeOutResetVoiceAllocAndEnvelopes();  // NOTE: currently resets all (internal AND external) notes
         break;
+      case PollStart:
+      {
+        nltools::msg::HardwareSourcePollState msg{true};
+        nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
+      }
+      case PollEnd:
+      {
+        nltools::msg::HardwareSourcePollState msg{false};
+        nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
+      }
       default:
       case NOOP:
         break;
--- a/projects/epc/audio-engine/src/synth/C15Synth.cpp
+++ b/projects/epc/audio-engine/src/synth/C15Synth.cpp
@@ -182,17 +182,14 @@ void C15Synth::doChannelModeMessageFunctions()
       case AllNotesOff:
         m_dsp->fadeOutResetVoiceAllocAndEnvelopes();  // NOTE: currently resets all (internal AND external) notes
         break;
-      case PollStart:
-      {
-        nltools::msg::HardwareSourcePollState msg{true};
-        nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
-      }
       case PollEnd:
       {
-        nltools::msg::HardwareSourcePollState msg{false};
+        nltools::msg::HardwareSourcePollEnd msg;
+        msg.m_data = m_inputEventStage.getPolledHWSourcePositions();
         nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
       }
       default:
+      case PollStart:
       case NOOP:
         break;
     }
--- a/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
+++ b/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
@@ -93,8 +93,6 @@ void InputEventStage::onMIDIMessage(const MidiEvent &midiEvent)
   }
 }
 
-#pragma clang diagnostic push
-#pragma ide diagnostic ignored "bugprone-branch-clone"
 void InputEventStage::onTCDEvent()
 {
   auto decoder = &m_tcdDecoder;
@@ -191,13 +189,11 @@ void InputEventStage::onTCDEvent()
     }
     case DecoderEventType::PollStart:
     {
-      nltools::Log::error("got PollStart!");
       m_isPolling = true;
       break;
     }
     case DecoderEventType::PollStop:
     {
-      nltools::Log::error("got PollStop!");
       m_isPolling = false;
       m_channelModeMessageCB(MidiChannelModeMessages::PollEnd);
       break;
--- a/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
+++ b/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
@@ -198,6 +198,7 @@ void InputEventStage::onTCDEvent()
     case DecoderEventType::PollStop:
     {
       nltools::Log::error("got PollStop!");
+      m_isPolling = false;
       m_channelModeMessageCB(MidiChannelModeMessages::PollEnd);
       break;
     }
--- a/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
+++ b/projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
@@ -18,6 +18,9 @@ InputEventStage::InputEventStage(DSPInterface *dspHost, MidiRuntimeOptions *opti
 
   for(auto &hw : m_localDisabledPositions)
     std::get<0>(hw) = std::numeric_limits<float>::max();
+
+  for(auto& p: m_polledHWPositions)
+    p = std::numeric_limits<float>::max();
 }
 
 template <>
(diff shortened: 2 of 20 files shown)


MatthiasSeeber (approved):
- [x] all PlayController Poll Events (Start, HW, Stop) now implemented and verified (closes #3051)
- [x] changed Shutdown State remains changed Boot State
- [x] non-returning Pedals do not affect Boot state
- [x] returning Sources modulate Boot state (HWUI changed indication until drop to return pos)
- [x] non-returning Ribbons associated to Macros reflect Macros
- [x] HW Polling already suppresses Midi Out (#3068) (although before HW Polling, there seems to be no guarantee for that)
- [x] code review revealed no questions for me

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -182,7 +182,14 @@ void C15Synth::doChannelModeMessageFunctions()
       case AllNotesOff:
         m_dsp->fadeOutResetVoiceAllocAndEnvelopes();  // NOTE: currently resets all (internal AND external) notes
         break;
+      case PollEnd:
+      {
+        nltools::msg::HardwareSourcePollEnd msg;
+        msg.m_data = m_inputEventStage.getPolledHWSourcePositions();
+        nltools::msg::send(nltools::msg::EndPoint::Playground, msg);
+      }
```
hhoegelo, line 190: pls add break (in case some one adds another case later)

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -182,7 +182,14 @@ void C15Synth::doChannelModeMessageFunctions()
...
       default:
+      case PollStart:
```
hhoegelo, line 192: PollStart doesn't seem to be used at all, maybe remove?

projects/epc/playground/src/parameters/HardwareSourceSendParameter.cpp
```diff
@@ -178,7 +178,12 @@ size_t HardwareSourceSendParameter::getHash() const
 
 bool HardwareSourceSendParameter::isLocalEnabled() const
 {
-  return m_isEnabled;
+  if(!m_isEnabled.has_value())
```
hhoegelo, line 181: Please simplify by replacing the three optional<bool>'s by simple bools and init them in ctor

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -43,6 +45,37 @@ AudioEngineProxy::AudioEngineProxy(PresetManager &pm, Settings &settings, Playco
...
+      int index = 0;
+      for(auto value: msg.m_data)
+      {
+        auto param = m_playcontrollerProxy.findPhysicalControlParameterFromPlaycontrollerHWSourceID(index);
+        index++;
+        if(auto p = dynamic_cast<PhysicalControlParameter*>(param))
+        {
+          PhysicalControlParameterUseCases useCases(p);
+
+          if(value != std::numeric_limits<float>::max())
```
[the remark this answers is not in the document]
hhoegelo: nice to have: send eb only if (at least) one use case actually changed something

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
```diff
@@ -173,20 +176,26 @@ void InputEventStage::onTCDEvent()
...
-      // todo: disable midi send?
-      // todo: notify playground (hw poll ack/start)?
+      m_isPolling = true;
       break;
     }
     case DecoderEventType::PollStop:
     {
-      // todo: re-enable midi send?
-      // todo: notify playground (all hw sources were polled, initial sound can be loaded)
+      m_isPolling = false;
```
hhoegelo, line 197: can this introduce a dead lock?
- AE requests polling
- LPC sends "PollStart"
- LPC crashes/reboots
-> m_isPolling will never be reset

(...just a thought)
-

MatthiasSeeber: As I understand it, the Polling will not only happen during actual booting, but rather every time in which either the Synth, the PlayGround or the PlayCobtroller(?) restart and reconnect.

projects/epc/playground/src/use-cases/PhysicalControlParameterUseCases.cpp
```diff
@@ -20,3 +21,29 @@ void PhysicalControlParameterUseCases::changeFromAudioEngine(double value, HWCha
...
+{
+  auto scope = UNDO::Scope::startTrashTransaction();
+  m_physicalParam->setIndirect(scope->getTransaction(), position);
+}
+
+bool PhysicalControlParameterUseCases::applyPolledHWPosition(float value)
+{
+    if(value != std::numeric_limits<float>::max())
+    {
+      auto oldPos = m_physicalParam->getValue().getRawValue();
```
hhoegelo, line 35: Consider using 
if(m_physicalParam->getValue().differs(dValue)) {... return true; }

so floating point comparison problems are handled and setIndirect/changeFromAudioEngine is not called unnecessarily. But maybe I am wrong and we indeed want to call the methods in question even if there is no change? But then - why setting values w/o sending the edit buffer?

schlenkibus: done!

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
schlenkibus: What @MatthiasSeeber said is correct. The playcontrollerProxy will send the request every time the LPC HB stumbled or Proxy Ctor is called. As long as the LPC responds, no matter when, we will be fine IMO.

hhoegelo (approved):

PR #3113, opened 2022-04-19
#3111

### LINKED ISSUE #3111: Split-Sounds: Im Feedback Mixer für Effects "From I/II" anbieten
Da der FB-Effects-Zweig ein monophones Signal führt, kann es auch sinnvoll in den anderen Part eines Split-Sounds geroutet werden.

schlenkibus:
enable 'From I/II' for split sounds, cleanup PanelUnitParameterMode to use C15::PIDs instead of hardcoded numbers
closes #3111


schlenkibus:
- [ ] Parameter Selection Split -> Single keeps From II selected (should go to Effects)
- [ ] Convert Layer -> Split Value is lost and mod-aspects aswell

MatthiasSeeber (approved):
One tiny thing: When converting Split --> Layer, the From I/II Parameters are correctly transferred (meaning untouched). However, Effects still is selected, which is not necessary (probably left-over from former hiding of parameter).

projects/epc/playground/src/testing/unit-tests/dual-mode-acceptance/UnselectDiabledParameterOnSoundTypeChange.cpp
```diff
@@ -144,13 +144,13 @@ TEST_CASE("FROM Layer")
     THEN("Split Preset Loaded")
     {
       ebUseCases.load(presets.getSplitPreset());
-      CHECK(eb->getSelected(VoiceGroup::II)->getID() == ParameterId { 160, VoiceGroup::II });
+      CHECK(eb->getSelected(VoiceGroup::II)->getID() == ParameterId { 354, VoiceGroup::II });
```
MatthiasSeeber, line 147: Just a suggestion: Maybe it would be benefitial/more readable to also use explicit Parameter IDs instead of hard coded values in Test Cases.

MatthiasSeeber (changes requested):
I'm sorry. We again seem to have a very incremental approach to the implementation.
- [x] Convert Split --> Layer: From I/II stays selected, but...
- [x] ... (similar to Layer --> Split), the Parameter's position (and Mod aspects) are lost/defaulted during conversion (which I didn't explicitly mention before)

MatthiasSeeber (approved):
:ok:
- deselection non-single to single
- no deselection or reset of pos/mod aspects on split <--> layer
- sound check

:question:
```
acceptance-tests/src/MidiSafeModeTests.cpp:97: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal
```

PR #3121, opened 2022-04-21
WebUI: Issue 3120: upload-firmware cancel-behaviour 

### LINKED ISSUE #3120: WebUI : cancelled "choose file" dialog for FW-upload is not reflected in calling window
When the "choose file" dialog is cancelled rather than an actual file selected, the dialog closes but the panel then wrongly shows "uploading" and you cannot enter the dialog again -- enabled, only after closing and re-invoking the "setup" panel.

schlenkibus:
moved setting label text to uploading.. only if we really are uploading, and remove last uploaded file after completion, so pressing cancel after succesfull upload will not pretend it has an file selected
closes #3120

### DIFF
--- a/projects/web/static/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
+++ b/projects/web/static/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/html/setup/Setup.java
@@ -1,10 +1,8 @@
 package com.nonlinearlabs.client.world.overlay.html.setup;
 
 import java.util.Map;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


KlausStrohhaecker (dismissed):
Does not work here.
Cancelling still gives the "uploading..." message

KlausStrohhaecker:
When I don't merge `main` into this branch (as we should do every morning) it suddenly works.
What's going wrong here?

hhoegelo (approved):

schlenkibus:
I see no reason to break with our convention: not to merge to `main` directly. 
A merge to `milestone-synthengine-update` would be fine with me. 
Given that the feature is stable enough to be merged. I cant really tell what @KlausStrohhaecker test verdict was.

schlenkibus:
@schlenkibus to answer my own question: seems to work fine still (tested from ide)

PR #3127, opened 2022-04-25
Supervisor dual chip support

### LINKED ISSUE #3132: Supervisor dual chip support
Latest batch -- 2022-04 -- of Uniboards (uniboard_2b / V7.2a) have ATmega164A µC chip, compared to previously used ATmega16A. ATmega16A is phased out and obsolete.

KlausStrohhaecker:
closes #3132

Latest batch -- 2022-04 -- of Uniboards (uniboard_2b / V7.2a) have ATmega16**4**A µC chip, compared to previously used ATmega16A. ATmega16A is phased out and obsolete.

This PR extends the code and the AtmelStudio 7 (Windows) project files to allow for both chip versions, see readme.md

NOTE: Build process is still manual, therefore in the current automated build process there are NO changes at all.


projects/supervisor/src/main.cpp
```diff
@@ -309,8 +333,21 @@ void SM_HotReboot(void)
 }
 
 //----------------------------------------------------------------------
+
+volatile char dummy;
```
hhoegelo, line 337: Did you try using "__attribute__((used))" instead?

for example:
$ cat a.c
static const char srcvers[] __attribute__((used)) = "VERSION/foo.c/1.01/09.04.15";
$ gcc -O3 -c a.c
$ strings a.o
VERSION/foo.c/1.01/09.04.15

projects/supervisor/src/comm.cpp
```diff
@@ -22,18 +23,29 @@ void COMM_Init(void)
...
+  UCSR0C = (0b00 << UMSEL00) | (0b00 << UPM00) | (1 << USBS0) | (0b11 << UCSZ00);  // async uart, no parity, 1 stop bits, 8 data bits
+#endif
   sei();
   enable = 1;
 }
 
 void COMM_DeInit(void)
 {
   enable = 0;
-  UCSRB  = 0;
```
hhoegelo, line 36: As I cannot compile and test supervisor code, I don't know if this proposal would actually work, but please consider reducing the #if/#else/#endif usages by adding a single abstraction for the 'things with same meaning'. For example in the beginning of the file there could be one block:
```
#if IS_TYPE_XX4_CHIP == 0
#define NL_UBRRH UBRRH
#define NL_UDR UDR
...
#else
#define NL_UBRRH UBRR0H
#define NL_UDR UDR0
...
#endif 
```

projects/supervisor/src/main.cpp
KlausStrohhaecker: Ah thanks, never heard of __attribute__((used)) before and did not find it when I searched for a solution to the problem.
Will use in future code/projects.

projects/supervisor/src/comm.cpp
KlausStrohhaecker: It may well be that #if clauses appears to clutter the code for many readers, but for me it is the opposite in this case. I prefer to make such dangerous low-level hardware dependencies as visible as possible, just because of their dangers.
In this case, it would not work anyway as not only the register locations have changed but also some bitfields have been redistributed among different registers now so access is different, too.
A true HAL would have been the best choice but probably not worth the effort here.

PR #3147, opened 2022-05-06
Update run.sh

KlausStrohhaecker:


schlenkibus:
same story: no merge to main without build + on-device tests. Lets move this PR to milestone-synthengine, or make this PR a draft.

schlenkibus:
@KlausStrohhaecker I would merge this PR into 'milestone-synthengine-update' later today, if there is no objections, or explanation why we need to merge this into main.

KlausStrohhaecker:
I'm fine with whereever it may go ;-)
Being a very trivial change I would think merging to main were safe (and of course it was tested).

As for the general discussion (@hhoegelo) where a PR should go... it always depends.

For example, some time ago I almost lost a day in searching for bug in https://github.com/nonlinear-labs-dev/C15/pull/3141 which wasn't one in the end. Rather, https://github.com/nonlinear-labs-dev/C15/pull/3136 did not make into main and the previous erroneous code clashed with my work. Then I had to manually extract the relevant changes and apply them manually in my branch because merging https://github.com/nonlinear-labs-dev/C15/pull/3136 into my branch would have created even more havoc (at the time).
For my work, I often base on previous fixes and features which often makes things complicated in this regard...

PR #3151, opened 2022-05-16
Aftertouch and Pitchbender updated processing

### LINKED ISSUE #2876: Pitch Bender: Values between -5 ... 0 ... 5 may cause offset
when holding the pitch bender steadily within the range of -5 ... 0 ... 5, the value is reset to 0.
This causes an offset (positive or negative), which does not reset itself to zero automatically, when the bender is left untouched.

KlausStrohhaecker:
### Aftertouch improvements
- calibration (also retained in EEPROM) now possible, to make response more even across keys
- (semi-)intelligent key-tracking to use the mono-aftertouch almost in a poly-aftertouch-like manner
- arbitrary keys can be masked from key-tracking
- legacy mode is still available (off by default but forced to on when no valid calibration is found)
- diagnostic/status messaging available
---
### Pitchbender improvements
In general, the pitchbender algorithms are now more robust.
- no auto-zero bug (fixes #2876)
- auto zero position does not affect end-of-range values (fixed bug)
- maximized available hardware range (fixed bug)
- allows for non-saturated end stop adc values (maximizes mechanical range)
- output is silent until a valid auto-zeroing has taken place (prevents wrong persistent output values other than zero)
- bender can be reset at runtime (to restart end-stop ranging, or reset bender offset in legacy mode)
- ~~two available auto-zero settling characterstics: insensitive(default), sensitive. A trade-off between _sending spurious values close to zero even without touching the bender at all_ and the risk of _entering a false (unintentionally) nulling by holding the bender in an exact position_.~~ Dismissed.
  - insensitive setting (default) : when the bender is in a settled position, it takes _more_ of a initial position change to start sending values. Similarily, the threshold of when a movement is considered to be over ("settled") is _larger_ and the time constant for this is _shorter_. This means, as a trade-off, it is more likely to initiate a false "nulling" by holding the bender exactly in the same position for some time.
  - ~~sensitive setting : from a settled position, it takes _very little_ initial movement to start producing values, so even just a little bit of vibration might start sending values. Similarily, the threshold of detecting a settled position is _lower_ and it must valid for a _longer_ time. By this, it is way less likely that a user might ever have a "false nulling" event by holding the bender in a steady position.~~ Dismissed.
- ~~legacy mode is still available (off by default)~~ Dismissed.
- diagnostic messaging available (eg to detect massively off-centered zero position --> bender needs servicing)
---
### Miscellaneous
- (bugfix) no hardware-source values are generated for bender, ribbon 1 and ribbon 2 when top cover is not attached
---
ToDo's
- [x] (@schlenkibus?) Implement "Aftertouch: legacy mode" on|off in HWUI and WebUI (but display and allow this option only when "aftertouch status" reports "calibrated" because non-legacy mode is only useable/available with a calibrated keybed anyway). EDIT: Maybe it's convenient enough to have this persistent option in the "Developer Options" of WebUI. Just make sure the LPC receives it some time after booting just like any other persistent setting  (and maybe check if LPC reports the choosen mode in AT-status).

closes #2876


KlausStrohhaecker:
@schlenkibus
There was a misunderstanding in the messaging details, Aftertouch Status is sent as the value of the Notification message.
You had used an extra message type "PLAYCONTROLLER_BB_MSG_TYPE_AT_CAL" which is something different -- for the calibration process itself (and a much longer message, with the memcpy() causing a stack overwrite crash in case this message were ever receieved ;-)
I've fixed this, now working as intended, by adding a new parser section in "PlaycontrollerProxy::onNotificationMessageReceived()"

---

> This should not be a problem as the calibration status shouldn't be changing during the runtime and by the time the user can press the Setup Button the correct state would already be installed.

True, but only if the C15 has *not* been shut off with the Settings page already selected (which is bug, in the end)

--- 
sorry for the additional (but required) reformatting of various unrelated stuff, do you use auto-format in your IDE?

KlausStrohhaecker:
I have tested the new feature mechanisms thoroughly and everthings seems to work as planned. The PR is ready to merge, from my side.

Except for the bug mentioned above, that is. Don't know if there is a priority to fix it, most likely it is not.
Da Boss will have to decide...

schlenkibus:
The new Options are displayed differently than other on/off options we already had. Others show the possible value above/below the current selected one and don't allow wrapping on inc/dec

hhoegelo (approved):

PR #3154, opened 2022-05-18
Issue 3152

### LINKED ISSUE #3152: introduce extended placeholders for parameter labels and info texts
Several parameters possess special labels/infos that include Unicode (like Macros) or change with the VoiceGroup (like FB Mix From I/II or Out Mix To FX I/II). Therefore, a placeholder mechanism was implemented, capable of string replacements.

For some new parameters (FX/Part Pan, FX I to II/II to I, ...), the requirements for string replacements now also include the current Sound Type. So, the placeholder mechanism should be extended and the Parameter-DB should use these placeholders (instead of directly providing unicode).

With this extension, we should be able to provide new parameters with context-aware labels and infos without having to define several variations of one parameter (with different ids).

schlenkibus:
closes #3152


MatthiasSeeber (approved):
Labels in Boled and WebUI seem correct now.

AB I-II

PR #3162, opened 2022-05-19
Fix tests leaks

hhoegelo:
* fix memory leaks 
* fix file handle leaks
* use applications main context instead of static main context wherever appropriate
* fix race condition in WebSocket channels


schlenkibus:
I don't like the history at all, what is going on with all those old commits?

PR #3173, opened 2022-05-25
changed attack velocities to be bipolar

schlenkibus:


MatthiasSeeber (approved):
Code seems legit. Synth Test ok. HWUI Test (BBBB) ok. WebUi couldn't be tested, because WebUi requires new software version.

But: we may have introduced a flaw:
- Attack Velocity: reduction (positive values), gain (negative values)
- Decay 1/2 Velocity: reduction (negative values), gain (positive value)
- (Release Velocity: reduction for positive values)

So, the Decay Velocities behave inversely proportional to Attack/Release Velocities. (maybe worth a ticket?)

PR #3178, opened 2022-05-25
fixup wrong scalings for parameters

schlenkibus:
changed order of parameter declarations to match order of parameter-list


MatthiasSeeber (approved):
tested via HWUI, webui requires new software version
- [x] Velocities
- [x] Pitch and Time and Res Key Trackings and Key Pannings
- [x] Scale Offsets
- [x] Shaper Mix - FB Env C
- strange: Unison Detune defaults to slightly greater than but not zero
- also enigmatic: ModulateableParameterWithUnusualAmount only for Release Times, but not any other Env Time (without any noticeable distinction in HWUI: mod amount in percent, and mod limits in milliseconds with all Env Times)

PR #3195, opened 2022-06-02
fix crash

### LINKED ISSUE #3194: HWUI: Unsion page carousel : Playground crash "after" PAN
At Unision page hitting soft-button-4  four times  leads to playground crash. 
Any entries missing ?

image

image

schlenkibus:
closes #3194


MatthiasSeeber (changes requested):
The main issue has been succesfully fixed: Unison Pan selection via SB4 doesn't crash anymore.
Unintended by-product and additional fix: same issue was with Mono Legato selection via SB4 ( 
:+1:  subsequent crash also solved)

However:
- Split Sounds: Part / Split Point selection via SB4 leads to crash as well
- Layer Sounds Part / Frade Range selection via SB4 leads to crash as well

MatthiasSeeber (approved):

PR #3196, opened 2022-06-02
updated scaling-mod-amounts

### LINKED ISSUE #3197: Scale Offsets - inconsistencies
observed on latest Preview: Scale / Offset - Mod Amounts:
- unit is wrong: st instead of ct (which unnecessarily requires more digits) (upper/lower limits are ok though)
- clipping is wrong: negative amounts indicate zereo (upper/lower limits are ok though)

schlenkibus:
closes #3197


MatthiasSeeber (approved):
Suggestion: usually, bipolar DisplayScaleTypes indicate their absolute maximum instead of the range. Examples: Bipolar100Percent --> [-100% ... 100%], Bipolar200Percent --> [-200% ... 200%]. According to this scheme, this ScaleConverter should be named Bipolar1200Ct.

PR #3202, opened 2022-06-07
closes #2891

### LINKED ISSUE #2891: BOLED: Für geschaltete oder gestufte Parameter sollten kein "Fine" ("F") aktivierbar sein
betroffen sind:
Oscillator A/B: Reset
Mono: Enable, Priority, Legato
Unison: Voices

schlenkibus:
added suppression of fine state for parameters where fine and coarse denominators are the same

closes #2891

### DIFF
--- a/projects/epc/playground/src/parameters/Parameter.cpp
+++ b/projects/epc/playground/src/parameters/Parameter.cpp
@@ -327,6 +327,13 @@ bool Parameter::isBiPolar() const
   return m_value.isBiPolar();
 }
 
+bool Parameter::isFineAllowed() const
+{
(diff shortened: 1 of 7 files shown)


MatthiasSeeber (approved):
**Note** This works and disables the Fine Mode alltogether when selecting a Parameter which has no distinct Fine resolution. However, there also is a permanent Fine Mode: Shift + Fine. The current implementation interferes with this, so that permanence can become broken upon selection.

It would be nice if the Parameter selection just would mask the display of `F`, instead of disabling the mode, so that permanence could be maintained.

PR #3203, opened 2022-06-07
closes #2886

### LINKED ISSUE #2886: HWUI: Preset selection screen: HOME and END function by holding Shift while inc/dec [ENHANCEMENT]
could be handy in preset menu : 
jump to **first** preset by holding  [ **'SHIFT' and 'Dec(-)'** ]  or [ **'SHIFT' and Encoder-CCW** ] (HOME or POS1 function)
jump to **last** preset by holding  [ **'SHIFT'  and 'Inc(+)'** ] or [**'SHIFT'  and Encoder-CW** ] (END function)

schlenkibus:
added shift modifier for Preset Navigation in the HWUI PresetList. Shift + Inc Dec will select the first/last preset in the current Bank

closes #2886

### DIFF
--- a/projects/epc/playground/src/use-cases/BankUseCases.cpp
+++ b/projects/epc/playground/src/use-cases/BankUseCases.cpp
@@ -342,14 +342,11 @@ PresetManager* BankUseCases::getPresetManager() const
 void BankUseCases::selectFirstOrLastPreset(int inc)
 {
   Preset* newPreset = nullptr;
-  if(inc > 0)
-  {
+
+  if(inc < 0)
     newPreset = m_bank->getPresetAt(0);
-  }
-  else if(inc < 0)
-  {
+  else if(inc > 0)
     newPreset = m_bank->getPresetAt(m_bank->getNumPresets() - 1);
-  }
 
   if(newPreset)
   {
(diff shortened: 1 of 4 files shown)


MatthiasSeeber (changes requested):
- [x] Preset Screen: Shift + Inc/Dec works but has opposite direction as compared to normal Inc/Dec
- [x] :+1: Bank Screen: Shift + Inc/Dec works with correct direction

projects/epc/playground/src/use-cases/BankUseCases.cpp
```diff
@@ -255,9 +257,9 @@ void BankUseCases::insertBank(Bank* toInsert, size_t insertPosition)
                                                    m_bank->getName(true));
   auto transaction = scope->getTransaction();
 
-  toInsert->forEachPreset([&, i = 0](auto p) mutable {
-    m_bank->insertPreset(transaction, insertPosition + (i++), std::make_unique<Preset>(m_bank, *p));
-  });
+  toInsert->forEachPreset(
+      [&, i = 0](auto p) mutable
+      { m_bank->insertPreset(transaction, insertPosition + (i++), std::make_unique<Preset>(m_bank, *p)); });
```
MatthiasSeeber, line 262: Was there a reason to put `i` into the lambda capture? F.e. see line 114, where `i` is explicitly typed outside of the capture (which doesn't need the `mutable` keyword) - which I find to be a bit more readable.

MatthiasSeeber (approved):
Tested Preset and Bank Screens with and without Direct Load - everything seems to be working.

_(Maybe, the Setup Screen would also benefit from providing this navigational feature?)_

PR #3208, opened 2022-06-08
issue 3183

MatthiasSeeber:
Shaper A/B FB Mix -  Gate-Env C:
- [x] make short label even shorter (no more text clipping): proposed label: `G - Env C`
Boled-Wed-Jun--8-14:12:13-2022
- [ ] make mod amount display span 200% instead of 100% (because parameter is bipolar - see for example Output Mix - A Level)


schlenkibus:
The shorter label seems to fit into the carousel now, even with the changed indication.

schlenkibus (approved):

schlenkibus:
@MatthiasSeeber I am requesting an review from you this way, as github does not let me to.

PR #3209, opened 2022-06-08
issue 2836

### LINKED ISSUE #2836: MIDI: Performance-Einbruch
Der C15 kann überlastet werden, indem man CCs mit unterschiedlichen Werten an gekoppelte HW-Sources (Pedal & Ribbon) schickt. Macht sich v.a. dadurch bemerkbar, dass die Ribbon-LEDS auch beim Stoppen des MIDI-Clips noch weiter laufen (bildet sich wohl ein Stau irgendwo). 

Lässt sich auch erreichen, wenn man einen Split-Sound hat und auf beiden Kanälen CCs an die HW Sources schickt - in diesem Fall braucht es keine gekoppelten HW-Sources. 

Je nachdem, wie lange man den Clip abspielt, bleibt auch die UI stehen und es wird nicht mehr auf Tastendrücke reagiert. AE läuft weiter. Im Splitsound-Fall wird der C15 schneller schwach.
[performance_midiclip.mid.zip](https://github.com/nonlinear-labs-dev/C15/files/7260507/performance_midiclip.mid.zip)

schlenkibus:
added sanitize to all set-ops for RoutingSetting, to prevent ReceiveSplit and ReceivePrimary not to be enabled at the same time for all entrys exept notes, those are allowed, closes #2836

### DIFF
--- a/projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
+++ b/projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
@@ -72,13 +72,24 @@ Glib::ustring RoutingSettings::getDisplayString() const
 void RoutingSettings::setState(RoutingSettings::tRoutingIndex hwIdx, RoutingSettings::tAspectIndex settingIdx,
                                bool state)
 {
+  auto updatePair = [](bool& toSet, bool& other, bool value) {
+    other = !value;
+    toSet = value;
+  };
+
   auto idx = static_cast<size_t>(hwIdx);
   auto& val = m_data.at(idx).at(static_cast<size_t>(settingIdx));
   auto changed = false;
 
   if(val != state)
   {
-    val = state;
+    if(hwIdx != tRoutingIndex::Notes && settingIdx == tAspectIndex::RECEIVE_SPLIT)
+      updatePair(val, m_data.at(idx).at(static_cast<size_t>(tAspectIndex::RECEIVE_PRIMARY)), state);
+    else if(hwIdx != tRoutingIndex::Notes && settingIdx == tAspectIndex::RECEIVE_PRIMARY)
+      updatePair(val, m_data.at(idx).at(static_cast<size_t>(tAspectIndex::RECEIVE_SPLIT)), state);
+    else
+      val = state;
+
     changed = true;
   }
 
--- a/projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
+++ b/projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
@@ -43,6 +43,11 @@ void RoutingSettings::load(const Glib::ustring& text, Initiator initiator)
       idx++;
     }
   }
+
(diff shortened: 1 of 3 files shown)


schlenkibus:
- [ ] Test MIDI Clip for Performance Drop
- [ ] MIDI Clip (linked in ticket as: https://github.com/nonlinear-labs-dev/C15/files/7260507/performance_midiclip.mid.zip) is not readable by aplaymidi on linux (performance_midiclip.mid: invalid MIDI data (offset 0x2f))

projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
```diff
@@ -141,6 +158,32 @@ void RoutingSettings::setAllAspectsForIndex(RoutingSettings::tRoutingIndex hwIdx
...
+    auto i = static_cast<size_t>(idx);
+    auto rP = static_cast<size_t>(tAspectIndex::RECEIVE_PRIMARY);
+    auto rS = static_cast<size_t>(tAspectIndex::RECEIVE_SPLIT);
+    auto receivePrim = m_data[i][rP];
+    auto receiveSplit = m_data[i][rS];
+    if(receivePrim && receiveSplit)
+    {
+      m_data[i][rS] = false;
+      changed = true;
+    }
```
MatthiasSeeber, line 185: It seems that the User won't be able to enable Receive Split when Receive Primary is enabled.

MatthiasSeeber (changes requested):
The solution works for

- [x] All Settings On/Off (Primary is chosen)
- [x] enable Primary (whether Split is enabled or not)
- [x] enable Split (when Primary is disabled)

It doesn't work for

- [x] enable Split (when Primary is enabled)

projects/epc/playground/src/device-settings/midi/RoutingSettings.cpp
schlenkibus: Good Catch, I fixed this issue.

MatthiasSeeber (approved):
UI looks good. Now, it depends on a thorough MIDI Test (see above)

davidm2894:
UI apparently does not freeze anymore, so this is a step forward. 

Apart from that, we still have the "Rückstau" considering the incoming MIDI data, which is apparently not processed fast enough by the C15, what you can tell by the LEDs that don't stop for quite a while after stopping the MIDI clip. This happens with Single **and** Split Sounds.

Another new issue is that the BOLED has rendering issues on the screens with the modulated parameters. The param stacks are appearing and disappearing constantly. This happens when the MIDI clip is playing **and** only with Split Sounds. 

https://user-images.githubusercontent.com/84857645/173790311-039cf3c3-55a5-430c-9541-f0d664481003.mp4

I will make a new MIDI file with reaper and hopefully you can use this @schlenkibus.

schlenkibus:
I also tested this midi clip using aplaymidi and my local C15 running (22-16) without any problems.

Maybe we should open a bug report with ableton then?
I will mark this PR as mergeable!

PR #3212, opened 2022-06-09
issue 3198

### LINKED ISSUE #3198: HWUI: Sound (Sub) Screen navigation - proposal
The Sound Screen has several Sub Screens: (Part), Voices, Master (Scale).
Currently, only the Mono Sub Screen has a consistent implementation of back-navigation via SB3 (back...). The Unison Sub Screen only implements SB3 for the first two parameters.
The back-navigation could also be extended to the Part/Master Sub Screens.

schlenkibus:
addded 'back..' button to all applicable parameters in soundscreen submenus

closes #3198

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ModulateableDualVoiceGroupMasterAndSplitPointLayout.cpp
+++ b/projects/epc/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/ModulateableDualVoiceGroupMasterAndSplitPointLayout.cpp
@@ -14,6 +14,8 @@
 #include <sigc++/adaptors/hide.h>
 #include "device-settings/Settings.h"
 #include "device-settings/SplitPointSyncParameters.h"
+#include "use-cases/EditBufferUseCases.h"
+#include "use-cases/SettingsUseCases.h"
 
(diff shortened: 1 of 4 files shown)


MatthiasSeeber (changes requested):
- [x] Voices, Mono and Unison Sub Screens are good

### Scale Sub Screen:

- [ ] SB3 "back ..." leads back to (root) Sound Screen instead of leading one step back to Master Screen
- [ ] SB1 "Master ..." would then be obsolete
- [x] Base Key Parameter lacks SB3 "back ..."

The requested changes seem the most generic and simple/user-friendly to me. However, this will require higher-level approval than just mine.

### Part Screen:
- [x] Part Volume (Split/Layer), Fade From, Fade Range (both Layer) lack SB3 "back ..."

schlenkibus:
Crash Report:
- [ ] init sound
- [ ] hwui: single -> convert to split -> to layer -> convert to Single

schlenkibus:
Open Question is still how we want the Scale Layout to look like. Either allow a back.. button or a Master... one but not both.

MatthiasSeeber (approved):

schlenkibus:
Removed back.. Button from Postion C in Scale Layouts and renamed Master.. Button_A to back..
This changes were made after a discussion with Stephan.

PR #3213, opened 2022-06-09
added CC64-CC69 Continuous Pedal Options

### LINKED ISSUE #2822: MIDI: CC64 - CC 69 auch "Continuous"
Für die CCs 64 bis 69 zusätzlich zu den jeweiligen "switching" Befehlen, auch die Möglichkeit geben, kontinuierliche Befehle zu schicken.

In der Auswahl zusätzlich zu "Switching" dann jeweils noch ein Eintrag mit "Continuous".

schlenkibus:
closes #2822


MatthiasSeeber (changes requested):
I tested MIDI Input via IDE and found that:

- [x] CC64 cont. will be received and control the assigned Pedal (like CC64 sw. already does)
- [ ] however, the switch-like behavior is still present (in UI and Synth)

As suspected in previous discussions, the continuous behavior requires being handled separately and accordingly by the EventStage. So, I'd suggest having another discussion/code review about this.

- [x] untested: does MIDI Out work?

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
```diff
@@ -961,6 +961,13 @@ void InputEventStage::onMIDIHWChanged(MIDIDecoder *decoder)
         const auto msb = hwRes.undecodedValueBytes[0];
         const auto lsb = hwRes.undecodedValueBytes[1];
         float realVal = processMidiForHWSource(m_dspHost, hw, msb, lsb);
+
+        const auto isPedal = hw >= HardwareSource::PEDAL1 && hw <= HardwareSource::PEDAL4;
+        if(isPedal && m_options->isSwitchingCC(hw))
```
MatthiasSeeber, line 966: I am not sure if this will suffice. As far as I can tell, `MidiRuntimeOptions::isSwitchingCC` will only take the CC Number into account, but the distinction between switch/countinous doesn't seem to be made.

schlenkibus: We are comparing the MIDI-Settings for each Hardware source, so we will check for only the CC64-CC69 Switching. As the continuous setting entries are different enum-items

MatthiasSeeber (approved):

PR #3215, opened 2022-06-14
3100

### LINKED ISSUE #3100: Recall, Boled on returning Source: reflect discrepancy to return position
A returning HW Source can be recalled in an arbitrary position - the MCs will reflect that. On the Boled, chaning a returning Source will produce a "!" sign, indicating the discrepancy to the return position.

When a Preset is loaded and a returning Source is visible on the Boled, the "!" sign should be indicated, when the source position differs from the return position.

schlenkibus:
added takeover of m_lastChanged from HUWI on Presetload, If the parameter was changed from hwui before the ! will not be removed on preset load and if the focus stayed the same (preset was loaded via webui or base-unit). closes #3100

### DIFF
--- a/projects/epc/playground/src/parameters/PhysicalControlParameter.cpp
+++ b/projects/epc/playground/src/parameters/PhysicalControlParameter.cpp
@@ -98,6 +98,7 @@ Glib::ustring PhysicalControlParameter::getDisplayString() const
 
 void PhysicalControlParameter::loadFromPreset(UNDO::Transaction *transaction, const tControlPositionValue &value)
 {
+  m_changingFromHWUI = m_lastChangedFromHWUI;
   m_returnModeBeforeLastLoad = getReturnMode();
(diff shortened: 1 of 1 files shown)


MatthiasSeeber (approved):
- when changing a returning source via the HWUI and then loading a new preset:
  - previously, the '`!` disappeared
  - [x] now, the `!` survives
- when the source is changed via WebUI or physical source or MIDI, no `!` will be indicated

The ticket intention was more to indicate that a returning source differs from its return position _(regardless of the event source - which the current implementation of the `!` does not really cover or was intended to do)_. If we still want to indicate this somehow, a better concept is needed.

Edit: I noticed, that often, the `!` is not indicated on the first Coarse Inc/Dec step (which seems to work for the Fine mode). Plus, no `F` is indicated for Fine Mode on HW Sources. (Maybe new Tickets?)

PR #3241, opened 2022-07-05
Use new param db

schlenkibus:
This PR provides a [`YAML`](https://yaml.org/)-based replacement for our Parameter Definitions (`./projects/shared/parameter-db`), _see [`./projects/shared/configuration`](https://github.com/nonlinear-labs-dev/C15/tree/use-new-param-db/projects/shared/configuration) for more details_.

Here is a quick overview:

- new project dependency: [`podman`](https://podman.io/getting-started/installation) _(for testing the configuration only without invoking the build process: [readme](https://github.com/nonlinear-labs-dev/C15/blob/use-new-param-db/projects/shared/configuration/readme.md))_
- [x] all (former spreadsheet) definitions have been replaced by (weakly typed) yaml files
- [x] compile errors have been fixed
- [x] a dockerized typescript-based process parses these files and provides string replacements for several `*,in.*` files _(such as cpp headers etc.)_, generated code artifacts are reachable from within the C15 project
- [x] validation: the process performs lots of "sanity" checks to validate the yaml input, an additional g++ compilation verifies, that provided code compiles
- [x] enhancements: the parameter definitions for cpp now also include Enums and Descriptors for DisplayScalingTypes and ParameterGroups _(more to come in the future)_
- the integration of further enhancements is of course possible and intended, but requires some knowledge about the yaml and typescript structure
- the original spreadsheet is still available, for legacy purposes _(not everything from the sheet has been transferred to yaml, only the essential stuff)_


projects/shared/configuration/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+        DEPENDS copy-results-to-source.sh
+        OUTPUT .build-headers
+        COMMAND rm -rf ${CMAKE_CURRENT_SOURCE_DIR}/generated
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/build-in-docker.sh ${CMAKE_CURRENT_SOURCE_DIR} ${PROJECT_BINARY_DIR}
+        COMMAND touch .build-headers
+)
+
+add_custom_target(
+        configuration
+        ALL
```
hhoegelo, line 30: Do we need 'ALL' here? I think, the dependencies to 'configuration' should rather be explicit.

projects/shared/configuration/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu
```
hhoegelo, line 1: Please pin the version to something that we know is working

projects/shared/configuration/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN apt-get update
+RUN apt-get install -y nodejs npm gcc build-essential
+
+RUN mkdir -p /tmp/package
+COPY package.json /tmp/package/package.json
+RUN npm install -g typescript
+RUN npm install -g /tmp/package
+
+CMD cd /sourceDir && bash process.sh && bash copy-results-to-source.sh /sourceDir /binaryDir
```
hhoegelo, line 14: Using 'sourceDir', 'binaryDir' and the shell scripts in the Dockerfile, even though all those things are undefined at this place and injected only by the 'run'-command seems a bit dirty to me. Consider removing the CMD and instead change the run command accordingly:

podman run -v "$sourceDir":/sourceDir -v "$binaryDir":/binaryDir nl-configuration-build-container "/sourceDir/process.sh && /sourceDir/copy-results-to-source.sh /sourceDir /binaryDir"

hhoegelo (approved):

PR #3248, opened 2022-07-07
Use param db broken

schlenkibus:
initial push to use new parameter db items (Display Scaling Type for Value and Modulation Amount)


projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
```diff
@@ -240,6 +240,10 @@ nltools::msg::SinglePresetMessage AudioEngineProxy::createSingleEditBufferMessag
     {
       //Ignore
     }
+    else if(auto scaleGroup = dynamic_cast<ScaleGroup*>(g))
+    {
+      //Ignore
+    }
```
MatthiasSeeber, line 246: I am not sure here, but is there a reason to ignore the Scale Group in a Single Preset? (Scale is present in any Sonud Type)

MatthiasSeeber (approved):
The code reduction with the new approach seems to be worth it. A superficial first test with running Playground and AudioEngine revealed no issues. We may still want to clarify some things (see comments).

projects/epc/playground/src/proxies/audio-engine/AudioEngineProxy.cpp
schlenkibus: Good Catch, the scale group would never show up, as long as the VoiceGroup will be Global. The parameters are processed by `fillMessageWithGlobalParams`.

projects/epc/playground/src/parameters/scale-converters/LinearBipolar160StScaleConverter.cpp
```diff
@@ -0,0 +1,10 @@
+#include "LinearBipolar160StScaleConverter.h"
+
+LinearBipolar160StScaleConverter::LinearBipolar160StScaleConverter()
+    : LinearScaleConverter(tTcdRange(-7500, 7500), tDisplayRange(-160, 160), PitchDimensionCoarse::get())
```
MatthiasSeeber, line 4: The chosen TcdRange (btw, do we still need that?) is inadequate. It should be an integer multiple of the Fine resolution. So, a TcdRange of +/-8000 (= 5 * 1600) should suffice.

MatthiasSeeber (commented):
I may be wrong here, but I think there are some flawed TcdRanges (see example).

MatthiasSeeber (approved):

PR #3256, opened 2022-08-03
move state from HWUI to own Manager

schlenkibus:
moved HWUI VoiceGroup and HWUI Load to Part data into own class owned by application. HWUI now holds less state that the real playground-lib used to call into, which would be a problem for the C16 without the HWUI


projects/epc/playground/src/proxies/hwui/HWUIActions.cpp
```diff
@@ -3,8 +3,9 @@
 #include "TestLayout.h"
 #include <presets/PresetManager.h>
 #include <libundo/undo/Scope.h>
+#include <Application.h>
 
-HWUIActions::HWUIActions(UpdateDocumentContributor* parent, HWUI& hwui, EditBuffer& eb)
+HWUIActions::HWUIActions(UpdateDocumentContributor* parent, HWUI& hwui, EditBuffer& eb, Application* app)
```
hhoegelo, line 8: If only the VGManager is used here, we could directly pass it instead of introducing a dependency to Application.

projects/epc/playground/src/proxies/hwui/HWUIActions.h
```diff
@@ -5,9 +5,12 @@
 #include <http/ContentSection.h>
 #include <http/SectionAndActionManager.h>
 #include "HWUI.h"
+#include "use-cases/VoiceGroupAndLoadToPartManager.h"
+
+class Application;
```
hhoegelo, line 10: Pls remove forward decl of Application, if not needed anymore

projects/epc/playground/src/proxies/hwui/HWUIActions.h
```diff
@@ -5,9 +5,12 @@
 #include <http/ContentSection.h>
 #include <http/SectionAndActionManager.h>
 #include "HWUI.h"
+#include "use-cases/VoiceGroupAndLoadToPartManager.h"
```
hhoegelo, line 8: pls replace include with forward decl of VoiceGroupAndLoadToPartManager

PR #3263, opened 2022-08-18
3023

### LINKED ISSUE #3023: BOLED + WebUI: An den vier Pedal-Parametern sollte im Zustand "Off" eine kleine Markierung gezeigt werden.
"Off" kann in den Menüs der Pedal-Settings eingestellt werden. Die Signale an den Pedalbuchsen werden dann ignoriert.
Die Bedienung der Parameter über die UI und über MIDI sollte trotzdem möglich bleiben.

Wir müssen uns ein kleines Zeichen (Pedalsymbol (?) mit "X" oder "Off") überlegen, dass neben dem Parameter-Slider eingeblendet wird.

schlenkibus:
closes #3023, for now without extra symbols and with the textual (off) suffix for pedals instead

@nonlinear-labs is that okay with you for now, that we are not using a new symbol, but rather append the (off) suffix to the parameter name?

- [x] yes, keep it like this
- [ ] yes (for now), but we will add a new symbol later
- [ ] no, wait for a new symbol
- [ ] no, here is the new symbol: (enter link here)
- [ ] no, we don't want this feature anymore


projects/epc/playground/src/parameters/PedalParameter.h
```diff
@@ -37,11 +40,13 @@ class PedalParameter : public PhysicalControlParameter
...
  private:
+  void onPedalTypeChanged(const Setting* s);
   const ScaleConverter *createScaleConverter() const;
   void sendModeToPlaycontroller() const;
   void setRoutersModeAccordingToReturnMode();
   bool isLocalEnabled() const override;
 
   PedalModes m_mode = PedalModes::STAY;
   tUpdateID m_updateIdWhenModeChanged = 0;
+  bool m_isHardwareDisabled = false;
```
hhoegelo, line 51: what is that boolean used for?

projects/epc/playground/src/groups/HardwareSourcesGroup.cpp
```diff
@@ -12,69 +12,74 @@ HardwareSourcesGroup::HardwareSourcesGroup(ParameterGroupSet* parent, Settings*
...
+                                                  ScaleConverter::get<Linear100PercentScaleConverter>()));
+}
+
+HardwareSourcesGroup::~HardwareSourcesGroup()
+{
+}
+
+void HardwareSourcesGroup::init()
+{
+  for(auto pedalID : { getPedal1ParameterID(), getPedal2ParameterID(), getPedal3ParameterID(), getPedal4ParameterID() })
```
hhoegelo, line 70: I think this could be written more easy to read and less complicated to executed, e.g.:

```suggestion
findAndCastParameterByID<PedalParameter>(getPedal1ParameterID())->init(settings.getSetting("Pedal1Type"));
findAndCastParameterByID<PedalParameter>(getPedal2ParameterID())->init(settings.getSetting("Pedal2Type"));
findAndCastParameterByID<PedalParameter>(getPedal3ParameterID())->init(settings.getSetting("Pedal3Type"));
findAndCastParameterByID<PedalParameter>(getPedal4ParameterID())->init(settings.getSetting("Pedal4Type"));
```


You would not only save a loops and branches here, but also in PedalParameter.

projects/epc/playground/src/parameters/PedalParameter.h
hhoegelo, line 51: MOT: I am not a big fan of this data duplication - in my opinion we should store a reference to the according setting here. The signal connection could just invoke invalidate without changing internal state (onPedalTypeChanged can be dropped then). The getXYZName methods could ask the setting for the current state.

projects/epc/playground/src/groups/HardwareSourcesGroup.cpp
```diff
@@ -12,69 +12,74 @@ HardwareSourcesGroup::HardwareSourcesGroup(ParameterGroupSet* parent, Settings*
...
+  auto ribbon2 = appendParameter(
+      new RibbonParameter(this, getLowerRibbonParameterID(), ScaleConverter::get<Linear100PercentScaleConverter>()));
 
   appendParameter(new HardwareSourceSendParameter(this, *ribbon2, getRibbon2SendID(),
-                                                  ScaleConverter::get<Linear100PercentScaleConverter>(), 0.5, 100, 1000,
-                                                  m_settings));
+                                                  ScaleConverter::get<Linear100PercentScaleConverter>()));
+}
+
+HardwareSourcesGroup::~HardwareSourcesGroup()
```
hhoegelo, line 64: Do we need a destructor for this class? If so, can we use the default dtor?

projects/epc/playground/src/parameters/PedalParameter.h
```diff
@@ -10,8 +10,9 @@ class PedalParameter : public PhysicalControlParameter
   typedef PhysicalControlParameter super;
 
  public:
-  using super::super;
```
hhoegelo, line 13: The change to an explicit constructor is unnecessary, no?

hhoegelo:
Sorry I always ignored this PR when I saw (1 of 5 tasks done), only now I noticed that this are not really tasks....

hhoegelo (changes requested):
see comments

schlenkibus:
move addParameter from constructor to init()

schlenkibus:
@hhoegelo fixed-up

hhoegelo (approved):

PR #3282, opened 2022-09-02
Pi4

hhoegelo:


os/pi4/build-rootfs/createRootFS.sh
```diff
@@ -0,0 +1,255 @@
+#!/bin/bash
+
+set -e
+
+# wenn CM4, nach /boot/overlays
```
schlenkibus, line 5: either remove these comments or rephrase it into proper documentation which can be understood by other people, if still relevant.

schlenkibus (commented):
some files still have a bit too much text as comments, which I cant understand if these are documentation or artifacts from development

schlenkibus (approved):

PR #3330, opened 2022-10-06
Fix5i3update

hhoegelo:


build-tools/create-c15-update/update_scripts/run.sh
```diff
@@ -170,7 +159,6 @@ epc_update() {
       return 1
     fi
 
-    epc_fix
     return_code=$?
```
AntonSchmied, line 162: unnecessary code below, no?

AntonSchmied (commented):
After executeAsRoot "reboot", do we not want to wait for playground?

AntonSchmied:
I cleaned up a little and pushed as discussed. We can take a look at it next week or trust the process and run checks on milestone-maintenance

PR #3347, opened 2022-10-13
added nltools-2 

### LINKED ISSUE #3346: nltools: dependency on configuration will never cross-compile on BBB
- [x] create a new 'library' nltools2
- [x] move preset-messages to nltools2
- [x] depend on configuration for nltools2
- [x] use nltools2 only on EPC (AE, PG), BBB will never be able to compile the configuration (C++11)

schlenkibus:
which will be used from EPC only (PG and AE), will close #3346


projects/shared/nltools-2/include/PresetMessages.h
```diff
@@ -0,0 +1,273 @@
...
+
+namespace nltools2
+{
+  namespace msg
+  {
+    namespace ParameterGroups
+    {
+      constexpr auto numMacros = 6;
+      constexpr auto numHWSources = 10;
+      constexpr auto numRouters = numMacros * numHWSources;
```
MatthiasSeeber, line 15: Just an observation: I get that copy/paste is the preferred method here in order to get things working as quickly as possible. The usage of enums like `C15::Parameters::Macro_Controls::_LENGTH_` should not be forgotten though.

projects/shared/nltools-2/include/PresetMessages.h
```diff
@@ -0,0 +1,273 @@
...
+
+      struct Parameter
+      {
+        uint16_t id {};
```
MatthiasSeeber, line 19: Just an observation: we might be able to deprecate the id field, since the parameter order is defined by the enums.

MatthiasSeeber (approved):
Seems legit to me.

hhoegelo (commented):
I'd suggest to not introduce new namespace 'nltools2' as, logically, it is still the same, only implemented in another library. This will also produce a more review-friendly patch.

cmake/NL.cmake
```diff
@@ -16,16 +16,43 @@ endfunction()
...
     target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/projects/shared/nltools/include/)
 endfunction()
+
+function(add_nltools_2 TARGET)
+    target_link_libraries(${TARGET} PUBLIC nltools2)
+    target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/projects/shared/nltools-2/include/)
+endfunction()
+
+function(add_nltools_1_and_2 TARGET)
+    target_link_libraries(${TARGET} PUBLIC nltools nltools2)
```
hhoegelo, line 55: this should better call add_nltools and add_nltools_2 instead of copy the code

projects/shared/nltools-2/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+add_library(${PROJECT_NAME} SHARED ${SOURCE_FILES})
+
+add_dependencies(${PROJECT_NAME} configuration)
+target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_BINARY_DIR}/projects/shared/configuration/generated/)
+target_include_directories(${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include/)
+
+prepare_nltools()
+add_nltools(${PROJECT_NAME})
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wdouble-promotion")
```
hhoegelo, line 19: We could even use c++ 17 (or maybe even c++20) here.
Pls consider modern cmake: set_property(TARGET tgt PROPERTY CXX_STANDARD 17)

projects/shared/nltools-2/include/PresetMessages.h
```diff
@@ -0,0 +1,276 @@
...
+
+namespace nltools2
+{
+  namespace msg
+  {
+    namespace ParameterGroups
+    {
+      constexpr auto numMacros = static_cast<size_t>(C15::Parameters::Macro_Controls::_LENGTH_) - 1; //not cool!
+      static_assert(numMacros == 6);
+      constexpr auto numHWSources = static_cast<size_t>(C15::Parameters::Hardware_Sources::_LENGTH_); //cool!
```
hhoegelo, line 15: what exactly is so 'cool' here?

projects/shared/nltools-2/CMakeLists.txt
hhoegelo: (would require cmake 3.8 or 3.12, though)

projects/epc/audio-engine/src/synth/C15Synth.cpp
```diff
@@ -25,9 +25,9 @@ C15Synth::C15Synth(AudioEngineOptions* options)
 
   using namespace nltools::msg;
 
-  receive<SinglePresetMessage>(EndPoint::AudioEngine, sigc::mem_fun(this, &C15Synth::onSinglePresetMessage));
-  receive<LayerPresetMessage>(EndPoint::AudioEngine, sigc::mem_fun(this, &C15Synth::onLayerPresetMessage));
-  receive<SplitPresetMessage>(EndPoint::AudioEngine, sigc::mem_fun(this, &C15Synth::onSplitPresetMessage));
+  receive<nltools::msg::SinglePresetMessage>(EndPoint::AudioEngine, sigc::mem_fun(this, &C15Synth::onSinglePresetMessage));
```
hhoegelo, line 28: Pls either remove explicit names or remove 'using ...' above

cmake/NL.cmake
```diff
@@ -16,16 +16,42 @@ endfunction()
...
     target_link_libraries(${TARGET} PUBLIC nltools)
     target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/projects/shared/nltools/include/)
 endfunction()
+
+function(add_nltools_2 TARGET)
+    target_link_libraries(${TARGET} PUBLIC nltools2)
+    target_include_directories(${TARGET} PUBLIC ${CMAKE_SOURCE_DIR}/projects/shared/nltools-2/include/)
+endfunction()
+
+function(add_nltools_1_and_2 TARGET)
```
hhoegelo, line 54: I would expect nltools2 to depend on nltools, so normally we should be able to form a dependency chain like so:
playground -> nltools2 -> nltools
bbbb -> nltools

This means, prepare_nltools _2 implies preparation of nltools, same for add_nltools_2.

projects/shared/nltools-2/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+cmake_minimum_required(VERSION 3.2)
+
+project(nltools2)
+
+set(BUILD_SHARED_LIBS ON)
```
hhoegelo, line 5: this can be removed (it is written in add_library again)

hhoegelo (approved):
There are some nitpicks I commented, but anyhow this PR is approved.

PR #3356, opened 2022-10-18
Issue 3343

schlenkibus:
- [x] add different messages depending on what went wrong while importing a backup
- [x] notice and change behaviour of FileInStream -> try to add .zip or .gz to filePath even if they already contain one!

 -> this lead to bugs where Opening a FileInStream with a 'file.xml.tar.gz' and the explicit tryZip option would lead to tryZip to fall back to false, as the FileStream can not open 'file.xml.tar.gz.gz' nor 'file.xml.tar.gz.zip' neither.
Now we do not reset the tryZip if we have .zip or .gz at the end of the requested Path and the filename is a correct and existing file


projects/epc/playground/src/xml/FileInStream.cpp
```diff
@@ -136,3 +141,27 @@ bool FileInStream::eof() const
 {
   return m_eof;
 }
+
+bool FileInStream::doesPathEndWithZip(const Glib::ustring &fileName)
```
hhoegelo, line 145: Consider using StringTools::hasEnding

hhoegelo (approved):

PR #3383, opened 2022-11-09
closes #3327

### LINKED ISSUE #3327: WebUI: Preset Load/Reload inconsistencies
In the Preset Tab, the Load/Reload icon on the right next to the shown bank behaves inconsistent, when Direct Load is enabled.

Load a Preset, change something --> the button should become active (which it does).

When the selected Preset is not "initial" and then Init Sound is triggered --> the button should become active (which it does when Direct Load is disabled, but not when it is enabled).

schlenkibus:
added check for Converted and Init uuids to enable load preset button when applicable, closes #3327


projects/web/static/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/presets/LoadPreset.java
```diff
@@ -16,6 +16,11 @@ class LoadPreset extends SVGImage {
 
 	LoadPreset(OverlayLayout parent) {
 		super(parent, "Load_Enabled.svg", "Load_Active.svg", "Load_Disabled.svg");
+
```
hhoegelo, line 19: In case we have a presenter (I am not sure, if) use it instead of the model

hhoegelo (approved):

PR #3400, opened 2022-11-17
Issue 3264

schlenkibus:
found and fixed the culprit: `onExclusiveRoutingLost` set the value of the mod-router, but it never did it inside an transaction (which was loading). calling functions that change state from within `onValueChanged` should raise not one but two red-flags


projects/epc/playground/src/parameters/Parameter.cpp
```diff
@@ -542,19 +542,19 @@ Parameter::VisualizationStyle Parameter::getVisualizationStyle() const
...
-    case 354:
-    case 362:
+    case C15::PID::Comb_Flt_PM_A_B:
+    case C15::PID::SV_Flt_FM_A_B:
+    case C15::PID::Comb_Flt_In_A_B:
+    case C15::PID::SV_Flt_In_A_B:
+    case C15::PID::MC_A:
+    case C15::PID::MC_D:
+    case C15::PID::MC_B:
+    case C15::PID::MC_C:
```
hhoegelo, line 552: what about e and f?

hhoegelo (approved):

PR #3440, opened 2022-12-20
Update c15-unit-check.sh

AntonSchmied:
clean up and since we now have 2 audio devices on the ePC, the check script had to be adjusted


projects/bbb/scripts/c15-unit-check.sh
```diff
@@ -194,27 +201,25 @@ function check_persistent_files() {
     printf "\n" | tee -a $LOG_FILE
 }
 
-function check_alsa_settings() {
-    executeAsRoot \
-        "Alsamixer set to 100%?" \
-        "amixer -D hw:PCH sget 'Master' | grep 100%"
-    printf "\n" | tee -a $LOG_FILE
+check_audio_settings() {
+    printf "Audio device $AUDIODEVICE settings:\n$(getInfoAsRoot "cat /proc/asound/$AUDIODEVICE/pcm0p/sub0/hw_params")\n" | tee -a $LOG_FILE
```
hhoegelo, line 205: I like the tinkering with sourcing the environment file from epc somehow, but on the other hand I am afraid that this could be forgotten/overlooked soon and later we will have problems or at least wonder why our environment (on bbb) is messed up with epc variables? I am unsure.

How about reducing the scope by starting a sub-shell like this:
```
(
  source some_script.sh
  #do something
)
```

projects/bbb/scripts/c15-unit-check.sh
```diff
@@ -194,27 +201,25 @@ function check_persistent_files() {
     printf "\n" | tee -a $LOG_FILE
 }
 
-function check_alsa_settings() {
-    executeAsRoot \
-        "Alsamixer set to 100%?" \
-        "amixer -D hw:PCH sget 'Master' | grep 100%"
-    printf "\n" | tee -a $LOG_FILE
+check_audio_settings() {
```
hhoegelo, line 204: Nitpick:
This function does actually not check anything, does it? 
I would like to separate print/getXYZ from checkXYZ in a way, that check functions may fail if expectations are not met.

hhoegelo (approved):

PR #3458, opened 2022-12-29
Mod rework cleaned up

hhoegelo:
For easier reviews I tried split up the work in several commits with a smaller scope each and a meaningful comment, so I'd recommend reviewing commit by commit.
Due to my limited understanding of the modulation I did not do a lot of manual testing but relied on automatic tests, so, @MatthiasSeeber if you could do some eagle eye testing, this would be much appreciated.


MatthiasSeeber (changes requested):
As usual, I started by running [playground-](https://github.com/nonlinear-labs-dev/C15/files/10331120/playground-test-mod-rework-cleaned-up.txt) and acceptance- tests (see provided log for more details on failing assertions, I suspect approx).

When starting Playground and WebUI after running the tests, there were several strange `Oracle-testbank-gwt-...` appearing?

The Split Point seems to behave as intended in the UI. In the AudioEngine, there still are minor differences, as the mechanism is not yet reworked there.

In the WebUI, setting ModAmounts of bipolar parameters is problematic, because the resolutions are doubled. So, when incrementing an Amount and expecting 0%, 1%, 2% etc. - instead it will go 0%, 0.5%, 1.% etc. This doesn't happen via BOLED, and the YAML definitions are correct.

For minimal ModAmounts (usually 0.1%) there are still inconsistencies i.m.o.:
- the upper/lower limits may not show a range (all values at position)
- when I set up a minimal ModAmount with an MC at 50%, the modulation will only occur when the MC reaches 100% (?)
- when I set up a minimal ModAmount with an MC at 0%, the modulation will occur when the MC reaches 50% (ok)

hhoegelo:
I fixed the tests, added tests for the mod amount webui bug and made them green, played with the WebUI and found no more bugs.
The inconsistencies you found for tiny mod amounts are IMO acceptable.

schlenkibus (approved):

MatthiasSeeber (approved):
- [x] ModAmounts for bipolar Parameters now have correct resolution in WebUI
- Is it intended that playground-test prints lots of messages in the form shown below, although these tests passed all checks?
``` java
javaTestModulation("nothing special",
	new Setup(274,0,1,0,156,0,1),
	new TasksAndExpectations(
        ...
```
- [ ] I played around with the Split Point some more. The Modulation of unlinked Split Points does still misbehave quite a bit (I saw collapsing and even "negative" Overlaps while slowly modulating). I'd suggest to redirect this into a new task for later, in order to finally complete this PR.

PR #3471, opened 2023-01-05
Fix buffer underruns by giving the IRQ handler of the sound card a hi…

hhoegelo:
…gher priority


KlausStrohhaecker:
We might want to check out the following things:
- use more detailed kernel parameters to isolate CPUs/Cores (but never try isolate Core 0), like irqaffinity, nohz, nohz_full, rcu_nocbs, rcu_nocb_poll etc.
- start audio-engine mainy proccess on a dedicated isolated core via taskset
- implement audio-engine cmd-line params to assign (maybe different) dedicated isolated core to specific RT/fast threads (already done in branch pi4-native).

PR #3489, opened 2023-01-10
proposed new set of rules to prevent FX-Disabled display if Fx-FB is used

### LINKED ISSUE #3465: Boled: add indication for 'unused FX' (cross in fx-part of preset-symbol)
FX I unused:
(FX_Mix == 1) && !( ( (Serial_FX > 0) || ( (FB_Mixer_Effects_From_I_II < 1) && (FB_Mixer_Effects != 0) ) ) && (To_FX_I_II < 1) )
   
FX II unused:
(FX_Mix == 0) && !( ( (Serial_FX < 0) || ( (FB_Mixer_Effects_From_I_II > 0) && (FB_Mixer_Effects != 0) ) ) && (To_FX_I_II > 0) )

schlenkibus:
closes #3465

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/VoiceGroupIndicator.cpp
+++ b/projects/epc/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/VoiceGroupIndicator.cpp
@@ -113,9 +113,9 @@ void VoiceGroupIndicator::drawCrossForSingle(FrameBuffer& fb, int centerX, int c
 
 bool VoiceGroupIndicator::drawSingle(FrameBuffer& fb)
 {
-  Rect polyRect = {getPosition().getX(), getPosition().getY() + 3, 6, 6};
-  Rect monoI = {getPosition().getX() + 6, getPosition().getY(), 6, 5};
-  Rect monoII = {getPosition().getX() + 6, getPosition().getY() + 7, 6, 5};
+  Rect polyRect = { getPosition().getX(), getPosition().getY() + 3, 6, 6 };
+  Rect monoI = { getPosition().getX() + 6, getPosition().getY(), 6, 5 };
+  Rect monoII = { getPosition().getX() + 6, getPosition().getY() + 7, 6, 5 };
 
   fb.setColor(FrameBufferColors::C179);
   fb.fillRect(polyRect);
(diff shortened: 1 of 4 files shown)


projects/epc/playground/src/proxies/hwui/panel-unit/boled/parameter-screens/controls/VoiceGroupIndicator.cpp
```diff
@@ -291,5 +290,12 @@ bool VoiceGroupIndicator::isSingleFXNotActive(VoiceGroup vg) const
...
 
-  return fx_mix_cond && !((serial_fx_cond || (from_fx_cond && fb_fx_cond)) && to_fx_cond);
+  auto fb_is_used = eb->findParameterByID({ C15::PID::FB_Mix_FX, VoiceGroup::I })->getControlPositionValue() != 0
+      && eb->findParameterByID({ C15::PID::FB_Mix_Lvl, VoiceGroup::I })->getControlPositionValue() > 0
+      && (eb->findParameterByID({ C15::PID::Osc_A_PM_FB, VoiceGroup::I })->getControlPositionValue() != 0
+          || eb->findParameterByID({ C15::PID::Osc_B_PM_FB, VoiceGroup::I })->getControlPositionValue() != 0
+          || eb->findParameterByID({ C15::PID::Shp_A_FB_Mix, VoiceGroup::I })->getControlPositionValue() > 0
+          || eb->findParameterByID({ C15::PID::Shp_B_FB_Mix, VoiceGroup::I })->getControlPositionValue() > 0);
+
+  return fx_mix_cond && !((serial_fx_cond || (from_fx_cond && fb_fx_cond)) && (to_fx_cond || fb_is_used));
```
MatthiasSeeber, line 300: I think the algorithm is not mature yet.
If for example the `fx_mix_cond` is false, the whole expression results to false, meaning that the corresponding FX are used - which is not the intention I think.

MatthiasSeeber (changes requested):
I'm afraid we first need to formulate the rules adequately before providing an implementation.

https://github.com/nonlinear-labs-dev/C15/issues/3465#issuecomment-1378471552 is also a valid point, indicating that the algorithm is still premature.

MatthiasSeeber (approved):
This looks and feels pretty solid:
- known scenarios that didn't work before now work reliably, I couldn't find any situations in which the indicators were wrong
- in both UIs, the indicators always seem up-to-date (including modulation)
- no unexpected failing tests

projects/web/static/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/Overlay.java
```diff
@@ -64,6 +64,144 @@ public int getSelectedPhase() {
...
+			var fx_input = out_mix || master_serial_fx < 0;
+			var fx_output = master_fx_mix < 1 || master_serial_fx > 0;
+			var fx_feedback = fb_mix_lvl > 0 && fb_mix_fx != 0 && fb_mix_fx_src < 1 && (osc_a_pm_fb != 0 || shp_a_fb_mix != 0 || osc_b_pm_fb != 0 || shp_b_fb_mix != 0);
+			var fx_used = fx_input && (fx_output || fx_feedback);
+			isUnused = !fx_used;
+			setVisible(isUnused && inSingle);
+		}
+	}
+	
+	private class SingleSoundFX_II_Unused extends SingleSoundFX_I_Unused
```
hhoegelo, line 186: I'd prefer the logic to be moved into EditBufferPresenterPorovider and the result into EditBufferPresenter.
The class hierarchy "SingleSoundFX_II_Unused" is a "SingleSoundFX_I_Unused" also looks a bit like the "LargeLabel" is a "SmallLabel" that we are constantly wondering about. I'd prefer to have a base class for both, if this is still needed after using presenters.

PR #3491, opened 2023-01-10
fixup SignalFlowIndication #3490

### LINKED ISSUE #3490: SignalFlowIndication does not update LEDs on modulation

schlenkibus:
connected bruteForceLED update of ParameterEditMode (SignalFlowIndication) to EditBuffer::onChange to update when a parameter is changed via for example: modulation, closes #3490


projects/epc/playground/src/proxies/hwui/panel-unit/PanelUnitParameterEditMode.cpp
```diff
@@ -36,6 +36,9 @@ PanelUnitParameterEditMode::PanelUnitParameterEditMode()
 
   Application::get().getPresetManager()->getEditBuffer()->onPresetLoaded(
       sigc::mem_fun(this, &PanelUnitParameterEditMode::bruteForceUpdateLeds));
+
```
hhoegelo, line 39: Probably the connection to onPresetLoaded can be removed then?

schlenkibus: makes sense!

PR #3492, opened 2023-01-10
Issue 3310, reset HW-Source, which was changed from HWUI, before Storing a Preset

### LINKED ISSUE #3310: Modulation Issues Load Preset With Bender on 1
[bank.zip](https://github.com/nonlinear-labs-dev/C15/files/9608576/bank.zip)
preset xml:
MacroControl C = 0.5
Pitchbender CP = 1.0  
  
after load:
Pitchbender CP = 0.0
MacroControl C CP = 0.0

and changed-indication is showing up for MacroControl

schlenkibus:
The 'reset' of the HW-Source, which was changed via HWUI, happens only when a new Parameter is Selected. So storing a preset without first selecting MC or any other parameter can lead to a bug where the returning-hw-source is not reset before saving.

I introduced an explicit reset, of those from-hwui-changes, before the storing the preset. 

With that change i can no longer reproduce the issue @davidm2894 explained in the ticket

closes #3310

### DIFF
--- a/projects/epc/playground/src/parameters/PhysicalControlParameter.cpp
+++ b/projects/epc/playground/src/parameters/PhysicalControlParameter.cpp
@@ -188,7 +188,11 @@ void PhysicalControlParameter::onSelected()
 void PhysicalControlParameter::onUnselected()
 {
   Parameter::onUnselected();
+  resetChangedFromHWUIPosition();
+}
 
+void PhysicalControlParameter::resetChangedFromHWUIPosition()
+{
   if(m_lastChangedFromHWUI && getReturnMode() != ReturnMode::None)
   {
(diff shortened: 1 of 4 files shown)


MatthiasSeeber (approved):
The returning Source != 0 is in fact reset to zero after pressing `Store` and then `Enter` (to confirm the Store Method and possibly landing in the Rename Screen).

However, I see no need to wait for `Enter`, the Screen changes anyway. Just pressing `Store` could suffice. Pressing `Store` twice (cancelling the process) leads back to the HW Source Screen and the Source is still != 0, although the Screen clearly changed. Maybe we should discuss this and maybe outsource this to another nicetto-have ticket?

PR #3520, opened 2023-01-16
prevent buffer underruns

hhoegelo:
when experimenting with different approaches of
using alsa, we found out that we can use a polling audio loop to drive 8 buffers of 16 frames without ever experiencing underruns on 5i3 and 11i3.

### DIFF
--- a/projects/bbb/scripts/c15-unit-check.sh
+++ b/projects/bbb/scripts/c15-unit-check.sh
@@ -5,7 +5,7 @@ SSH_OPTIONS="-o LogLevel=quiet -o ServerAliveInterval=1 -o ConnectTimeout=1 -o S
 EPC_IP="192.168.10.10"
 LOG_FILE=""
 
-function executeAsRoot() {
+executeAsRoot() {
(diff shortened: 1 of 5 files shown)


AntonSchmied:
While we were finding a way to make the C-Media chip work on 11i3 LITE, we decided to set alsa settings to max by deleting
[asound.state](projects/epc/audio-engine/system/asound.state), deleting it from [CMakeLists.txt](https://github.com/nonlinear-labs-dev/C15/blob/a0d00466c1e82c28dbac1a0509e80ebfbfe04cd8/projects/epc/audio-engine/CMakeLists.txt#L67) and replacing it by an editing  [../alsa/init/default](https://github.com/nonlinear-labs-dev/C15/blob/f227773e4ae268e39093e5971f4904731d5d2100/build-tools/pacman/c15/dockerized-create-c15-rootfs.sh#L43)

Should we do the same here?

AntonSchmied:
We should think about merging this into milestone-maintenance instead, for better testing. Let's discuss it tomorrow w/ @schlenkibus

AntonSchmied:
could we also add showing the amount of periods in the UI->System Info? Something like
`Buffer underruns: 0 @ 3p x 49fpp`

PR #3530, opened 2023-01-17
WebUI: fix Flanger VoiceGroup I/II Display

### LINKED ISSUE #3527: WebUI: Flanger has no Voice-Group Indication

schlenkibus:
fixup order of monophonics -> the ids were not sorted and this lead to bugs with using binarySearch, closes #3527


MatthiasSeeber (approved):
In Single Sounds, the WebUI Flanger Group now indicates I/II in the Group label.

During testing, I noticed that the I/II indication in Group labels behaves a little inconsistent (maybe worth a ticket?):
- Single: VoiceGroup-bound parameter groups (FX) show I/II, corresponding to the selected VG, non-bound global (Master ...) and Poly groups (Env ...) have no indicator
- Dual: VoiceGroup-bound and non-bound parameter groups all have no indicator

PR #3531, opened 2023-01-17
added/fixup single-sound compare

### LINKED ISSUE #3516: Compare Funktion zeigt nicht beide FX Gruppen an
Es werden nur FX Gruppe 1 angezeigt (ohne Nummerierung)

Version: 23-01 (e24b4)

schlenkibus:
poly-parametergroups will always show part I for single-presets and single-edibuffers, even when part-II is selected in the dropdown, which will now always be displayed, the monophonic parameters are displayed for each part individually, closes #3516


schlenkibus:
- [ ] a dropdown in Single appears only if there is actually a difference in the FX II (always display)

MatthiasSeeber (commented):
> a dropdown in Single appears only if there is actually a difference in the FX II (always display)
I always see a dropdown. Is this intended?

Otherwise, I think it's good enough for now.

PR #3552, opened 2023-01-24
for better ouput and code readability

AntonSchmied:


projects/bbb/scripts/c15-unit-check.sh
```diff
@@ -360,18 +356,18 @@ while getopts ":hgbs" option; do
             [ -z "${NUMARR[0]}" ] && DURATION=14400 || DURATION="${NUMARR[0]}"
             [ -z "${NUMARR[1]}" ] && FACTOR=60 || FACTOR="${NUMARR[1]}"
             LOG_FILE="/tmp/buffer_test.log"
-            rm $LOG_FILE; touch $LOG_FILE
+            rm $LOG_FILE &> /dev/null; touch $LOG_FILE
```
hhoegelo, line 359: not sure about the '&', at least for redirecting the output it is not needed
I am afraid this line could as well mean:
"remove the file in background and the same time touch the file" which would introduce a race condition.

hhoegelo (approved):

PR #3553, opened 2023-01-24
initial commit for convert from single rules with only FX I.

### LINKED ISSUE #3538: Convert-to-Layer und Convert-to-Split für alte Single-Presets "with FX I only"
Siehe Spalte E und F von
https://docs.google.com/spreadsheets/d/1JyMu0EROOBc1nT2WSHSTPJ9n2QGi8fkrHgzSsLcdYoU/edit#gid=289602573

Es werden zwei zusätzliche Zeilen im Sound-Edit-Menü des Single-Sounds (BOLED) und im Convert-Menü des Soundtabs (WebUI) gebraucht. Die ersten vier Einträge sehen dann so aus:

Convert to Layer (FX I only)
Convert to Layer (both FX)
Convert to Split (FX I only)
Convert to Split (both FX)

### LINKED ISSUE #3540: Sound-Edit-Menü und Sound-Tab-Convert-Menü: "Convert to Single (both FX)" statt "Convert to Single"

schlenkibus:
closes #3540
closes #3538

### DIFF
--- a/projects/epc/playground/src/presets/EditBuffer.cpp
+++ b/projects/epc/playground/src/presets/EditBuffer.cpp
@@ -1307,6 +1307,17 @@ void EditBuffer::initCrossFBExceptFromFX(UNDO::Transaction *transaction)
   }
 }
 
+void EditBuffer::initCrossFB(UNDO::Transaction *transaction)
+{
+  for(auto vg : { VoiceGroup::I, VoiceGroup::II })
+  {
+    for(auto param : getCrossFBParameters(vg))
+    {
+      param->loadDefault(transaction, Defaults::FactoryDefault);
+    }
+  }
+}
+
 void EditBuffer::undoableConvertSingleToSplit(UNDO::Transaction *transaction, VoiceGroup copyFrom)
 {
   setVoiceGroupName(transaction, getName(), VoiceGroup::I);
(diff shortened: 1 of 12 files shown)


projects/epc/playground/src/presets/EditBuffer.cpp
```diff
@@ -2180,4 +2191,33 @@ void EditBuffer::copyToFXAndFxFrom(UNDO::Transaction *transaction, VoiceGroup co
...
+}
+
+void EditBuffer::undoableConvertSingleToDualWithFXIOnly(UNDO::Transaction *transaction, SoundType type)
+{
+  SendEditBufferScopeGuard scopeGuard(transaction, true);
+  undoableSetTypeFromConvert(transaction, type);
+  setVoiceGroupName(transaction, getName(), VoiceGroup::I);
+  setVoiceGroupName(transaction, getName(), VoiceGroup::II);
+
+  {
```
hhoegelo, line 2203: do the braces and the 'using' serve any purpose?

MatthiasSeeber: I think they don't. C15::Descriptors is not useful here and transaction exists before and after the braced scope.

MatthiasSeeber (approved):
It seems to work as intended.

I noticed a possible quirk in the conversion rules regarding Master FX Pan and Serial FX - which may lead to a follow-up ticket, if considered worth it.

projects/epc/playground/src/presets/EditBuffer.cpp
schlenkibus: its a relic, will have to remove, there used to be a ScopedLock here

MatthiasSeeber:
here's the follow-up ticket: https://github.com/nonlinear-labs-dev/C15/issues/3560

MatthiasSeeber (changes requested):
- [x] convert Single to Split/Layer (FX I only) correctly defaults Master Serial FX and FX Pan
- [x] convert Single to Split/Layer (both FX) leaves Master Serial FX and FX Pan intact
- [ ] convert Split/Layer to Single (both FX) incorrectly defaults Master Serial FX and FX Pan

MatthiasSeeber (approved):

PR #3561, opened 2023-01-25
Boled: fix animation for sound-edit convert and copy-fx entries

### LINKED ISSUE #3544: BOLED: Sound Edit Menü: Es fehlen Animationen für die Convert- und Copy-Aktionen und Enter-Pfeile für die Copy-Zeilen.

schlenkibus:
fixup all current animated sound-edit-menu items to show the animation before closing/changing the screen, added return symbol at the end of the lines of copy FX I/II items, closes #3544

### DIFF
--- a/projects/epc/playground/src/proxies/hwui/descriptive-layouts/concrete/sound/menus/items/ConvertToSoundTypeItem.cpp
+++ b/projects/epc/playground/src/proxies/hwui/descriptive-layouts/concrete/sound/menus/items/ConvertToSoundTypeItem.cpp
@@ -8,12 +8,14 @@
 #include "ConvertToSoundTypeItem.h"
 #include "use-cases/SettingsUseCases.h"
 #include "use-cases/EditBufferUseCases.h"
+#include "proxies/hwui/descriptive-layouts/ConditionRegistry.h"
 
 ConvertToSoundTypeItem::ConvertToSoundTypeItem(const Rect& rect, SoundType toType)
     : AnimatedGenericItem(nltools::string::concat("Convert to ", toString(toType)), rect,
                           OneShotTypes::StartCB(
                               [=]
                               {
+                                ConditionRegistry::get().lock();
                                 EditBufferUseCases useCases(*Application::get().getPresetManager()->getEditBuffer());
                                 auto selectedVG = Application::get().getVGManager()->getCurrentVoiceGroup();
 
@@ -37,10 +39,13 @@ ConvertToSoundTypeItem::ConvertToSoundTypeItem(const Rect& rect, SoundType toTyp
                                   default:
                                     break;
                                 }
-
+                              }),
+                          OneShotTypes::FinishCB(
+                              []()
+                              {
                                 SettingsUseCases sUseCases(*Application::get().getSettings());
                                 sUseCases.setFocusAndMode({ UIFocus::Sound, UIMode::Select, UIDetail::Init });
-                              }),
-                          OneShotTypes::FinishCB([]() {}))
+                                ConditionRegistry::get().unlock();
+                              }))
 {
 }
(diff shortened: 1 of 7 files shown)


MatthiasSeeber (approved):

schlenkibus:
TODO: check if we can circumvent the unlocking by for example selecting a parameter before the animation completes

davidm2894:
> TODO: check if we can circumvent the unlocking by for example selecting a parameter before the animation completes

if you select a parameter before the animation completes, the screen changes but the actual action (converting for example) has already happened by then.

However, I'm not sure if it's that what you wanted to know.

schlenkibus:
> > TODO: check if we can circumvent the unlocking by for example selecting a parameter before the animation completes
> 
> if you select a parameter before the animation completes, the screen changes but the actual action (converting for example) has already happened by then.
> 
> However, I'm not sure if it's that what you wanted to know.

I suspect that the DescriptiveLayouts won't behave after the steps you describe. For example the sound type won't be properly reflected in the sound-layout when it changes. 

This should happen because the animation finish callback won't ever be called and the condition-registry keeps 'locked' meaning won't propagate changes.

projects/epc/playground/src/proxies/hwui/descriptive-layouts/concrete/sound/menus/items/ConvertToSoundTypeItem.cpp
```diff
@@ -79,11 +87,13 @@ ConvertToSoundTypeItemWithFX::ConvertToSoundTypeItemWithFX(const Rect& rect, Sou
                                   default:
                                     break;
                                 }
-
+                              }),
+                          OneShotTypes::FinishCB(
+                              []()
+                              {
+                                ConditionRegistry::get().unlock();
```
hhoegelo, line 94: Is it correct that you release the lock in this class before setFocusAndMode, whereas you do it vice versa in ConvertToSoundTypeItem?

projects/epc/playground/src/proxies/hwui/descriptive-layouts/concrete/sound/menus/items/ConvertToSoundTypeItem.cpp
```diff
@@ -47,21 +52,24 @@ ConvertToSoundTypeItem::ConvertToSoundTypeItem(const Rect& rect, SoundType toTyp
...
 }
 
-
 ConvertToSoundTypeItemWithFX::ConvertToSoundTypeItemWithFX(const Rect& rect, SoundType convertTo)
 
     : AnimatedGenericItem(nltools::string::concat("Convert to ", toString(convertTo), " (FX I only)"), rect,
                           OneShotTypes::StartCB(
                               [=]
                               {
+                                ConditionRegistry::get().lock();
```
hhoegelo, line 72: I think we have a serious problem with AnimatedGenericItem:
If you manage to destroy the item while the animation is still running (for example by pressing some button fast enough), the finishedCB will never be executed. This might be ok, if the finishedCB would only trigger actions, but if - like here - counters are incremented/decremented, this will cause irreparable invalid state (lock will be held forever).
I see three possible solutions:
- avoid mandatory actions in the finishCB
- add a 'catchAllCB' that is unconditionally executed in the AnimatedGenericItem destructor  
- have an object representing the locked state (will release the lock in its destructor) and capture this object in the finishedCB-lambda: ```OneShotTypes::FinishCB([lock = ConditionRegistry::get().createLock()](){ ...})```. Once the AnimatedGenericItem gets destroyed, it will destroy its finishedCB and thus release the lock.

schlenkibus: i decided to add a optional catchAllCB to the animator which gets executed when the animator is destroyed

hhoegelo (approved):

PR #3606, opened 2023-02-02
Preset-Search: preserve white-spaces for long preset-names

### LINKED ISSUE #3589: Preset-Search: Presets sometimes don't have proper white-spaces
Leerzeichen Preset Search

schlenkibus:
changed white-space to break-spaces to preserve space between first and second preset part of preset-name lengths over 10 characters. closes #3589

### DIFF
--- a/projects/web/client/nonmaps-preset-search/search.less
+++ b/projects/web/client/nonmaps-preset-search/search.less
@@ -599,7 +599,7 @@ body {
 
                 .name-first {
                     display: inline-block;
-                    white-space: break-spaces;
+                    white-space: pre;
                     overflow: hidden;
                     text-overflow: ellipsis;
                     vertical-align: middle;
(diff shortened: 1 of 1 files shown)


projects/web/client/nonmaps-preset-search/search.less
```diff
@@ -599,7 +599,7 @@ body {
 
                 .name-first {
                     display: inline-block;
-                    white-space: nowrap;
```
hhoegelo, line 602: I see and understand the problem but I am really nervous with this solution. 
The table here https://developer.mozilla.org/en-US/docs/Web/CSS/white-space?retiredLocale=de#syntax shows, that "nowrap" is different to "break-spaces" in every single behavior. I am not saying that the proposed solution does not work, though, it just feels strange.
Did you carefully test it with long/short/empty/line-break-containing/etc strings?

hhoegelo: I found some problems with 'break-spaces' - it keeps line breaks and the same problem exists with name-second. I found, that "pre" for both fields seems to work.

hhoegelo (approved):

PR #3608, opened 2023-02-02
WebUI: Preset-Compare use 'FX I/II' for single-sounds in drop-down instead of 'Part I/II' 

### LINKED ISSUE #3592: WebUI: Single-Sound-Compare: "Part I" should rather be "FX I"

schlenkibus:
renamed part I/II to FX I/II for preset-compare, closes #3592


MatthiasSeeber (approved):
FX I/II is now correctly shown if Editbuffer or selected Preset is of type Single.

Note: The implementation checks for both Preset "slots" if the Editbuffer is involved in the comparison. As far as I can tell, when comparing a Preset to the Editbuffer, the Editbuffer will always show in the second "slot" (matching the grammar of "Compare to Editbuffer"). This means that the first "slot" technically does not need to perform the nullptr check, there will always be a Preset ptr. _(I don't mind keeping this code though, better safe than sorry?)_

PR #3640, opened 2023-02-21
WebUI/HWUI: added auto-generated properties/hashtags for presets

### LINKED ISSUE #3448: BOLED: Preset Info: Preset-Namen in den Titel schreiben, automatisch generierte Hashtags für Type etc. in die erste Zeile
Automatische Hashtags: #Layer, #Split, #mono, #unison, #scale

Der Preset-Name wäre im Header dann wieder ohne "u" und "m".

// Damit ist der Header konsistent mit der Parameter Info, wo im Header der Parameter Name steht.

// Die Suche müssen wir so erweitern, dass nach Auto-Hashtags gesucht werden kann.
// Möglichkeit für Negierung bei Sucheingabe wäre schön. (z.B. "Flute NOT #mono")
// oder die Suche auch nach #poly (= NOT #mono), #Single (= NOT(#Layer OR #Split))

schlenkibus:
closes #3448

| property  | inverse property | explanation |
| ------------- | ------------- | ------------- |
| #soundtype | - | preset-type |
| #Mono | #Poly  | "mono" when mono-enable is set to on|
| #Unison  | #`1-voice`  | "unison" when unison-voices is > 1 |
| #Scale | #`12-TET` | "scale" when any scale-offset is set |


MatthiasSeeber (approved):
- [x] search option seems to work
- [x] auto-generated hashtags seem to work

Here i am not sure:
- normal text search uses smallest possible matches (so even fragments of words can lead to results)
  - is this really the right approach for hashtags also? (searching for `#Sc` for example will match all `#Scale` presets)
- matches are case-insensitive (so why capitalize hashtags?)
- split sounds present a conundrum:
  - as there are two parts with independant unison/mono, a split preset can be both Mono and Poly at once, or both Unison and 1-voice. Currently, the "Non-Default" Tags win and searching for f.e. Poly with a Split Preset that has a mono part will not match.

schlenkibus:
TODO: remove from real attributes, and write the hastags into writeDocument and serialize on demand

hhoegelo (approved):
Can be merged, if you like/use my suggested alternative string building algorithm or not.

davidm2894:
Teilweise gibt es einen Zeilenumbruch in den Hashtags, wenn das Fenster nicht breit genug ist (WebUI). Das sieht nicht besonders schön aus, vielleicht könnte man das Fenster etwas breiter machen?

PR #3665, opened 2023-03-02
HWUI: added setup fine support

### LINKED ISSUE #3661: Boled: Setup Value-Edits do not show if 'Fine' is enabled
- allow enable/disable of Fine inside the Setup
- show 'F'
- review all value displays if they are able to show all fine-steps

schlenkibus:
added proper support for Fine in Setup, removed explicit classes for SettingItems, fixup value-ranges for some value-type-settings, remove invalid module-caption entry for single-sound global groups, closes #3661


projects/epc/playground/src/groups/ParameterGroup.cpp
```diff
@@ -249,6 +249,11 @@ bool ParameterGroup::isPolyphonic() const
   return m_parameters.begin()->isPolyphonic();
 }
 
+bool ParameterGroup::isMonophonic() const
+{
+  return m_parameters.begin()->isMonophonic();
```
hhoegelo, line 254: In case we have empty parameter groups (sounds stupid but I can image instrument states when this is the case) - this function would just crash. Consider checking for empty groups.

hhoegelo (approved):

projects/epc/playground/src/groups/ParameterGroup.cpp
schlenkibus: i introduced an assertion for now, if we need empty groups in the future we will have to revisit this assertion and change it to a normal check

PR #3669, opened 2023-03-07
Issue 3654

schlenkibus:


projects/epc/playground/src/sync/SyncMaster.cpp
```diff
@@ -72,18 +72,25 @@ nlohmann::json SyncMaster::api(const ClientPtr &client, const nlohmann::json &in
 
 void SyncMaster::add(SyncedItem *item)
 {
-  m_items.push_back(item);
-  item->setDirty();
+  auto it = std::find(m_items.begin(), m_items.end(), item);
```
hhoegelo, line 75: I think we can just say m_items.insert(item) here.

projects/epc/playground/src/sync/SyncMaster.cpp
```diff
@@ -72,18 +72,25 @@ nlohmann::json SyncMaster::api(const ClientPtr &client, const nlohmann::json &in
...
+  {
+    m_items.emplace(item);
+    item->setDirty();
+  }
 }
 
 void SyncMaster::remove(SyncedItem *i)
 {
-  m_items.erase(std::remove(m_items.begin(), m_items.end(), i), m_items.end());
+  auto existingIt = m_items.find(i);
```
hhoegelo, line 86: for unordered_set, we can just say m_items.erase(i), I think.

hhoegelo (approved):

PR #3689, opened 2023-03-16
WebUI: fix dimension and step-size for Tune-Reference

### LINKED ISSUE #3687: WebUI: Tune-Reference is not showing decimal places

schlenkibus:
fixup dimension display for tune reference webui and fixup step size for incremental (with arrow keys) steps, closes #3687


MatthiasSeeber (approved):
- the dimension seems to be set up correctly (one decimal place, coarse 1 Hz, fine 0.1 Hz increments - decimal is always shown in HWUI, in WebUI nonzero decimals are shown and `.0` is truncated/trimmed)
- arrow keys on the WebUI slider are working (but not 100%-reliably - more than ten steps are needed to move by 1.0 Hz)
- the WebUI slider (arrow key and mouse actions) resolves fine values and coarse/fine distinction is not really possible with this implementation, it is hard/impossible to set the slider to 440.0 Hz by mouse only

Concludingly, I would say it works (more or less, with some inconveniences) and is good enough?

hhoegelo (approved):

PR #3692, opened 2023-03-22
Issue 3690 fix

schlenkibus:
added argument to sendParameterMessage to indicate wether or not the
event should be send via midi. this argument is always true except for the case where a ribbon is changed via the bound macro-control. then we dont want to send the value, but we have to update the latched values (but I have not yet understood why)

### DIFF
--- a/projects/epc/playground/src/parameters/messaging/ParameterMessageFactory.h
+++ b/projects/epc/playground/src/parameters/messaging/ParameterMessageFactory.h
@@ -9,19 +9,19 @@
 namespace ParameterMessageFactory
 {
 
-  template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *);
-
+  template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *, bool sendAsMidi);
 };
 
 template <>
 inline auto ParameterMessageFactory::createParameterChangedMessage<C15::Descriptors::ParameterType::Hardware_Source>(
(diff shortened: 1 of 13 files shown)


projects/epc/playground/src/parameters/messaging/ParameterMessageFactory.h
```diff
@@ -9,27 +9,27 @@
 namespace ParameterMessageFactory
 {
 
-  template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *);
-
+  template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *, bool sendAsMidi);
```
MatthiasSeeber, line 12: Isn't it unfortunate that all ParameterMessages now expect a bool to determine if MIDI should be sent, even though this is only intended for HardwareSources?

Wouldn't it be possible to have a separate overload?

~~~ c++
template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *); // any parameter except hw sources
template <C15::Descriptors::ParameterType, typename T> auto createParameterChangedMessage(const T *, bool); // hw sources
~~~

MatthiasSeeber (approved):
- [x] a Ribbon changed by MC (f.e. via a Pedal) doesn't send MIDI
- [x] Local On, MIDI Loopback: Ribbon touchstrip or slider changes don't cause event cascades and glitchy behavior anymore

hhoegelo (approved):

PR #3702, opened 2023-03-28
fixup: access to hashtags now works again

### LINKED ISSUE #3700: WebUI: Preset-Search with new properties is not working
**Describe the bug**
When searching for '#Mono' in the new preset-search with proprties no preset seems to match on @nonlinear-labs pc, on my device this bug is not reproduceable 

**C15 Software Version as found in systeminfo**
main 23-12

**To Reproduce**
1. Open Preset-Search
2. enable 'Properties'
3. type `12-TET' or with hashtag
4. See no matched presets

**Expected behavior**
expect all presets to show up that are not changed in the Scale (offsets) Group, which would be most of them

schlenkibus:
feature now works for all preset-managers, as I developed this feature i had the hashtags/properties as part of the attributes for a short while, then i removed that but did not pay attention that my local preset-manager still had those attributes. local testing suggested that the preset-search by hashtags was working, but it was only working for the old preset-manager. closes #3700

### DIFF
--- a/projects/epc/playground/src/presets/Preset.cpp
+++ b/projects/epc/playground/src/presets/Preset.cpp
@@ -378,7 +378,7 @@ nlohmann::json Preset::serialize() const
            { "part-I-name", getVoiceGroupName(VoiceGroup::I) },
            { "part-II-name", getVoiceGroupName(VoiceGroup::II) },
            { "attributes", AttributesOwner::toJson() },
-           { "Hashtags", getHashtags() } };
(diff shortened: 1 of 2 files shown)


MatthiasSeeber (commented):
After restoring from backup (created: 2023-01-19), I wasn't able to succsefullly search for auto-generated Hashtags.

MatthiasSeeber (approved):
Sorry, my bad. I didn't notice that the "Properties" checkbox wasn't enabled.

PR #3717, opened 2023-03-30
Issue 3708

MatthiasSeeber:


projects/web/client/nonmaps-preset-search/search.ts
```diff
@@ -121,19 +111,83 @@ function getPresetSortResult(lhs: any, rhs: any): number {
...
+        // when any provided function returns a match, the search is successful
+        return preset => queryCbs.some(cb => cb(preset));
+    });
+}
+
+function prepareSearchFilter(colors: StringArray, opt: SearchOptions, query: PresetMatchCbArray): PresetMatchCb {
+    // prepare one filter callback, omitting unnecessary filterings
+    const filterCbs: PresetMatchCbArray = [
+        // preset has to be valid (?)
+        preset => (preset && preset['name']) ? true : false,
```
MatthiasSeeber, line 160: This still bothers me. It seems like we require the "name" field to be present in `preset`. But if the field is an empty string, it will fail.. _(currently, we technically allow for an empty preset name)_

It might be better to refactor the test into

~~~ ts
preset => preset && ('name' in preset)
~~~

projects/web/client/nonmaps-preset-search/search.ts
```diff
@@ -121,19 +111,83 @@ function getPresetSortResult(lhs: any, rhs: any): number {
...
+                ...(opt.searchInComment ? [
+                    preset => preset['attributes']['Comment']
+                        && preset['attributes']['Comment'].toLowerCase().split(whitespaceRegex)
+                            .some(word => hashtagRegex.test(word) ? false : word.includes(v))
+                ] : [])
+            ]);
+        }
+        // if no option checkbox is enabled, it should be considered a match (similar to empty query)
+        if(queryCbs.length === 0) return _ => true;
+        // when any provided function returns a match, the search is successful
```
MatthiasSeeber, line 151: This is just a proposal and would break with former behavior (currently: no checkbox enabled, no results whatsoever).

MatthiasSeeber (commented):
- [x] test: hashtags work out-of-the-box, option is gone

PR #3725, opened 2023-04-04
added missing explicit parameter-types (for unison)

### LINKED ISSUE #3720: HWUI: Carousel for Unison-Group not working in Part II of Layer
when other parameter than voices is selected the small slider values in the carousel display the values of part II

schlenkibus:
fixes #3720


MatthiasSeeber (approved):
- [x] boled: traserving caroussel always shows part I parameters in unison group (also if part II is selected), webui as well

PR #3729, opened 2023-04-05
MIDI: only send active-sensing after midi message from PG got received

### LINKED ISSUE #3728: MIDI: when AE is starting we send Active-Sensing until the playground settings are received
this is a bug in the case that the playground disabled active-sensing

schlenkibus:
added boolean flag that will be set when midi-settings are received, and if we did not receive the settings from the playground yet we do not send active-sensing, closes #3728


MatthiasSeeber (approved):
Not tested, but code looks convincing.

hhoegelo (approved):

PR #3761, opened 2023-05-16
fix undo window problems: 3758

### LINKED ISSUE #3758: WebUI: Resize Window problem
**Describe the bug**
If you drag a window (e.g. Undo History) so large that it goes over the Belt, the window gets (additional) scroll bars outside the actual window. This cannot be undone, also because the drag handle no longer reacts. It also crashed once when I dragged the window and then refreshed the website, but this may be coincidentally.

**BOLED or WebUI**
WebUI

**C15 Software Version as found in systeminfo**
next-milestone
eb4aec607c18e926bfc7284a0bd4fb240a25eb05
(but can also be observed on main)

**Expected behavior**
Maybe we just block the window on the belt borders? On Safari it is already blocked, but the additional scrollbars are also here.

**Screenshots**
Bildschirmfoto 2023-05-16 um 11 34 20
Bildschirmfoto 2023-05-16 um 11 34 08

**WebUI Metadata (please complete the following information):**
 - OS: MacOS
 - Browser: chrome, firefox, safari

schlenkibus:
remove undo window height limit, but prevent the inner height to be set larger than the inner-viewport-height, closes #3758


package/nonmaps/src/main/webapp/NonMaps.css
```diff
@@ -1975,4 +1975,8 @@ we have to use !important here as the styles get set from within the GWTDialog
 
 .preset-property-label-lowlight {
   color: #939393;
+}
+
+.undo-window-fix {
```
hhoegelo, line 1980: Maybe I am missing something, but as fas as I can see we do not need the findDIVByClassName foo in UndoTreeWindow. Instead we could re-phrase this rule like so:

div.undo-tree-window div.dialogMiddleCenterInner {
    max-height: unset !important;
}

schlenkibus: your suggestion seems to work!

hhoegelo (approved):

PR #3766, opened 2023-05-17
Setting Conversion

schlenkibus:
added new version -> one that indicates that the times were reset, introduced a call to a virtual function in each setting that allows conversions to take place. right now only the transition time and the edit smoothing time will get 'converted' to a fixed value, closes #3763


package/playground/src/device-settings/EditSmoothingTime.cpp
```diff
@@ -8,13 +8,23 @@
...
     : super(parent)
     , m_time(nullptr, ScaleConverter::get<EditSmoothingTimeMSScaleConverter>(), 0, 100, 1000)
 {
 }
 
+Glib::ustring EditSmoothingTime::convert(const Glib::ustring &text, int version)
+{
+  if(version < (int)VersionAttribute::History::FIXUP_SETTING_CONVERSION_SMOOTHING_AND_TRANSITION_TIME)
+  {
+    return "0.01";
```
MatthiasSeeber, line 23: I may be wrong, but this is no conversion, but rather a hard "default". So no matter what the setting was previously, it is ignored, as the "default" is returned - I think.

package/playground/src/device-settings/TransitionTime.cpp
```diff
@@ -8,13 +8,23 @@
...
     : super(parent)
     , m_time(nullptr, ScaleConverter::get<EnvelopeAttackDecayTimeMSScaleConverter>(), 0, 100, 1000)
 {
 }
 
+Glib::ustring TransitionTime::convert(const Glib::ustring &text, int version)
+{
+  if (version < (int)VersionAttribute::History::FIXUP_SETTING_CONVERSION_SMOOTHING_AND_TRANSITION_TIME)
+  {
+    return "0.0";
```
MatthiasSeeber, line 23: Same thing - I think.

MatthiasSeeber (commented):
I'm not sure about the approach here..

hhoegelo (approved):

PR #3773, opened 2023-05-24
C16 UI

hhoegelo:
not really needed, but maybe it's time to start more serious development with issues etc and PRs to next-milestone.


os/pi4/factory-img/tweakFactoryImage.sh
```diff
@@ -0,0 +1,110 @@
...
+            ! sed -i '13iExecStartPost=/bin/sh -c "systemctl restart udhcpd"' /tmp/a
+            upload /tmp/a /usr/lib/systemd/system/hostapd.service
+            
+EOF
+    
+    fi
+}
+
+copyOut
+tweakBootPartition
```
schlenkibus, line 109: why would we call 'tweakBootPartition' twice?

hhoegelo: do we?

PR #3778, opened 2023-05-31
issue-3774: Allow concurrent continuous transactions

hhoegelo:
With c16 we will have the use case of simultanuous changes of different parameters. Currently, our undo scope closes continuous transaction of parameter A as soon, as parameter B is edited. When the user turns two ERPs the same time, this will produce thousands of transactions per second.
This patch will try to find a matching continuous transaction for the current parameter change by walking up the current undo stack. 'Matching' means:
- the parameter (cookie) of the transaction is identical to the current one
- the transaction on stack is not older than 2s

As soon as we find a 'normal' (i.e. not-continuous) transaction when walking the stack, we stop searching as we assume that this transaction is 'more important', e.g. a preset load.

Also, I removed the unnecessay 'libundo' folder, as the undo system is actually not a library but origin playground code.

Also, I removed the unused nlrotary folder.


hhoegelo:
@MatthiasSeeber If accepted, the documentation may have to be updated accordingly.

hhoegelo:
@davidm2894 In file 

https://github.com/nonlinear-labs-dev/C15/pull/3778/files#diff-007f3f956775fc51832dd4c54e90237ab44f3cc2d795021fedfc282cacf643f4 

you find the test case "Continuous Transactions" describing the suggested new behavior. Please check if this is what we want and if the drawbacks are acceptable.

MatthiasSeeber (approved):

schlenkibus:
I want to propose a minor change:
When we change one parameter continuously one transaction will be created every 2 seconds. Even if the transaction could possibly be squashed into one undo-step. 

We could refresh the `age` of a `ContinuousTransaction` on every `setClosingCommand` which would in turn prevent the behavior described above, as we stop updating the last matching transaction only when no new value has been received for more than 2 seconds instead of always creating a new transaction after 2 seconds.

PR #3787, opened 2023-06-08
introduce type ParameterNumber (which is C15::PID::ParameterId)

hhoegelo:
* tweak ::ParameterId => be clear about the types: Use ParameterNumber instead of plain int
* replace all usages of int by new ParameterNumber type


package/playground/src/testing/unit-tests/parameters/ParameterIdTests.cpp
```diff
@@ -12,7 +12,7 @@ TEST_CASE_METHOD(TestHelper::ApplicationFixture, "ParameterId hashes")
     {
       for(int i = 0; i < 400; i++)
```
MatthiasSeeber, line 13: :warning: the value is still hard-coded, and we did overshoot the 400 limit some time ago (highest number is currently 428 for Master FX Mix)

I'd suggest to use `C15::Config::tcd_elements` instead, it can be found in `<generated/c15_config.h>`.

MatthiasSeeber (approved):

PR #3796, opened 2023-06-14
add method to guess parameter id from voice group, sound type and par…

hhoegelo:
…ameter number

remove syncParts-feature
re-add dependecy from local playground target to web-targets to pass the version check on debugging


package/playground/src/ParameterId.cpp
```diff
@@ -48,6 +54,78 @@ ParameterId::ParameterId(const Glib::ustring &str)
...
+}
+
+VoiceGroup ParameterId::parseGroup(const std::string &in)
+{
+  auto p = in.find('-');
+
+  if(p == std::string::npos)
+  {
+    auto n = parseNumber(in);
+    return ParameterId::isGlobal(n) ? VoiceGroup::Global : VoiceGroup::I;
```
schlenkibus, line 74: i don't think that there exists any sensible default. 
Are we sure that we want to return some VoiceGroup for example for the string: `12` ?

schlenkibus (approved):

PR #3805, opened 2023-06-21
Midi Settings - Reset Hotfix

### LINKED ISSUE #3803: Midi: send all notes Off when midi send or receive channel was changed
we had one hung note while testing split sounds with external midi input

MatthiasSeeber:
When Midi Settings (f.e. Send/Receive channels) change, reset behavior did not check for active Midi Keys (only checking internal Keys), leading to hanging notes. This has been fixed, checking for active Keys can now optionally also take Midi Keys into account.

The Midi Settings Reset behavior is now "broad and lazy": if ANY Key is pressed, a reset occurs internally (resetting VoiceAlloc and Envelopes) AND externally (Midi Out AllNotesOff) without further context distinction.

closes #3803


projects/epc/audio-engine/src/synth/c15-audio-engine/voice_allocation.h
```diff
@@ -358,6 +359,7 @@ class VoiceAllocation
         // confirm
         keyDown_confirm(keyState);
         m_internal_keys.m_global += isInternalKey;
+        m_external_keys.m_global += !isInternalKey;
```
schlenkibus, line 362: Maybe introducing a local variable `isExternalKey` would help remove one layer of indirection (`!isInternalKey` leads to some friction in my head)

projects/epc/audio-engine/src/synth/c15-audio-engine/dsp_host_dual.h
```diff
@@ -181,7 +189,7 @@ class dsp_host_dual : public DSPInterface
   using SimpleRawMidiMessage = nltools::msg::Midi::SimpleMessage;
   float getReturnValueFor(HardwareSource hwid) override;
   void resetReturningHWSource(HardwareSource hwui) override;
-  bool areKeysPressed(SoundType _current) override;
+  bool areKeysPressed(SoundType _current, const IncludeKeyOrigin _include_origin = IncludeKeyOrigin::Internal) override;
```
schlenkibus, line 192: I thought we would introduce a simplified and explicit interface for the checking of external/internal keys.
I was thinking about an interface like `areInternalKeysPressed` and `areExternalKeysPressed`
this would keep more simple than an interface with default arguments.

projects/epc/audio-engine/src/synth/input/InputEventStage.cpp
```diff
@@ -1135,7 +1135,10 @@ bool InputEventStage::didRelevantSectionsChange(const InputEventStage::tMSG &msg
 
 void InputEventStage::onMidiSettingsMessageWasReceived(const tMSG &msg, const tMSG &oldmsg)
 {
-  if(didRelevantSectionsChange(msg, oldmsg) && m_dspHost->areKeysPressed(m_dspHost->getType()))
+  using IncludeKeyOrigin = DSPInterface::IncludeKeyOrigin;
+  // broad and lazy approach: resets internally and externally when any (internal or external) key is pressed
+  if(didRelevantSectionsChange(msg, oldmsg)
```
schlenkibus, line 1140: The code seems to conflict with the comment:
In the comment it says 'when any (internal or external) key is pressed' but the function interface would suggest that we are only checking for ::External keys.

MatthiasSeeber: IncludeExternal suggests to you that only external is counted?

MatthiasSeeber:
Nevermind.

PR #3812, opened 2023-06-27
Change backup filename

schlenkibus:


projects/shared/configuration/process.sh
```diff
@@ -2,6 +2,7 @@
 
 sourceDir=$1
 binaryDir=$2
+isDevPC=$3
```
MatthiasSeeber, line 5: This was recently removed, as the parameter is unused within the script.
Is there a reason for this parameter, as it still is unused?

projects/shared/configuration/build-in-docker.sh
```diff
@@ -2,7 +2,15 @@
...
-podman build -t $container $1
-podman run -v "$sourceDir":/sourceDir -v "$binaryDir":/binaryDir $container /sourceDir/process.sh /sourceDir /binaryDir
+if [ "$isDevPC" == "On" ] || [ "$isDevPC" == "ON" ] || [ "$isDevPC" == "on" ]
+then
+  echo "building configuration in podman (Development PC)"
+  podman build -t $container $1
+  podman run -v "$sourceDir":/sourceDir -v "$binaryDir":/binaryDir $container /sourceDir/process.sh /sourceDir /binaryDir
+else
+  echo "building configuration with native tools (Build Update)"
+  /bin/bash "$sourceDir"/process.sh "$sourceDir" "$binaryDir"
```
MatthiasSeeber, line 15: process.sh is called without the introduced `isDevPC` parameter?

MatthiasSeeber (changes requested):
tested webui only
- [x] new backup exports produce new .nlbackup file format
- [x] old .xml.tar.gz and new .nlbackup files can be imported
- [ ] support for zip files seems to be broken now (file type is not included in supported types and will not be visible in dialog, when changing dialog to show all files and selecting a .zip backup, this will result in an error message)

MatthiasSeeber (approved):
tested restore/save in WebUI (download) and BBBB (usb-stick) on ubuntu
- [x] save provides working `.nlbackup`
- [x] restore works with all three file formats (`.xml.zip`, `.xml.tar.gz`, `.nlbackup`), also file filter works as intended (showing all supported types by default)
- currently unclear: WebUI provides `[timestamp]-nonlinear-c15-banks.nlbackup`, whereas BBBB provides `[timestamp]-c15-banks.nlbackup`

hhoegelo (approved):

PR #3830, opened 2023-07-06
issue-1807: fix missing svg image redraws

### LINKED ISSUE #1807: WebUI: Es kommt vor, dass ein Teil der HW Amounts nicht angezeigt wird (leere graue Felder).
Draw-Bug

Mit der nächsten Zoom-Aktion verschwindet das Problem.

hhoegelo:
the load state of the svg image was ignored when calculating the check sum, so CachedBitmapControl had not reason to redraw the children

closes #1807


schlenkibus:
could you describe what kind of bug this might fix? It seems like this could fix some redrawing issues (maybe also the one in the matrix users and colleagues pointed out over the last years)

schlenkibus (approved):

PR #3841, opened 2023-07-19
added python parameter ids enum

schlenkibus:


configuration/lib/tasks/declarations.ts
```diff
@@ -100,7 +103,26 @@ export const DeclarationsParser = new Parser<DeclarationsType>(
...
+                    if((key === "None") || (props !== null))
+                        out.push(key);
+                    return out;
+                }, []).join(",\n"),
+                smoother_section: Object.keys(declarations.smoother_section).join(",\n"),
+                smoother_clock: Object.keys(declarations.smoother_clock).join(",\n"),
+                smoother_scale: Object.keys(declarations.smoother_scale).join(",\n"),
+                display_scaling_type: Object.keys(declarations.display_scaling_type).join(",\n")
+            };
+            return { declarations, enums, python_enums };
```
MatthiasSeeber, line 125: There is no need for this redundancy, generating and returning the same information twice has no benefit.

configuration/lib/build.ts
```diff
@@ -387,6 +387,7 @@ function processDefinitions(result: Result) {
         return out;
     }, []).join("\n");
     result.enums.pid = ["None = -1", ...pid.filter((id) => id !== undefined)].join(",\n");
+    result.python_enums.pid = ["none = -1", ...pid.filter((id) => id !== undefined)].join("\n");
```
MatthiasSeeber, line 390: I would prefer capital case for "None" (consistent with all other enums and enum entries).

PR #3850, opened 2023-08-09
BOLED: fix Voices-Screen incorrectly updates on Sound-Type change

schlenkibus:
original bug-report:
```
Bug:
Voices-Screen hat I/II Button (der auch funktioniert), wenn man von Single Sound auf Layer Sound wechselt (sic. ueber WebUI)
```


MatthiasSeeber (changes requested):
:warning: The Sound Screen is broken. Pressing the Sound button (no matter which Sound Type) in the bbbb-ui application leads to the following screen:

boled-sound-deprecated

MatthiasSeeber (approved):
With the Command Line Argument `-l` _(path/to/C15/package/c15-playground/resources)_, the Sound Screen is available and the Voices Screen doesn't show a Soft Button I/II when in a Single or Layer Sound.

package/c15-playground/src/hwui/descriptive-layouts/ConditionRegistry.cpp
```diff
@@ -44,7 +44,7 @@ sigc::connection ConditionRegistry::onChange(const std::function<void()>& cb)
 void ConditionRegistry::onConditionChanged()
 {
   if(!isLocked())
-    m_signal.deferedSend();
+    m_signal.send();
```
hhoegelo, line 47: maybe this could be a problem?

package/c15-playground/src/hwui/descriptive-layouts/ControlInstance.cpp
```diff
@@ -23,4 +23,15 @@ namespace DescriptiveLayouts
 
     return new GenericControl(*this, eventProvider);
   }
+
+  bool ControlInstance::compare(const ControlInstance& other) const
+  {
```
hhoegelo, line 28: consider using std::tie(foo, bar) == std::tie(other.foo, other.bar);

hhoegelo (approved):

PR #3865, opened 2023-08-16
added on-off 'schalter' using the interface we discussed

schlenkibus:


package/meteor/client/c16/parameters/parameters.ts
```diff
@@ -6,6 +6,7 @@ import { C16ButtonEvent, C16RotaryEncoderChangedEvent, C16RotaryEncoderEvent } f
 import { Button, Rotary } from "../bridge/ui-elements";
 import { EditMode, ParameterMode, bridge, focusAndMode, getCurrentVoiceGroup, playground, proxyFactory } from "../shared-state";
 import "./lib/lib.ts";
+import "./lib/onOff/onOff.ts"
```
hhoegelo, line 9: I think, lib.ts should import lib/onOff/onOff.ts - but this is debatable

PR #3874, opened 2023-08-23
Introduce return values for RPC calls

hhoegelo:
rpc calls have now a serial number, so that return values can be mapped to the pending, result awaiting caller.

* Tests testing reactive stuff have now a ReactiveTest class for better structuring of action, reaction and clean up
* modify test so it tests rpc result
* port typescript tests to new ReactiveTest class
* connect shaker to return value so it reacts immediately after control position change


package/meteor/client/synced-item-database.ts
```diff
@@ -7,29 +7,55 @@ interface SyncedItem {
...
     constructor() {
         console.log("synced item db ctor");
         this.connect();
     }
 
-    public invoke(topic: string, rpcName: string, arg: any) {
-        const call = { call : { [topic]: { [rpcName]: arg }}};
+    private cleanupOutdatedCallbacks() {
+        const now = Date.now();
+        this.pendingCallbacks = new Map([...this.pendingCallbacks.entries()].filter(([_, value]) => (now - value.createdAt) > 5000));
```
schlenkibus, line 38: cleanupOutdatedCallbacks should be only called when the syncedItemDatabase is reconnected (after playground crash/restart for example)
and maybe reuse this function to warn on long pending callbacks instead

schlenkibus (approved):

PR #3879, opened 2023-08-24
Test-Tone Stereo and UI

### LINKED ISSUE #3871: make Test Tone a Setting
## Status Quo

Currently the Test Tone is a developer-only feature, available only as:

- "Hotkey" binding in AudioEngine application
- supported by `AE_PROTOCOL_CMD_IDS` (PlayController to Synth)

Its amplitude is hard-coded as -6 dB_FS and its frequency is bound to the Tuning Reference setting (usually and by default 440 Hz). It currently features three modes:

- Synth-only (default)
- Tone-only (left and right)
- both (Tone and Synth)

## Requested Features

- [x] add left-only and right-only modes for the Test Tone
- [x] expose Test Tone Modes as a Setting in the UI (proposal: Setup > Device Settings)
- [ ] when the Test Tone is active, the reception of any Event (internal or external Key or HW Events, Playground Messages) should let the Synth switch into Synth-only mode (similar to how the Screen Saver is stopped by any user activity)

### Design Decision

- the LPC-Synth Protocol and hotkey bindings in Audio Engine application should be deprecated, as this is will now be exposed in the UI

schlenkibus:
- added test-tone-type setting which is not persisting and decodes foll…owing states (Off, Left, Right, Both) Off meaning synth only and the others describe which stereo channel the test-tone should play on
- added test-tone editor for webui and hwui
- when the webui dialog is closed the setting gets set to Off 
- we can't handle the closing of the webui -> possible to have the setting persisting 
- when the boled menu is closed the setting gets set to Off 
- implement left-right panning for test-tone
- refactor index mechanism in ae_global_section to be more descriptive by using an enum

- no special event connections are implemented (the similarities to the screensaver klaus requested) we want to keep this feature as simple as possible

closes #3871


MatthiasSeeber (changes requested):
Can we talk about this? I see and get the simplicity towards the user, but I'm not really sure if the ticket request is fully met, and I find the approach a little confusing:

- so, we only create ONE enum called `TestToneType`, describing both On/Off and Panning
- the yaml definition also names ONE new setting, calls it Type, but represents it as boolean
- then we package this as two new messages (Type/Pan), each starting a fadeout?
- the UI only auto-disables the test tone on dialog close, the following events do not disable it (but the screen saver is affected):
  - playing keys
  - loading presets (still possible in hwui:soled and webui)
  - chaning parameters (webui, hwui:soled), using hw sources
- the ON/Off switches between Synth-only (OFF) and Synth+Tone (ON), Tone-only is not available
- the hotkey still reacts in the audio-engine application and cycles through the modes without having any effect (I suspect, the same would be true for LPC protocol events)

MatthiasSeeber (approved):

PR #3919, opened 2023-09-21
translate c16 debug ui to react

hhoegelo:


package/meteor/client/c16/bridge/ui-elements.ts
```diff
@@ -21,6 +21,8 @@ export enum Button {
 
     Dec = "Dec",
```
schlenkibus, line 22: please rename left/right

schlenkibus (changes requested):

package/meteor/client/c16/bridge/ui-elements.ts
schlenkibus: done!

schlenkibus (approved):

PR #3924, opened 2023-09-26
react Preset-Manager

schlenkibus:
added basic functionality of preset-layout for c16-hwui, in order to do that implement calls for preset manager and bank (step preset and bank selection) and settings.


package/meteor/client/c16/c16.less
```diff
@@ -14,3 +14,27 @@
     --selected: #fff;
     --scale_color: #ababab;
 }
+
+.selected {
```
hhoegelo, line 18: this rules should  really  apply to every element with 'selected' class  in the whole c16 hwui project?

PR #3927, opened 2023-10-04
preset-search

schlenkibus:
added basic mockup for preset-search dialog content, added bank list with selection, added sort by selector and color-tag area


davidm2894:
- [ ] softbutton 3 should never be highlighted but just toggle between "browse" and "search"
- [ ] Highlight should be on search field when page is accessed.  
- [ ] SB 1 Bank on browse screen and search screen are not the same, so it should only affect their respective page
- [ ] when Bank is active (drop down open), encoder + up/down should work there
- [ ] connect Encoder 
- [ ] selection gets out of list aspect (when stepping up/down). Let's do it like it is in the browser but with the tweak that we don't leave the blank space when we reach the start and end of the list but to move the selector towards start & end.
- [ ] info field is missing (only visible when keyboard collapsed)
- [ ] touch on search field should open keyboard
- [ ] idea for navigation & keyboard hardware control: left + right changes between search field and list. when search field is in highlight, keyboard expands and by pressing shift/func + left/right/up/down we can move through the keyboard (highlights respective character). when shift/func is released, character is selected. and for now, we maybe drop the idea to navigate through the search elements as it could be too confusing. **open for opinions on that.**

hhoegelo (approved):

PR #3943, opened 2023-10-26
Rework bank export

hhoegelo:


hhoegelo:
@schlenkibus The first commit is a clean up and can be more or less ignored. The second one is the one to review.

package/nonmaps/src/main/java/com/nonlinearlabs/client/world/overlay/belt/presets/BankExport.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static void start(String currentName, String uuid) {
+
+        if (lastUsedUrl.length() > 0)
+            freeUrl(lastUsedUrl);
+
+        lastUsedUrl = "";
+
+        String bankName = URL.encodePathSegment(currentName);
+        String uri = "/banks/download-bank/" + bankName + ".xml?uuid=" + uuid;
```
schlenkibus, line 22: We used to pass the <bankName>.xml as a URL path segment for ever now, but judging from the playground-code we never really used it. Maybe we should remove it from the interface as to not confuse us later?

schlenkibus (approved):

PR #3949, opened 2023-11-16
introduce tool for experimenting with alsa-in/alsa-out loopback with …

hhoegelo:
…access

to samplerate, format, num periods, frames per period and streaming mode parameters. The Streaming mode can be --dma, which makes use of the implementation of the zita-alsa-pcm project found here http://kokkinizita.linuxaudio.org/linuxaudio/downloads/. If this argument is missing, blocking pcm_read/pcm_write is done as usual.


package/alsa-in-out/src/main.cpp
```diff
@@ -0,0 +1,251 @@
...
+  auto pcm = std::make_unique<Alsa_pcmi>(playbackDevice.c_str(), captureDevice.c_str(), samplerate, framesPerPeriod,
+                                         numPeriods, static_cast<snd_pcm_format_t>(sampleformat), 1);
+  if(pcm->state())
+    nltools::throwException("Can't open ALSA device");
+
+  if(pcm->ncapt() < 2)
+    nltools::throwException("Capture device is not capabale of handling 2 channels");
+
+  if(pcm->nplay() < 2)
+    nltools::throwException("Capture device is not capabale of handling 2 channels");
```
schlenkibus, line 136: Isn't this the Playback Device?

schlenkibus (approved):

KlausStrohhaecker:
Sorry, I don't understand this complicated ways of doing things. I would assume someone already successfully compiled and tested this so why not send me the standalone excutable rather than forcing me to wastefully repeat steps already done?

hhoegelo:
> Sorry, I don't understand this complicated ways of doing things. I would assume someone already successfully compiled and tested this so why not send me the standalone executable rather than forcing me to wastefully repeat steps already done?

Not sure I can follow. Is this just the rude way of asking if someone (probably Justus or me) can compile the project for you?

KlausStrohhaecker:
Sorry, I don't wanted to come accross as rude, rather frustrated if anything.
I was under the assumption that the code in this PR has been compiled successfully and probably has undergone basic tests.... which means the final executable is _already physically available_ and hence there is no need for me as user to do all the setup and build etc again. Saving time and money.

PR #3950, opened 2023-11-16
Updated macro layout

schlenkibus:


davidm2894:
- [ ] when open rename, keyboard should be open and cursor in text field
- [ ] when rename dialog is open, left/right (+ Encoder) should move text cursor and up/down should navigate between dialog items (during that time no impact on parameter)
- [ ] Add MC Button to Hardware Amount Screen SB 4 (so you can toggle between both pages easily)
- [ ] close rename dialog when touch on Screen-SB 2 (= save)

hhoegelo (approved):

PR #3954, opened 2023-11-22
add undo-functionality

schlenkibus:
added undo/redo call to scope, and connect c16 undo/redo buttons to functionality


package/playground-lib/src/libundo/undo/Scope.cpp
```diff
@@ -349,16 +350,19 @@ namespace UNDO
...
-      });
+      writer.writeTag("undo",
+                      [&]()
+                      {
+                        if(getUndoTransaction() && canUndo())
+                          writer.writeTextElement("undo", StringTools::buildString(getUndoTransaction()));
+                        else
+                          writer.writeTextElement("undo", "0");
+
+                        writer.writeTextElement("redo", StringTools::buildString(getRedoTransaction()));
```
hhoegelo, line 361: why is there extra handling needed for the nullptr of undo transaction, but not for redo? I'd expect either both needs this or none.

davidm2894:
knobs and their functionality seems fine. however, the undo system itselfs seems buggy. value changes are not always according to focus changes. also once a whole series of different parameter changes was just skipped (all parameters of one page, it just skipped to the page before).

PR #3962, opened 2023-12-12
implement Undo Lists with new ScrollableList

schlenkibus:
- [x] undo list implementation complete
- [x] added undo+redo button press at the 'same' time opens undo-layout
- [x] started porting of Bank/Preset Manager and Preset-Search lists to ScrollableList
- [x] added automatic visual studio code open when meteor-full is started (when installed)
- [x] preset and bank lists are no longer fully functional


package/meteor/CMakeLists.txt
```diff
@@ -73,9 +73,14 @@ else()
   
   include(${CMAKE_SOURCE_DIR}/cmake/crossbuild.cmake)  
   crossBuild(NAME web-app-bundle MACHINE web DEPENDS nonmaps mc-view online-help recorder)
-  
+
+  execute_process(
```
hhoegelo, line 77: find_program(VSCODE code)

would be the cmake way of searching for binaries.
Anyway you don't actually seem to use the result, do you?

package/meteor/client/c16/debug/debug.tsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { PointerEvent } from "react";
```
hhoegelo, line 1: this looks a bit odd, do we need the braces, if so, why only for the pointerEvent?

package/meteor/client/c16/lib/lists/scrollable-list.tsx
```diff
@@ -1,94 +1,104 @@
-import { Tracker } from "meteor/tracker";
-import React from "react";
-import { wrapStatefulReact, wrapStatelessReact } from "../../../glue-react";
+import React, { useEffect, useRef, useState } from "react";
```
hhoegelo, line 1: maybe we can move React into the braces?

package/meteor/client/c16/lib/lists/scrollable-list.tsx
```diff
@@ -1,94 +1,104 @@
...
 
 export namespace ScrollableList {
     export interface ListProps {
         id: string;
+        heightOfChildInPX: number,
         numElementsBeforeFirstVisible: number; // this is for fast up scrolling, to fill the space that is shown when scrolling more than available
         numElements(): number;
+        listRevision(): any;
```
hhoegelo, line 9: is listRevisions return value really any?

package/meteor/client/c16/lib/lists/scrollable-list.tsx
```diff
@@ -1,94 +1,104 @@
...
-        }
+    export const Control = (p: ListProps) => {
+        const [scrollY, setScrollY] = useState(0);
+        const scroller = useRef(null);
+    
+        useEffect(() => {
+            p.initialRenderCallback()
+        }, [scroller, p.listRevision()]);
+        
+        const scroll = scrollY;
```
hhoegelo, line 36: The variables scroll, scrollY and newScrollY seem to be used a bit randomly - at least I as a reader could not tell when to use which and if the usages are correct.
Maybe I am wrong, but I'd say the usage of scroll in Ln 53 uses outdated value - and, if this 'scroll' can be removed there, it is unused all together.

package/meteor/client/c16/parameters/parameters.tsx
```diff
@@ -191,7 +185,7 @@ export namespace ParametersControl {
         }
 
         const parametersProps: Array<ParameterControlProps> = getSubPageParameters(pageBundle, subPage).map(foo => {
-            let prefix_control_factory = macro_controls.has(foo!.id.num) ? (id: ParameterId) => {
+            let prefix_control_factory = foo && macro_controls.has(foo!.id.num) ? (id: ParameterId) => {
```
hhoegelo, line 188: pls invent a proper name for foo

package/meteor/client/c16/presets/lists/bank-list.tsx
```diff
@@ -48,76 +59,79 @@ export namespace BankEntry {
...
+            id: "bank-list",
+            numElementsBeforeFirstVisible: 10,
+            numElements: () => props.banks.length,
+            createChildContent: (idx: number): React.JSX.Element => {
+                const index_correct = idx >= 0 && idx != undefined;
+                const bank_proxy = index_correct ? proxyFactory!.getBank(props.banks.at(idx)) : undefined;
+
+                const entry_props: BankEntry.Props = {
+                    bank: bank_proxy,
+                    name: () => bank_proxy ? bank_proxy.name : "Placeholder",
```
hhoegelo, line 111: is the 'Placeholder" supposed to stay in production code?

package/meteor/client/c16/presets/preset-search/ResultList.tsx
```diff
@@ -38,95 +48,101 @@ export namespace ResultEntry {
...
+                    display: () => "" + bankNumber + "-" + presetNumber + ": " + presetName,
+                    proxy: () => preset,
+                    isSelected: () => props.getSelectedPresetResult == uuid,
+                    isLoaded: () => preset && preset.uuid == proxyFactory!.getEditBuffer().loadedPreset,
+                    clickHandler: (id) => {
+                        setPresetSearchFocus(PresetSearchFocus.Results);
+                        const preset_proxy = proxyFactory!.getPreset(id);
+                        if (preset_proxy) {
+                            props.setSelectedPresetResult(id);
+                            // preset_proxy.select(); noot sure what to do here
```
hhoegelo, line 131: can we clarify this prior to merge?

package/meteor/client/c16/presets/presets.less
```diff
@@ -171,42 +92,14 @@
...
-    
         .presetmanager {
             display: flex;
             flex-direction: row;
             flex-wrap: wrap;
             align-content: stretch;
             justify-content: space-evenly;
             align-items: flex-start;
-            height: 100%;
+            height: 456px; // THIS must be Correct, i measured it
```
hhoegelo, line 102: is this over specific height value still needed, now that we have lowered the requirements for the list (fix children height)?

package/meteor/client/c16/presets/presets.tsx
```diff
@@ -95,24 +95,24 @@ export namespace C16PresetManagerLayout {
...
-                            :
-                            <div className="empty-banks-placeholder">No Banks!</div>
-                        }
-                    </div>
+                    {
+                        num_banks > 0
+                        ?
+                        <BankList.Control {...bankListProps} />
+                        :
+                        <div className="empty-banks-placeholder">No Banks!</div>
```
hhoegelo, line 103: Maybe we can have a more friendly wording here and also for the preset lists?

package/playground-lib/src/libundo/undo/Algorithm.cpp
```diff
@@ -56,22 +56,16 @@ namespace UNDO
     return ret;
   }
 
-  Transaction* Algorithm::find(Transaction* root, const Glib::ustring& targetID)
+  Transaction* Algorithm::findByUUID(Transaction* root, long targetTransactionID)
```
hhoegelo, line 59: the name should be findById()

package/meteor/client/c16/debug/debug.tsx
schlenkibus: it seems like we can't change that. When I try to put the React into the braces the compiler complains with this error:
`'React' can only be imported by using a default import.ts(2595)`
the agreed upon solution for this error (not for this specific use-case) seem to point to set `esModuleInterop` to true in tsconfig.json
I tried that but to no avail. I suspect that we are stuck with this syntax.

package/meteor/client/c16/lib/lists/scrollable-list.tsx
schlenkibus: see above

package/meteor/client/c16/presets/preset-search/ResultList.tsx
schlenkibus: i removed that comment and the dead code, but will have to do another iteration on the result-list with david anyways

package/meteor/client/c16/presets/presets.tsx
schlenkibus: I would leave this in place until @davidm2894 defined what we want to display instead (I would think an empty list would be better than any text, and this text will remind us to redo this)

package/meteor/client/c16/debug/debug.tsx
```diff
@@ -75,15 +75,23 @@ namespace C16Debug {
 
     const Led = (props: { id: LedId }) => (<div className="led"></div>);    
     const Button = wrapStatefulReact(() => { return {clickHold: new ReactiveVar<boolean>(false)}}, (props: {id: ButtonId}, _, state) => {
-        return <button className={state.clickHold.get() ? "held-button" : ""} onPointerDown={() => bridge!.injectButtonPress(props.id)} onPointerUp={() => bridge!.injectButtonRelease(props.id)} onContextMenu={e => {
+        return <button className={state.clickHold.get() ? "held-button" : ""} onPointerDown={(e: PointerEvent) => {
+            if(e.button != 2) {
```
hhoegelo, line 79: Consider introducing a documenting name for the magic number

hhoegelo (approved):

PR #3980, opened 2024-01-11
Dependency injection

hhoegelo:


package/playground-lib/src/presets/EditBufferConverter.cpp
```diff
@@ -6,15 +6,17 @@
...
+void EditBufferConverter::convertToDual(UNDO::Transaction *transaction, AudioEngineProxy &ae, SoundType newType,
+                                        VoiceGroup currentVG)
 {
   const auto oldType = m_editBuffer.m_type;
 
   if(oldType == newType)
     return;
 
-  SendEditBufferScopeGuard scopeGuard(transaction, true);
+  SendEditBufferScopeGuard scopeGuard(transaction, ae, true);
```
schlenkibus, line 19: not sure what is going on here, but the c15-playground-tests crash, and the closure in the scope-guard seems to be the issue. But I have not been able to find out the root-cause yet

